PR #34, opened 2013-12-19
Enhanced OpenJSCAD.org code with Touch device support.

miguelitoelgrande:


z3dev:
Folks. I think this pull request is no longer required. I believe that OpenJSCAD.org has support for touch events, and supports all touch devices. Please correct me if wrong.

PR #42, opened 2014-04-11
Parameter Dependencies

abudden:
This code adds the ability to show and hide parameters depending relevance (determined by another parameter).

It adds two new keys to the getParameterDefinitions array: 'depends' and 'dependvalue'.  If the former is defined, the latter must also be.

I have one concern with this code but I don't know what to do about it.  The formatting of the parameter table doesn't seem to be as consistent as it was before I applied this patch and I'm concerned this is a regression.  However, my knowledge of html formatting is close to zero and I don't know what to do about this.

### DIFF
--- a/openjscad.js
+++ b/openjscad.js
@@ -1245,8 +1245,7 @@ OpenJsCad.Processor.prototype = {
     var values = this.getParamValues();
     for(var i = 0; i < this.paramDepends.length; i++)
     {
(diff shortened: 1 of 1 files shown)


z3dev:
@abudden Sorry to say this but we are no longer accepting changes to V1.

However, I like these changes, and think these can be beneficial for others. Can you port these changes to V2?

PR #48, opened 2014-09-24
Issue#47 also accept parameters on url

ErikDeBruijn:
Also does something different... I've added a php remote script (to fetch) and a .htaccess rewrite rule so it can also be installed on a php enable server.


Spiritdude:
I merged it with the new dev-branch, so eventually will become master again.

PR #66, opened 2015-02-09
Fix mirror implementation

### LINKED ISSUE #65: mirror(vec, object) doesn't properly mirror
The behavior of mirror doesn't reflect what OpenSCAD does, nor what I would expect it to do. I actually think the entire [mirror function](https://github.com/Spiritdude/OpenJSCAD.org/blob/08f603bac36d056a81ecaaf234a9217db48b9fc0/openscad.js#L659) needs to be reimplemented. I'm guessing when this was implemented it was a copy-paste of rotate (note the comment) and it doesn't actually work correctly like rotate. I have not yet tried rotate though, so I don't know if that is broken.

Existing code:
```javascript
function mirror(v,o) { 
   var a = arguments, v,o,i = 1, r = 0;
   if(arguments.length==3) {  // mirror(r,[x,y,z],o)
      r = a[0];
      v = a[1];
      i = 2;
      if(a[2].length) { a = a[2]; i = 0; }
      
   } else {                   // rotate([x,y,z],o)
      v = a[0];
      i = 1;
      if(a[1].length) { a = a[1]; i = 0; }
   }
   for(o=a[i++]; i<a.length; i++) { 
      o = o.union(a[i]);
   } 
   if(r!=1) {
      return o.mirroredX(v[0]*r).mirroredY(v[1]*r).mirroredZ(v[2]*r);
   } else {
      return o.mirroredX(v[0]).mirroredY(v[1]).mirroredZ(v[2]); 
   }
}
```

Notes:
* AFAICT these mirrored{X,Y,Z} functions don't take a parameter like that, so the whole r calc doesn't apply
* additionally, since those params don't matter, the code is *always* doing 3 mirrors
* [OpenSCAD mirror](http://en.wikibooks.org/wiki/OpenSCAD_User_Manual/The_OpenSCAD_Language#mirror) doesn't have 3 cases like what is described here. It has just a vec and objects

I believe the code should really be:

```javascript
function mirror(v) { 
   var o = null,
       i = 0,
       a = arguments.slice(1, arguments.length - 1);
   for(o=a[i++]; i<a.length; i++) { 
      o = o.union(a[i]);
   }

   var plane = new CSG.Plane(new CSG.Vector3D(v[0], v[1], v[2]), 0);
   return o.mirrored(plane);
}
```

Which I'll PR over in a bit, once I've finished testing.

richievos:
Previous was a copy-paste of rotate that didn't really work

Fixes #65

### DIFF
--- a/openscad.js
+++ b/openscad.js
@@ -656,27 +656,15 @@ function rotate() {
    }
 }
 
-function mirror(v,o) { 
-   var a = arguments, v,o,i = 1, r = 0;
-   if(arguments.length==3) {  // mirror(r,[x,y,z],o)
-      r = a[0];
-      v = a[1];
-      i = 2;
-      if(a[2].length) { a = a[2]; i = 0; }
-      
-   } else {                   // rotate([x,y,z],o)
-      v = a[0];
-      i = 1;
-      if(a[1].length) { a = a[1]; i = 0; }
-   }
-   for(o=a[i++]; i<a.length; i++) { 
+function mirror(v,o) {
+   var a = Array.prototype.slice.call(arguments, 1, arguments.length),
+       o = a[0];
+
+   for(var i=1; i<a.length; i++) {
       o = o.union(a[i]);
-   } 
-   if(r!=1) {
-      return o.mirroredX(v[0]*r).mirroredY(v[1]*r).mirroredZ(v[2]*r);
-   } else {
-      return o.mirroredX(v[0]).mirroredY(v[1]).mirroredZ(v[2]); 
    }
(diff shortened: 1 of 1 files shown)


Spiritdude:
Thanks for the fix, I didn't test 'mirror' so thoroughly before, I vaguely remember doing just 1-2 tests with my code and produced the same results as OpenSCAD, but you seem to have looked at it quite close.

Spiritdude:
Forgot to ask, I just changed license information of openscad.js from GPL to MIT (before I accepted your pull request), to make the whole OpenJSCAD package more consistent license-wise; can I consider your changes be also MIT licensed?

richievos:
MIT license is fine with me. Thanks for the merge

PR #71, opened 2015-02-21
Update csg.js

NateTG:
Change to handle

https://github.com/Spiritdude/OpenJSCAD.org/issues/63

...
Line 5098 wants a fscking semicolon, but I can't edit the patch and javascript will probably insert it so it works.

It also might be more appropriate to change the matching radius to 1e-10 to match the CAG cannonicalization params.

### DIFF
--- a/csg.js
+++ b/csg.js
@@ -5090,9 +5090,12 @@ CAG.fromPoints = function(points) {
 		var point = new CSG.Vector2D(p);
 		var vertex = new CAG.Vertex(point);
 		var side = new CAG.Side(prevvertex, vertex);
-		sides.push(side);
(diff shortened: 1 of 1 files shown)


Spiritdude:
@bebbi is currently working on cgs.js from OpenJsCad to merge to OpenJSCAD.org, perhaps he can include your small update too.

bebbi:
we'll have a new version of csg.js here today or tomorrow. It's pulled in from a different repo, and it'll contain info for where to contribute in the top section.

bebbi:
I believe this has been fixed before in joostn which is now integrated here. Please re-open an issue there if not the case.

PR #73, opened 2015-03-01
Touch up index html

richievos:
I was going through the index.html code, and noticed it was missing some quotes on attributes, and the indentation was different in some places, so I went in and moved some stuff around.

This PR makes it look like there are a ton of changes, but it's all just simple text formatting / moves. I made all the commits very focused, so you could see all the changes independently.

In general, there's no functional changes here, and I went through and tested all the big hitters (editor, file upload [jscad, stl], auto reload) and things all worked. I did however fix a couple places where there was bad markup, such as </p>s without any <p>. I believe there are still a couple in there, but didn't want to go any further until you confirmed you're ok with this.

I also made some minor refactors where I switched the onload hook to use $(document).ready, since that's what the rest of the ui hooks were doing.

Future improvements:
* move the css to its own file
* move the onclicks into index.js (unobtrusive js style)
* convert some <br> tags to either css or <p>
* cleanup any other unpaired blocks (div/p in particular)

I'll take a pass at some of those if you're cool with these changes.


richievos:
Also, if you'd prefer a different naming structure than "index.js" and "ui-*" I'm happy to do some renames.

richievos:
Also, I should note the context for me doing this. I'm hoping to move this code around so that I could create another webpage/script for locally editing openjscad files. In particular, I'd like to be able to do something like:

```
openjscad some_file.jscad # generate the stl (already exists)
openjscad-viewer some_file.jscad # open a browser window viewing that file with auto reload on
openjscad-viewer some_file1.jscad some_file2.jscad # show multiple files
```

So having the content in separate javascript files / cleaning up index.html will help me create a separate viewer.html for that purpose.

Assuming I can get it to work, I'll PR that as well.

My end goal would be to write a small set of CLI tools to simplify my creating of files. Let me view the output, auto reload that UI, and auto produce STLs from it.

Spiritdude:
A quick comment, I really like also others focusing on the CLI level, as it was important for me to bring OpenJSCAD there too - so I welcome focus and improvements there. Give me a few days to look at it ...

Spiritdude:
@richievos I gonna wait until https://github.com/Spiritdude/OpenJSCAD.org/issues/74 is done (newer csg.js included and separating import/exports). Perhaps we can then still automatically merge your changes. Perhaps @bebbi and @z3dev take a look at it as well.

richievos:
Fine with me. I'll bump this in a week or so if I don't hear anything by
then.

richievos:
@Spiritdude just FYI, I pushed some more changes. I've tried to put a bit more separation between the different functionality going on in here. In particular, the editor code integrating with the viewer code integrating with the drag and drop code.

I've also put some documentation defining what stuff each of those pieces depends on.

z3dev:
I also have some refactoring of index.html and the java scripts, so let's work on this together once the CSG.js / formats.js is closed. I'll review your changes this week.

richievos:
A couple additional points:

* when everything's all good to go, I think it'd be good to take another formatting pass at the javascript files. There appears to be some 2 space tabbing, some 3 space, ... I'm honestly not sure if that originated from me, but I think it was that way before I pulled the code out
* I'd like to move the js files I've been creating out of the root dir, but don't want to do that until after a merge
* I think it'd be good to modernize some of the js files, and do explicit public/private scoping, but again, I'm trying to go step by step
* @z3dev this may be more relevant for you, by me calling out the dependencies, I think I'm close to having the coupling between this UI and the code in openjscad being broken apart. In particular, it seems like there's only a couple things still leaking through (such as the list of file objects)

@z3dev I'm not sure how much time I'll have for digging into this more this week, but if you could push whatever changes you have around index.html and the javascripts somewhere, so I can compare mine to yours, that'd be useful. The main thing I'm trying to pull off right now is decoupling the different parts of the UI code from each other, and defining the entrance points.

Optimally I'm thinking it would end up with the code broken out into:
* the viewer (already mostly part of the other libs, but things like setting it up to use #viewer)
* the editor UI (mostly extracted out in this diff)
* file loading
* drag and drop
* file export

and then also take a pass at separating out the styles from the markup and/or the javascript

z3dev:
I reviewed these changes and have only a couple of questions.
1) I'm not sure if a 5 line JavaScript file is better than just putting the script in the HTML. This is probably a style question.
2) <body onload=...> was replaced with $document.ready() so I want to make sure that this works without side affects. Did you test on-line and off-line? That's probably enough.
3) Separation of EDITOR scripts from VIEWER scripts is good. I did the same, so I'm fine with this.

One other thought on the separation of functionality. Each part should be able to stand alone. For example, the viewer should work if the editor is disabled.

I suggest pulling these changes into a new branch. That will allow slightly easier testing, and further changes.

richievos:
> 2) "body onload=..." was replaced with $document.ready() so I want to make sure that this works without side affects. Did you test on-line and off-line? That's probably enough.

yes

> 1) I'm not sure if a 5 line JavaScript file is better than just putting the script in the HTML. This is probably a style question.

Partially a style thing, partially a best practice:
* http://en.wikipedia.org/wiki/Unobtrusive_JavaScript
* http://programmers.stackexchange.com/questions/86589/why-should-i-avoid-inline-scripting
* http://stackoverflow.com/questions/2837921/is-it-bad-practice-to-embed-javascript-into-the-body-of-html

There's a good deal of stuff that could use a freshening up in that page (css, tag use, ...), but I didn't want to touch all of it.

> One other thought on the separation of functionality. Each part should be able to stand alone. For example, the viewer should work if the editor is disabled.

As designed right now, that's very difficult. My changes make it _closer_ to being possible, by defining  hooks between the pieces, but they're still overly coupled. I don't want to change that until afterwards.

> I suggest pulling these changes into a new branch. That will allow slightly easier testing, and further changes.

I don't understand what you're advocating for there.

z3dev:
I'm fine with the merge of these changes.

I think we both have further changes planned for the HTML, javascripts, etc. I thought it might be easier if we work together on a few points, and then provide that back to @Spiritdude when ready. A branch would be ideal.

z3dev:
@richievos Please see #108 as many of these changes were incorporated directly, and more.

If satisfied with #108 then I'll close this, and we can continue on with further partition of functionality.

richievos:
Thanks so much for the work @z3dev, and sorry for my incredibly delayed response.

z3dev:
No problem. FYI, I reused most of your changes. It was a snap. Thanks.

PR #75, opened 2015-03-02
Contribution to CSG

vr050714:
Hi

There are two small changes in CSG.
1. Fixed  CSG.Matrix4x4.mirroring method. The minus sign in the last row of reflection matrix resulted in wrong reflection. For example code
```javascript
function main() {
    var MirrorNormal = [0,0,1];
    var MirrorPoint = [0,0,1];
    var plane = CSG.Plane.fromNormalAndPoint(MirrorNormal, MirrorPoint);
    var a = cube([5,2,1]);
    var b = a.mirrored(plane);
    return [a, b.setColor([1,1,0])];
}
```
resulted in geometry like on the picture
img1
whereas yellow box should be on top of the magenta one. This mirroring method is used in mirroredX, mirroredY, mirroredZ methods of CSG and CAG objects prototypes. But everywhere the origin point belongs to the mirroring plane and variable w = 0.
2. Added Euler rotation angles in ZXZ notation. It has been added to prototypes next to the other transformations.

Thanks
Vladimir


bebbi:
plan to integrate this into joostn/openjscad csg.js tomorrow

PR #93, opened 2015-07-04
Install openjscad cli with npm and use as Node module

runvnc:
These changes do not seem to affect the browser actually.  I set it up so if you do an 'npm publish'  then people can just npm i -g jscad and then the openjscad program will install.  Also, adding jscad to a node project and then requiring allows use of cube() etc.   as a dsl in a Node program or module.  And there is an inc() that globalizes exports so you can add on to your cad dsl with Node modules.  And a renderFile function for outputting an STL etc.


z3dev:
I have a question about the exports in openscad.js. The export of the function look fine but I wonder about the export of CSG. Of course, openscad.js requires the functions found in CSG.js but why export CSG? Unlike formats.js, there are no changes made to CSG methods or data members within openscad.js.

Thoughts?

jeromew:
I understand that this PR was made into an npm module here - https://www.npmjs.com/package/jscad even if the PR has not been merged.

This could be a start for the NPM'ification & Browserify'ing of openjscad but I would prefer if openjscad commiters are fully onboard with @runvnc's initiative. This module could be a start for implementing the steps in https://github.com/Spiritdude/OpenJSCAD.org/wiki/Proposed-Changes-for-NPM-and-Browserify

@z3dev regarding the export of CSG, I think it comes from the fact that CSG primitives are sometimes used in jscad files. Cf for example the celtic-ring example which uses `CSG.Matrix4x4()` among other methods. 

a server side usage of openjscad without using a .jscad file (direct API calls) would need a CSG instance to build complex things I suppose.

The real problem I see with the current jscad npm module is that it uses `g` to make a lot of things global. This may be an acceptable temporary thing in order to have a starting point for the npm cleanup (because openjscad makes a lot of "global methods" assumptions - but it can't stay this way very long before people start saying that is jscad cannot be included safely.

runvnc:
I don't really have time to work on this anymore.  But anyway it was a few hours of work at most, so you guys are happy to handle npm however you want.  I did it the way that made sense to me based on how OpenJSCAD worked, so that is the explanation for why things are exported or global, and I don't agree that globals are 'unsafe' especially in this context.   

I think I will leave the current npm module as is and of course anyone can publish another one named openjscad or whatever-cad etc.

jeromew:
@runvnc, I did not want to sound negative on what you have done. I am new to the project do not know enough on it. This is great to have a working npm patch and it is too bad that you do not have time for this anymore. Thanks for the working version.

z3dev:
@runvc Some of the changes have been merged into OpenJSCAD.org. Thanks.

And we would love to have full NPM and Browserify support but there are still some fundamental changes happening to the library, i.e. reorganization of scripts, object encapsulation, name space changes, libraries, etc. I believe we are still moving towards this goal, but slowly.

@jeromew Please review the latest version of OpenJSCAD.org, as there have been some improvements. It might be time to revisit this again.

Also, I'm not rushing to complete this functionality. I would prefer to have a feature complete OpenJSCAD.org. I think a few comments from @Spiritdude would be welcomed as well.

jeromew:
for future reference, I mention the maker.js project - http://microsoft.github.io/maker.js/ that handles the models as npm modules. This is maybe something that openjscad could do one day.

PR #96, opened 2015-08-12
Preparations for More Changes

z3dev:
Changes include:
- Added OpenJsCad.version and changed code to use this version
- Changed the comment that is shown while rendering "Rendering code" to "Rendering"
- Removed tabs and indented properly

If you run 'diff --ignore-all-space' with old and new versions then you find that there are just a few minor changes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,6 +34,7 @@ History:
 Todo:
 - saving source code edited in ACE locally, perhaps 'SaveAs' (.JSCAD, .STL, .AMF)
 - DONE: include(); support of multiple files, see http://jsfiddle.net/yybRu/
+
 -->
   <meta http-equiv="content-type" content="text/html; charset=UTF-8"/>
   <title>OpenJSCAD.org</title>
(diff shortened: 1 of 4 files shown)


z3dev:
The 7bb9f4 commit contains the following changes.
* Added OpenJsCad.env() for logging the current environment and version
* Added OpenSCAD compatible version()
* Replaced all usage of global version with either OpenJsCad.version() or version()
* Removed dependency on me in openjscad.js and openscad.js
* Added module support to openjscad.js and openscad.js
* Updated openjscad to use modules properly

There's a conflict in index.html due to the version number change. But version is now part of OpenJsCad, so this can be ignored.

z3dev:
The latest commit contains the reworked code for the parameters. This work enables a lot more parameter types, and groups. A summary of the testing is attached.

[OpenJsCad Parameters.pdf](https://github.com/Spiritdude/OpenJSCAD.org/files/6587/OpenJsCad.Parameters.pdf)

PR #107, opened 2015-12-08
fix plane value in fixTJunctions

sergerusso:


z3dev:
@sergerusso can you provide some examples for testing? that would help us to determine the issue, and check if this fix does the job. i think that others will be really excited if this fixes the issues with STL normals.

sergerusso:
I had some problems in slic3r with models like this, for example.

`function main() {
   return union(
      difference(
         cube({size: 10, center: true}),
         cube({size: [8,8,10], center: true})
      )
   );
}
`

z3dev:
@sergerusso Thanks again for the changes!

I ran several different JSCAD and SCAD scripts, producing simple and complex 3D meshes. All exported STL meshes were perfect (no T-junctions) after using those changes.

kaosat-dev:
@z3dev & @sergerusso  do you know if this is still applicable? If it is it would be nice if we could update & merge it into https://github.com/jscad/csg.js ?
If not we should close this PR rather than keeping it open forever in limbo :)

z3dev:
@kaosat-dev Thanks for the reminder.

I don't think that these changes were merged, but it should be easy to distinguish.

I tested via my website.

kaosat-dev:
@z3dev  turns out that yes, these changes have already been merged : https://github.com/jscad/csg.js/blob/master/src/CSG.js#L1142

I am closing this.

PR #108, opened 2015-12-09
Partitioning of HTML from JavaScript in Pages

z3dev:
These changes separate the JavaScript from the HTML, and create manageable pieces for the page, the editor, the drag-and-drop, etc. This will help to create further pages as well as extend functionality.

@richievos was a great inspiration, and many of his changes were bundled into these.

This is fully tested off-line, on-line, and CLI. In addition, testing was performed with Firefox, Chrome, Safari, and Opera on Mac OSX.

### DIFF
--- a/index.html
+++ b/index.html
@@ -58,79 +58,17 @@ Todo:
     <script src="ace/ace.js?0.3.1" charset="utf-8"></script>
     <script src="js/ui-drag-drop.js" charset="utf-8"></script>
     <script src="js/ui-editor.js" charset="utf-8"></script>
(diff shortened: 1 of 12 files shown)


Spiritdude:
I just checked the "Options" (not sure it was activated in this PR), on Firefox the select from multiple does not work, the list disappears just before I can select it. On Chrome I can select from the multiples.

Spiritdude:
Also, there should be a slight visual difference of the editable area vs the rendered area, before there was a slight difference (editing area slightly more milky), now it's gone, and only the cursor indicates this. I recommend a visual indication be back, one should see through it but clear see also, this is the text-area which behaves differently.

Issue #79, opened 2015-03-26
Feature Request: Handle to adjust code pane width


thehans:
It would be nice to have a handle which can be dragged by mouse so the user may adjust the code pane width.  

I find the code pane to be too narrow if I tile two windows vertically, which I often do so I can reference OpenJSCAD documentation and other information in the adjacent window.

Another possible improvement could be a toggle to dock the code pane to the bottom instead of the side.

z3dev:
I made a slight change that adds a tab to the editor, and allows the editor to be slide in/out on the right. Coming soon!

z3dev:
This change is now available on the DEV branch.

z3dev:
Closing.
OpenJSCAD.org 0.4.0 added a tab to the editor panel, which allows opening and closing the panel.

PR #109, opened 2015-12-16
Enhanced the editor frame to slide in and out by clicking a tab

z3dev:
This is a nice enhancement, allowing the editor frame to slide in and out. There's a new tab which can be clicked to control the view.

This addresses #79 by @thehans 

I've been using this for some time at www.z3d.jp/lab/


Spiritdude:
Nice feature, but it would be better to actually drag it ... to narrow and make it wider.

Spiritdude:
The left-hand handlers works different, hovering is enough (self explanatory, does not need click, it acts right away, and the behaviour is introduce at loading of the page). Alike the right handler should be, self-explanatory.

Spiritdude:
Btw, the handle you can adapt as defined in imgs/misc.svg, I designed left-right, right-left, top-down, down-top versions, best use Inkscape to add it to imgs/misc.svg

Spiritdude:
I provide suitable handle for in/out switch.

Spiritdude:
I just added it, proper handle and transitional animation.

z3dev:
@Spiritdude 
Cool. I'll use Inkscape to create ew graphics next time.
However, the individual PNG files are missing. Can you provide?

PR #116, opened 2016-01-06
Reworked JSCAD Script Processing

z3dev:
This is a fairly significant change, moving away from code within comments within code. The functionality is now split into jscad-worker.js (ASYNC) and jscad-function.js (SYNC). The code can now be fixed/debugged easily, and enhanced if additional functionality is required.

Note: The old async and sync functions still exist, and should be removed once the new code has been tested completely.

In addition, there are a few small changes to drag and drop to accommodate and streamline processing.

Finally, the strange global variable _includePath is now derived from the URL of script. This global is no longer required.


Spiritdude:
I just ran through all the Examples, they all work ...

z3dev:
Excellent!

Issue #137, opened 2016-02-25
smooth lighting


makc:
I am sure [you are aware](https://github.com/Spiritdude/OpenJSCAD.org/blob/adc37e44a6df1074581983a22be9f21cdc072ade/openjscad.js#L507) of smooth lighting problems. Well, I have been fixing them in the project that is basically private fork of your code. Although I cannot just copy the code over here, I can describe what I did. 1st, you have the bug [here](https://github.com/Spiritdude/OpenJSCAD.org/blob/adc37e44a6df1074581983a22be9f21cdc072ade/openjscad.js#L527) - it should check the color against colors[vertexindex], or else colors are interpolated. 2nd, you want to add different colors to CSG primitives' surfaces (such as cylinder top, bottom and side), and 3rd you want to make sure these colors are never the same, e g [here](https://github.com/Spiritdude/OpenJSCAD.org/blob/adc37e44a6df1074581983a22be9f21cdc072ade/openscad.js#L344) you set the same color for the whole surface, and that effectively merges all cut surfaces. I hacked around it by re-using existing "shared"-s on polygons, but you need to think of better solution to all this, and then you will have this kind of shading:

![screen shot 2016-02-24 at 23 54 20](https://cloud.githubusercontent.com/assets/242577/13330765/a37436ee-dbfa-11e5-9761-b778ede3629b.png)

z3dev:
I wasn't aware of this as all. Interesting... The choice to use smoothing or not should be an Viewer.option as well.

We would love to have the bug fixes at least. Can you create a pull request for your changes? Please use the DEV branch when creating the pull request.

kaosat-dev:
this is going to be added to the new version of the 3d viewer , coming soon

z3dev:
Smooth lighting is the default for V2 rendering. Closing this issue.

PR #138, opened 2016-02-26
stop color interpolation with smooth lighting on

makc:
Ah what a hell. The color spill bug fix referenced in #137

### DIFF
--- a/openjscad.js
+++ b/openjscad.js
@@ -523,8 +523,13 @@ OpenJsCad.Viewer.csgToMeshes = function(initial_csg) {
 
     var indices = polygon.vertices.map(function(vertex) {
       var vertextag = vertex.getTag();
-      var vertexindex;
-      if(smoothlighting && (vertextag in vertexTag2Index)) {
+      var vertexindex = vertexTag2Index[vertextag];
+      var prevcolor = colors[vertexindex];
+      if(smoothlighting && (vertextag in vertexTag2Index) &&
+        (prevcolor[0] == color[0]) &&
+        (prevcolor[1] == color[1]) &&
+        (prevcolor[2] == color[2])
+      ) {
         vertexindex = vertexTag2Index[vertextag];
       } else {
         vertexindex = vertices.length;


makc:
That's what I did, but perhaps proper solution would be to compare polygon.shared

z3dev:
@makc thanks
The smoothing of rendered objects is interesting.
I tried the changes and viewed some of the examples. Some turned out very nice, like the Spring example. While others turned out a little strange, like edges on cubes or flat faces.
It seems to me that smoothing could be beneficial for different applications, i.e. those with different requirements for viewing objects. Why are you using smoothing?
Although not perfect, I think that the changes improve this little know feature.

makc:
> While others turned out a little strange, like edges on cubes or flat faces

this PR only fixes colors leaking into unrelated faces. in order for the cube to look as expected, its faces need to have different colors (I used colors that differ by 1e-4 so that actual color looks the same).

makc:
> Why are you using smoothing?

I do not actually use smoothing, I used matcap materials:

screen shot 2016-02-25 at 14 09 46
screen shot 2016-02-25 at 17 27 59
screen shot 2016-02-25 at 17 43 07
screen shot 2016-02-25 at 17 43 33

but this is the client request, they were not happy with flat look it had.

z3dev:
Thanks for all this very useful information.

How did you incorporate the matcap lighting into the openjscad.js library? That would be a very interesting addition to OpenJSCAD.org. You could introduce this to the OpenJSCAD.org community, and see if anyone has interest.

I do!

Making OpenJSCAD.org more useful and flexible to others is a major goal of mine.

makc:
once you have smoothing shading fixed, you can swap the shader to whatever you want... including matcap ) you can find reference implementation [here](https://www.clicktorelease.com/blog/creating-spherical-environment-mapping-shader), although it is somewhat broken, too, if you look at the edges, for example, and relies on undefined behavior of pow.

z3dev:
I'm closing this pull request as the changes have been incorporated and tested in the 0.5.1

Thanks again!

PR #155, opened 2016-05-28
AMF Parsing and Conversion based on SAX XML

z3dev:
Here's the new version of the AMF parsing, found in js/jscad-parseAMF.js

Of course, there were some changes to the other scripts to use this properly. And some simplification of the CLI script as well.

Thanks for the nice code base which provided the base logic for the conversion to JSCAD.

New Features:
1) Support for multiple objects
2) Support for color on triangles, objects, etc.
3) Support for scaling based on 'unit' of AMF, e.g. inch, meter, etc.

To Be Done:
1) Support for color interpolation on vertices
2) Support for compressed AMF files (ZIP compression)


ObDest:
FYI, ‘AMF’ here appears to refer to https://en.wikipedia.org/wiki/Action_Message_Format (‘Action Message Format (AMF) is a binary format used to serialize object graphs such as ActionScript objects and XML, or send messages between an Adobe Flash client and a remote service, usually a Flash Media Server or third party alternatives.’)

PR #156, opened 2016-06-16
Viewer rewrite

apla:
Hi!

I've started to merge development histories to make one very good codebase instead of three just good:
- https://github.com/joostn/OpenJsCad — joostn
- https://github.com/Spiritdude/OpenJSCAD.org — this
- https://github.com/matiasinsaurralde/atom-scad-preview — atom plugin to edit jscad files

This patch set separate viewer from openscad.js and now viewer supports both three.js and lightgl.js renderers.

Minor features:
- HiDPI display support


z3dev:
Cool! Thanks for these changes.

My preference would be to move all OpenJSCAD.js scripts into the js/ directory. So, the new scripts should be..
- js/openjscad.viewer.js
- js/openjscad.viewer.lightgl.js
This will help prepare OpenJSCAD.org for packaging by Browserify.

Also, when developing the Three.js viewer, please put the "libraries" into the js/lib/ directory. This will allow OpenJSCAD.org to run as a local application.

apla:
@z3dev I did it

z3dev:
Thanks.

We have a few changes which need testing and release. That release will be created first, then I'll merge these changes.

z3dev:
@apla 

Question... Why change the based class to support Atom? You broke the Viewer out to allow derivations. It seems to me that writing a derived "Atom" version of Viewer would be a better choice.

apla:
Closing in favor of #202

PR #171, opened 2016-09-04
Unbreak intersection() & difference() for CAGs.

fischman:
Only .setColor() in intersection() & difference() for non-CAGs.

A further improvement would implement CAG.setColor(), but that's not here.

To see the problem this fixes, observe this works:
`function main() { return union(circle(), circle().translate([1, 0, 0])); }`
but replacing `union` with `difference` or `intersection` errors out with:
`Error in line 353: Uncaught TypeError: a[i].setColor is not a function`
This patch removes this error and results in correct-looking intersection/difference.

### DIFF
--- a/openscad.js
+++ b/openscad.js
@@ -340,9 +340,13 @@ function union() {
 function difference() { 
    var o,i=0,a=arguments; 
    if(a[0].length) a = a[0]; 
-   for(o=a[i++]; i<a.length; i++) { 
-      o = o.subtract(a[i].setColor(1,1,0));     // -- color the cuts
-   } 
+   for(o=a[i++]; i<a.length; i++) {
+      if (a[i] instanceof CAG) {
+        o = o.subtract(a[i]);
+      } else {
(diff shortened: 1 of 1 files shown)


z3dev:
@fischman Thanks for the changes. I think these are fine.

However, I don't like hard coded "colors", scale, etc. I'm wondering if the "cut' color should be configurable. And as you suggested, both the "union" color and "intersection" color should be changeable. Some other people want to change the defaults for color and resolution as well. A "common" approach would be best.

fischman:
@z3dev your comments about configurable colors sound fine to me, but seem unrelated to the change proposed here (and I don't grok precisely what you're proposing so I can't implement it).  Since this PR fixes a blocking bug for me would you mind applying it as-is?

fischman:
Not sure why I thought this was blocking me before - I can work around the bug easily enough by using foo.subtract(bar) instead of difference(foo, bar), and so on.  Do with this patch what you like :)

PR #180, opened 2016-10-04
Camera fix

jallwine:
Decoupled zoom and viewpointZ and fixed touch/mouse events to properly shift the camera.


z3dev:
@jallwine thanks for the nice changes.

fyi, release 0.5.2 is in progress already, which includes testing. so, i'm not sure these changes will get into 0.5.2 or not.

just for a sanity check, did you test all the examples with these changes? that's the lint test for all changes.

jallwine:
Not all of them, just a couple of my own and the default example. I didn't touch anything that would be different for different examples. This change just makes sure the camera is rotating about the point you're looking at, whereas currently when you rotate you're always rotating about the origin.

jallwine:
You can see it in action here: http://www.allwinedesigns.com/blog/work-table

z3dev:
@jallwine Could you reapply your changes to the new restructured source code? It would be a great help, and allow testing to be completed.

jallwine:
Yeah, I'll do my best to find some time soon. I've been meaning to check out how everything was restructured.

PR #192, opened 2016-11-16
fix `make deinstall` to properly deinstall openjscad

hobofan:
Pretty small fix, that makes `make deinstall` actually deinstall the bin and the directory that was created on `make install`.

Nothing too major but good hygiene for any project.


kaosat-dev:
Looks good to me ! tested on my machine, works as expected !

PR #202, opened 2016-12-07
Viewer rewrite, second attempt

apla:
I see, this is quite hard task to understand what's really changed in viewer in my  previous pull request #156  and I've decided to make a proper changeset, corrected for current OpenJSCad.org version.

Here is the demo: https://apla.github.io/OpenJSCAD.org/side-by-side.html


kaosat-dev:
This is looking very good ! 
Thanks for the great work !
I checked out the code, and it seems good, all examples run fine for me on Chrome ,Firefox & Safari (Mac).
Can we merge it ?

PR #209, opened 2016-12-31
Some fixes

z3dev:
Corrected parameters to putSourceInEditor()


kaosat-dev:
The fixes to the parameters for putSourceInEditor look good to me !
The makeFile, however will not work : there is no more 'js' folder in the new structure, and install should ideally be done via npm install -g Spiritdude/OpenJSCAD.org .
I can look into having custom installation paths, but that might break things under some node.js environments.
@z3dev  could you perhaps seperate this into two different PRs please, so we can merge the first part and deal with the installation seperatly ?

kaosat-dev:
Looks good to me !

PR #210, opened 2016-12-31
Fix for echo() function

kaosat-dev:
This adds back/ fixes the 'echo' feature and adds a simple cli test (for now) to test it


z3dev:
The changes for echo() look fine.
FYI, if echo() is decommissioned then some of the examples need to be updated.

PR #211, opened 2017-01-01
Modeling tests & superficial code cleanup

kaosat-dev:
Ready for MERGE !
This PR adds unit tests for all the functional modeling helpers, so that we have a good , tested baseline to expand things from.

- [x] 2d primitives
- [x] 3d primitives
- [x] extrusion ops
- [x] boolean ops
- [x] transformation ops
- [x] colors 
- [x] maths 
- [x] text 

To run all the test (including the already passing ones for the CLI & module) type `npm test`
or `npm t`

Non passing tests are there for a reason:
- unexpected behaviour 
- weird api issues 
- this PR is not meant to fix issues, just to add the tests highlighting them :)
unless I got something really wrong :)

PR #212, opened 2017-01-07
Node x3d support

kaosat-dev:
This PR adds node.js support to export to X3D (you should be happy about this one @mtippett :)

- [x]  x3d export works in browser and node (tested)
- [x]  added unit test as part of src/module.test.js

IMPORTANT : please merge this PR after #211


z3dev:
Why DOM? Why not reuse the SAX library? Just wondering...

kaosat-dev:
SAX is just a parser , not a writer, and I used `xmldom` because it is the same api as in the browser ☺

PR #217, opened 2017-01-17
Updated & fix integration of openscad-openjscad translator

kaosat-dev:
This PR :
- fixes the issues with the translator and our new infrastructure.
- depends on an updated (read, cleaned up) translator within the JSCAD org.
- removes the last 'hard dependency' : ie removes the js/libs folder from this project
- fixed scad code display


z3dev:
Are you sure that the switch to SCAD syntax is correct? I believe there's code to determine the syntax and switch the editor to SCAD when SCAD is loaded.

PR #222, opened 2017-01-21
split out openscad/scad api to separate repository 

kaosat-dev:
This PR  goes in tandem with the new repository here https://github.com/jscad/jscad-scad-api
- removed local code
- modified a few imports
- modified package.json with VERSIONED dependencies
- it also means all tests for this repo are now passing :)


z3dev:
I like it but have some questions about the versioning. There are really two (or more versions) needed. One for the SCAD API supported or simulated. And one for the release. How to manage these?

kaosat-dev:
The SCAD API  should  follow strict [semantic versioning](http://semver.org/) : so that any breaking change increase the major version number ie : go from 1.0.0 to 2.0.0 etc this is essential as it guarantees that any parts coded with that version will work with that version .
For this repo I am not entirely sure : it depends entirely on WHAT we consider it to be:  a front end for the SCAD version ? for the SCAD and csg.js ? What is the api of OpenJSCAD.ORG: version changes should reflect a level of 'safety' for users. 
It does not impact this specific PR I think , since I added explicit SCAD api version number , so it locked for now :)
But I believe a discussion about this in the Issues is needed.
(Also I am preparing a small diagram to visualize more clearly the 'layers of the onion' ie what depends on what.

PR #224, opened 2017-02-11
Changes to support viewer-minimal.html and viewer-options.html

z3dev:
These changes are for min.html which has been modified to use the new distribution of index.js

The HTML has been changed but additional changes need to be made to allow the loading of a specific script, and set into the processor.

index.js has been changed to be more flexible, as HTML pages should be able to control functionality by specifying specific divisions. For example, the About dialog is not required if the "about" DIV is not present.


kaosat-dev:
Gosh , I am sorry , I totally missed this ! 
Very cool, I'll take it for a spin later today and post back.

z3dev:
min.html is now working fine. however, big.html needs to modify the options of processor and viewer. i'm open for suggestions on how to implement this.

how about another build script to create a special big.js?

kaosat-dev:
Sorry for the delay @z3dev !
Adding a build script for big.js is not a big issue, but I think there is already quite a big of complexity going on in that area,  is there anything preventing 'big' from , just passing a different set of options to the processor and viewer ?

kaosat-dev:
actually after taking a better look at the code , perhaps they should indeed be seperate versions of jscad but a few questions:
- min & big are not very telling on what they imply feature wise, could we rename them to something more clear ? like min => jscad-min-viewer etc
- do they actually belong in the main project ? if something seems like a 'different' tool , it just might actually be one ?

kaosat-dev:
For now I think :
- add the seperate build script for big
- merge this so I can integrate the changes to dropZone :)

z3dev:
These are examples of how to use JSCAD components in a HTML page, using a minimalistic approach. Maybe some more thought can be given to the organization of the HTML components as well, but for now... I'll rename min and big appropriately.

min.html to viewer-minimal.html
big.html to viewer-options.html

kaosat-dev:
@z3dev : could you please finish this one & merge it , in order to help me finish the drag & drop part ? :) I am scared of the possible merge conflicts :)

z3dev:
OK. I'll try to complete this in the few days. Just back from a nice break...

z3dev:
@kaosat-dev If you could do the needful and review these changes, then I would be deeply in debt.

Issue #227, opened 2017-02-26
Makefile has hardcoded sudo install locations conflicting with Homebrew, etc.


andrewgho:
The `make install` command from the Makefile has a couple idiosyncrasies:

* The `openjscad` program is installed into hardcoded `/usr/local/bin`, which, on OS X boxes using Homebrew, is controlled by Homebrew; and is also a common directory for other managed systems (including corporate managed boxes).
* The `install` and `deinstall` commands have `sudo` hardcoded into them. This is highly irregular and somewhat dangerous (users should `sudo make install` or `sudo make deinstall` if they want to run with elevated privileges, like they would most other open source distributions—the Makefile shouldn't do it for them).

I'll submit a pull request with fixes for these.

kaosat-dev:
@andrewgho  thanks for the input! the makefile will actually not be useful much longer: install & uninstall will be done like any other npm package via npm install xxx etc (no need for sudo anymore), please see the changes in the dev branch for more clarity on this : the makefile is still there but is more or less obsolete except for a few edge cases.

https://github.com/Spiritdude/OpenJSCAD.org/blob/dev/package.json

kaosat-dev:
I am closing this , as the makefile is fully obsolete and those paths are not used anymore, thanks for the PR back then @andrewgho :)

PR #228, opened 2017-02-26
chore(build): fix Issue #227: hardcoded bin path, sudo

andrewgho:
Also fixes paths for files in root directory, not js subdir.


andrewgho:
As mentioned in the commit message, I noticed while testing `make install` locally that the locations for the files copied by `make install` in the master branch are wrong—the files are presently checked in to the root directory, not in the `js` or `js/lib` subdirectories—so I fixed that, too, but let me know if that's a mistake and I can revert that portion.

PR #230, opened 2017-03-01
Add additional check for `lib` in format.js module require section.  …

johnwebbcole:
…While running in electron, the node module objects exist, but the global `lib` doesn’t, so the code fails attempting to require in modules.  Checking to see if `lib` exists before requiring in from the `lib` directory solves the electron issue.

### DIFF
--- a/formats.js
+++ b/formats.js
@@ -9,7 +9,7 @@ Copyright (c) 2011 Evan Wallace (http://evanw.github.com/csg.js/)
 Copyright (c) 2012 Alexandre Girard (https://github.com/alx)
 
 Exporting CSG into various formats:
-   - AMF 
+   - AMF
    - X3D
    - STL (ASCII & Binary)
 Exporting CAG into various formats:
@@ -22,7 +22,7 @@ License: MIT license
 
 // import the required modules if necessary
 
-if(typeof module !== 'undefined') {    // used via nodejs
+if(typeof module !== 'undefined' && typeof lib !== 'undefined') {    // used via nodejs
     CSG = require(lib+'csg.js').CSG;
     CAG = require(lib+'csg.js').CAG;
     Blob = require(lib+'Blob.js').Blob;
@@ -484,4 +484,3 @@ CAG.prototype.toSvg = function() {
     module.CSG = CSG;
     module.CAG = CAG;
 })(this);
-


kaosat-dev:
hi @johnwebbcole , and thanks for the fix !
Unfortunatly, the structure of the repo has changed quite significantly : please take a look at the dev branch [here](https://github.com/Spiritdude/OpenJSCAD.org/tree/dev) to see the new structure.
The good new is, I think the issue you mention should be already fixed, since it is now working flawlessly in node.js aswell.

johnwebbcole:
I've been following the new dev branch with excitement.  However since I'm using `trunk` and it has the issue, I was hoping an easy fix could be added. I can continue to monkey patch around it though.  It would be nice to have bug fixes and releases made for the 0.5.x versions for those people using it.  It would give us a chance to port over to the new vNext version when dev gets promoted.

kaosat-dev:
Yeah you are right, the in-between situation right now is not ideal, there has been some talk to have a regularly updated master with hot-fixes and a dev that is truly for dev, but to be honest that might need to wait until next release. Normally there are only a few things left to fix/update before the release, sorry for the complications in the meantime.

z3dev:
@johnwebbcole I don't have any issues with the change, as the change looks reasonable.

I have a feeling that a new release is about to happen. Can you wait?

johnwebbcole:
@z3dev I don't think it's a question of waiting :-)  Since I have a lot of existing JsCad code that could benefit from a fix and tagged release, it's a question of how much time do I want to spend upgrading projects to the new release.  Some I will, some I'd rather keep at 0.5.x.  

It would be very nice to have a tagged 0.5.x branch for future small fixes, like this one.  It would be even better to publish those to NPM along with the new release!

z3dev:
@johnwebbcole Sounds fine with me.

@Spiritdude Are you fine with a patch to 0.5.1?

kaosat-dev:
@Spiritdude @z3dev   sorry to bug you again, but it would be really important to include & tag this before we switch to the new version. Since this does not introduce any breaking changes I think we should just merge it

kaosat-dev:
@johnwebbcole guess what ? there is now a tagged release :)

z3dev:
A very good thing indeed...

PR #244, opened 2017-04-21
Fix include conflicts

kaosat-dev:
This PR combines the upgrade to the include() system done previously on the dev branch with those kindly provided by pr #239 by @jcoleman plus some needed fixes.
Edit : just to make sure it is clear, this includes & superseeds #239


kaosat-dev:
@jcoleman  nice, thanks ! At least I did not break that haha 👍

PR #247, opened 2017-04-22
Fixes to Correct Lost Functionality

z3dev:
Please review these fixes.
These correct functionality for:
- Resizing the view canvas when the window size changes
- Resting the viewer camera upon hitting the HOT key


kaosat-dev:
Perfect ! tried out both, all fixed 👍

Issue #274, opened 2017-05-28
XHR for remote resources


Spiritdude:
https://github.com/jscad/OpenJSCAD.org/commit/4a05755d0107491a18ca4eb1d613cf7fcc2d2014 broke remote URLs

kaosat-dev:
fixed in #275

PR #275, opened 2017-05-29
fix(remote loading)

kaosat-dev:
This PR adds a possible fix & upgrade for #274 
- add standard (normal) url param parsing instead of the '#' based remote document param 
- adds back and cleans up previous # + proxy
- proxy can be disabled by setting proxyURL to undefined
- removed url clearing so that parameters & uris stay and are clear
- fleshed out readme

This means you can use things like: 
- http://localhost:8081/?uri=http://www.thingiverse.com/download:164128 
- http://localhost:8081/?uri=examples/example001.jscad (works)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ as well CLI (command-line interface) for server-side computations with NodeJS.
 ## Table of Contents
 
 - [Usage](#usage)
+- [Development](#development)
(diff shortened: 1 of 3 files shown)


z3dev:
I use the remote.php sometimes. It seem to work better then the Perl version.

PR #278, opened 2017-06-20
Fix example code on README.md

jeromew:


kaosat-dev:
Thanks a lot for spotting and fixing this :)

Issue #281, opened 2017-07-04
MouseClick on parameters moves 3D View


gilboonet:
I'm adding parameters to a script and when I click on the parameter panel to change a value, it moves the 3D View. It also does the same when I click on the Code panel.. version 0.5.2 didn't have this bahaviour.

z3dev:
@gilboonet thanks

I found the same while testing but your explanation is much better.

kaosat-dev:
@gilboonet this issue is also present when moving /dragging the mouse inside the code editor:) (the new behaviour is my bad the old camera controls where full of hacks, so I changed quite a bit: ) 
one easy fix is to have the camera controls ONLY rotate/pan etc when your mouse is in the 3D VIEW and nowhere else (which makes the most sense to me imho)

hawkerm:
Yeah, I'm clicking in the code editor to edit my code and its rotating my 3d view, it's very confusing and distracting.  It's making the tool unusable for me. :(

z3dev:
Definitely a major issue.

@kaosat-dev do you have a quick solution? If not then I will fix this weekend.

gilboonet:
v 1.4.0 closes this

PR #286, opened 2017-08-10
fix(controls): changed camera controls to only work in 3d view

kaosat-dev:
solves annoying issues with camera rotation when editing code mentioned in #281


kaosat-dev:
Now works ok locally, mind trying this out @z3dev ? :)

z3dev:
@kaosat-dev The fix works just fine. I tested with Firefox, Safari, Chrome, and Opera. All work as expected. I did a quick test on the iPad as well with Safari, which also worked just fine.

I was about to look into this today as well. Thanks for the change.

kaosat-dev:
cool thanks a lot @z3dev !

PR #293, opened 2017-10-01
Limit URL to http or https

z3dev:
From Rene...

Here the current remote.pl 0.005 which limits to http/https (quits quitely if not) only and a few file extensions, to add new file extension should be simple. In case an invalid file is tried to download an JSON error message is return (which you might catch in the JS code and show it the user): { "error": "only amf, js .. supported" }

### DIFF
--- a/remote.pl
+++ b/remote.pl
@@ -2,9 +2,10 @@
 
 # -- Fetch Remote File into Cache, written by Rene K. Mueller <spiritdude@gmail.com>
 #
-$VERSION = '0.004';
+$VERSION = '0.005';
 #
 # History:
+# 2017/09/23: 0.005: fix vulnerability: disallow file://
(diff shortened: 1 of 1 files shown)


z3dev:
Should we add additional restrictions?

kaosat-dev:
I will double check and let you know @z3dev  !

kaosat-dev:
It is nice & simple so I think it is ok to merge as it is. Have you tried it 'live' though? (no server with Perl to test it on) @z3dev

z3dev:
This version is running on the website (as per Rene).
I cannot test unless I allow perl (cough). I'm a PHP purest.

kaosat-dev:
Haha well, we can merge it , and if it turns out there are issues, we can always revert !

PR #304, opened 2017-10-25
Do not reload the entire page on F5, run the normal rerendering process.

programmin1:
Whenever you hit F5 to rerender when your cursor happens to not be in the editor, it will reload the entire editor instead, losing your changes.

This fix fixes it, tested in Chrome and Firefox.

Happy Hacktoberfest :)


kaosat-dev:
Much appreciated @programmin1 ! Will take this for a spin and merge !

Issue #316, opened 2017-11-08
request: please make viewer menu open/close on click instead of hover


donhatch:
On the main website page, the viewer menu open/close on hover is clever, but I find it makes the page significantly less usable and less pleasant for me.

To enumerate my problems with it:
  - the menu opens frequently when I don't want it to
  - the menu closes frequently while I'm trying to read it
  - the menu opens-and-closes every time the mouse pointer crosses over it, which is useless and distracting
  - it's impossible to keep it open to read it while I am trying out the varous hotkeys it describes, or while I am comparing it with other documentation
  - it's impossible to keep the Examples submenu open while I select various examples and interact with some of them, or while I compare it with other documentation

How about changing it to an explicit open/close on click, just like the editor?
I've changed it to that in my local repo, and I am much happier.

kaosat-dev:
Hi @donhatch , lots of good points !
I think:
- We would need to keep the initial 'default to open then close after a short while' behaviour to signify its presence
- add the explicit on click behaviour you mention

Btw , I am less than in love with the whole current UI, so there will likely be a major overhaul at some point in the future, even if that is not high on the list so far.

donhatch:
Hi Mark, thanks for considering this.
Yes, retaining the initial 'default to open then close after a short while' sounds fine.

Clicking should probably cancel the initial timeout (so it doesn't automatically close after the user has explicitly opened it, which would be unfriendly).

PR #318, opened 2017-11-08
Implement request #316, and minor ui fixes.

donhatch:
- Implement request #316: viewer menu now opens/closes on handle click instead of hover, just like editor.
- Menu handle icon now changes on open/close as appropriate.  Added necessary image files.
- Default is still open, auto-closing after 3 seconds.
- Explicit open or close cancels the initial auto-close timeout.
- Since Examples doesn't close on hover-out any more, there needs to be a way to close it, so made clicking on Examples toggle it instead of just opening it.
- Fix: Clicking on Examples now cancels the initial auto-close timeout.
- Fix: The initial auto-close now closes Examples too, if open.
- Fix: Editor handle icon now changes on open/close as intended (the icon was getting set on the wrong element before)


z3dev:
@donhatch Wow! Thanks for,the nice changes.

Just for reference, which browsers are you testing with?

kaosat-dev:
Very nice work @donhatch  , much appreciated !
I'll do a full review/ local test tonight.

donhatch:
Hi @z3dev,
I tested this using chrome 61.0.3163.100 and safari 10.0,
both on a macbook pro running macos 10.12 (sierra).

kaosat-dev:
@donhatch I tried your PR locally, and I like it a lot !
But one thing is still weird : unless you explicitly click on the handle it still opens 'on hover'
I think it would feel more natural if after the initial closing after the timeout, it would only open/close with explicit clicks.

donhatch:
Hi @kaosat-dev, I agree it shouldn't open on hover any more; I'm surprised if it's still doing that for you.
I don't think it's doing that for me, on the browsers I mentioned above.
And I thought I removed the lines that implement open-on-hover, in style.css, didn't I?
So it's very strange if it's still doing it for you; are you sure?

kaosat-dev:
Hi @donhatch double checked after your message & turns out it is the Chrome's caching that was messing things up! 
Testing in Chrome, Safari and Firefox on MacOs.
Will merge asap

src/ui/index.js
```diff
@@ -86,32 +86,42 @@ function init () {
     const versionText = `Version ${version}`
     document.getElementById('menuVersion').innerHTML = versionText
 
-    setTimeout(function () {
+    let initialMenuHidingTimeoutID = setTimeout(function () {
+      initialMenuHidingTimeoutID = null;
```
kaosat-dev: The only issue I see before merging is the use of semicolons here and there.
Sorry but as per our CONTRIBUTING.md , we use [standardjs style](https://standardjs.com/)
(we will likely add some automated linting like in the IO package very soon)

src/ui/index.js
```diff
@@ -86,32 +86,42 @@ function init () {
...
 
       // -- Examples
       examplesTitle.addEventListener('click', function (e) {
-        setElementHeight(examples, 'auto')
-        examples.style.display = 'inline'
-      // setElementHeight(options, 0)
-      // options.style.display = 'none'
+        if (initialMenuHidingTimeoutID !== null) {
+          clearTimeout(initialMenuHidingTimeoutID);
+          initialMenuHidingTimeoutID = null;
```
kaosat-dev: ditto

src/ui/index.js
```diff
@@ -125,6 +135,29 @@ function init () {
           list[i].addEventListener('click', onLoadExampleClicked)
       }
     }
+
+    let menuHandle = document.getElementById('menuHandle')
+    menuHandle.addEventListener('click', function (e) {
+      if (initialMenuHidingTimeoutID !== null) {
+        clearTimeout(initialMenuHidingTimeoutID);
```
kaosat-dev: etc

src/ui/index.js
donhatch: fixed.

src/ui/index.js
donhatch: fixed.

src/ui/index.js
donhatch: fixed

donhatch:
@kaosat-dev , I fixed the semicolons.
I guess my two commits should be squashed into one-- I'm not sure whether that's something that I should do somehow, or whether you will do it during the pull?

kaosat-dev:
@donhatch thanks ! It is fine , we always use 'squash and merge' strategy, so we can ensure a single commit per PR and standardised final commit names (that get used by tooling down the line for things such as changelog generation, etc)

PR #319, opened 2017-11-10
Fix and clarify npm link instructions in README.md.

donhatch:
- Changed `npm link ../csg.js/ @jscad/csg` to `npm link ../csg.js`.
  I'm not sure what the old recipe was supposed to mean (`npm help link`
  doesn't mention that usage), but it gave me errors when I tried it
  (npm v3.10.10).  The new recipe is what the help page describes,
  and it works for me.
- Changed the `git clone` urls to the ones advertised on the respective
  github pages, which use https (the old ones weren't working for me).
- Added explicit recipe for making the main repo live-updating (`npm link`),
  and separated out the csg.js-specific commands into a distinct section
  for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,8 +205,8 @@ the easiest method is to use ```npm link``` to have a 'live' updating developmen
 - clone this repository ```git clone https://github.com/jscad/OpenJSCAD.org.git```
 - go into OpenJSCAD.org folder ```cd OpenJSCAD.org```
 - install dependencies ```npm install```
-- start dev server if desired: ```npm run start-dev```
-- make it so global ```openjscad``` refers to local code: ```npm link```
+- if desired, make the ```openjscad``` command refer to the code in this folder: ```npm link```
+- if desired, start dev server: ```npm run start-dev```
 
 Then, for example for CSG.js:
 - go back to base directory ```cd ..```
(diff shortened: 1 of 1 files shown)


kaosat-dev:
Very useful set of changes, thanks @donhatch !
The only one I was on the fence about is the change from git to https, but since that is less hassle, and works behind proxies, I am ok with it.

README.md
```diff
@@ -201,17 +201,18 @@ but you can also rebuild them manually if you need :
...
-
-For example for CSG.js
 - create a base directory
-- clone this repository ```git clone git@github.com:jscad/OpenJSCAD.org.git```
+- clone this repository ```git clone https://github.com/jscad/OpenJSCAD.org.git```
 - go into OpenJSCAD.org folder ```cd OpenJSCAD.org```
 - install dependencies ```npm install```
-- start dev server : ```npm run start-dev```
+- start dev server if desired: ```npm run start-dev```
+- make it so global ```openjscad``` refers to local code: ```npm link```
```
kaosat-dev: I think this can be rephrased a bit : something like 
if desired, make the ```openjscad```command refer to the code in this folder

README.md
```diff
@@ -201,17 +201,18 @@ but you can also rebuild them manually if you need :
...
+
+Then, for example for CSG.js:
```
kaosat-dev: Much clearer, good !

README.md
donhatch: I like that.  That also makes it clear that the `npm link` isn't a prereq for the `npm run start-dev`, even if I reverse the order of the two steps.   So I reversed them, which I think is the more natural order.

PR #335, opened 2017-12-29
plumb GL options through viewer

danmarshall:
My goal is to be able to use canvas.toDataURL() from the viewer. 

To do that, we need to be able to pass `{preserveDrawingBuffer: true}` as the `options` to

```gl = canvas.getContext('webgl', options);``` [here](https://github.com/jscad/OpenJSCAD.org/blob/master/src/ui/viewer/lightgl.js#L1137)

This PR is the plumbing which passes the gl options through. So we can create a `new Viewer(viewerdiv, { glOptions: {preserveDrawingBuffer: true} })`

### DIFF
--- a/src/ui/viewer/jscad-viewer-lightgl.js
+++ b/src/ui/viewer/jscad-viewer-lightgl.js
@@ -43,7 +43,7 @@ LightGLEngine.prototype = {
   },
   createRenderer: function () {
     // Set up WebGL state
-    var gl = GL.create()
+    var gl = GL.create(this.options.glOptions)
     this.gl = gl
     this.gl.lineWidth(1) // don't let the library choose
 
--- a/src/ui/viewer/jscad-viewer.js
+++ b/src/ui/viewer/jscad-viewer.js
@@ -15,7 +15,8 @@ function Viewer (containerelement, options) {
   if ('plate' in options) { this.setPlateOptions(options['plate']) }
   if ('axis' in options) { this.setAxisOptions(options['axis']) }
   if ('solid' in options) { this.setSolidOptions(options['solid']) }
-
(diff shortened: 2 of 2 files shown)


z3dev:
I did the same thing at Z3D.
I think this can be the default setting as well, as browsers and computers are a tad faster these days. :)

danmarshall:
@z3dev - I'm not too familiar with this, but my understanding is that it still may bog down mobile devices.

kaosat-dev:
This is a good addition, totally forgot that I used to add this by default as well.
@danmarshall  are you ok with these changes being somewhat short-lived ? The new viewer is right around the corner, and I will add similar settable options so this is good either way :)

danmarshall:
@kaosat-dev yes I'm ok with that :)

BTW is there a place I can add feature requests for the new viewer? I'd like to separate the gesture handling (I want to control the camera without user mouse input) or use a different pointer events lib.

Also, I am exploring a bug where the viewer doesn't like to be not in full screen mode. It sometimes disappears when scrolled in the page.

kaosat-dev:
@danmarshall I just open sourced an early version of the new viewer [here](https://github.com/jscad/csg-viewer), however, please add issues, feature requests etc in THIS repo with the 3D viewer label, to keep things simple as the repo/org structure might change.
I will be honest the viewer is opinionated, but has a lot of features built in (you can set the camera manually etc)
I will likely split it up into multiple utility repositories so that they can be reused 'to taste' but not right away.

danmarshall:
Ok thanks Mark, I added an issue #336 but I was not able to add the label. I think only admins can add labels. I'm excited to try out what you have so far!

kaosat-dev:
weird , I thought it was for all , gotta check the rights.
In the meantime this can be merged I think.

PR #338, opened 2017-12-30
ability to generate multiple variations from interactive parameter examples

jtoy:
### All Submissions:

### DIFF
--- a/src/cli/multiopenjscad.js
+++ b/src/cli/multiopenjscad.js
@@ -1,0 +1,79 @@
+#!/usr/bin/env node
+
+#a simple script that reads an interactive params example and generates every variation
+#to use this:  node multiopenjscad.js /path/to/gears.jscad
+#you must have the packages mkdirp and eval installed to use this.
+const fs = require('fs')
+const path = require('path')
+const {execSync} = require('child_process')
+var _eval = require('eval')
+var mkdirp = require('mkdirp');
+
+
+const args = process.argv.splice(2)
+if(args.length < 1){
+  console.log("pass file to parse")
+  process.exit()
+}
+
+inputFile = args[0]
+ftype = inputFile.split("/").slice(-1)[0].split(".")[0]
+console.log("class: "+ftype)
+let src = fs.readFileSync(inputFile, 'UTF8')
+src += "\ngetParameterDefinitions()\n"
+eval(src)
+function to_array(json){
+  array = []
+  if(json.hasOwnProperty('step')){
+    step = json['step']
+  }else{
+    step = 1
+  }
+  for(var i = json['min'];i <= json['max'];i+= step){
+    array.push(i)
+  }
+  if(array.length > 0){
+    return array
+  }else {
+    return [json['initial']]
+  }
+}
+
+getGetParameterDefinitions = new Function(src + ";\nreturn getParameterDefinitions();");
+data = getGetParameterDefinitions();
+json = {}
+for ( var i = 0; i < data.length; i++) {
+  json[data[i]['name']] = to_array(data[i])
+}
+combos = Object.keys(json).reduce((acc, key) => {
+  newArray = [];
+
+  json[key].forEach(item => {
+    if (!acc || !acc.length) { // First iteration
+      newArray.push({[key]: item});
+    } else {
+      acc.forEach(obj => {
+        newArray.push({...obj, [key]: item});
+      });
+    }
+  });
+
+  return newArray;
+}, []);
+jscadPath = "openjscad"
+console.log("total variations: "+combos.length)
+for(var i=0;  i < combos.length; i++){
+  params = ""
+  folder = ""
+  // TODO sort params so that the sha is always correct
+  Object.keys(combos[i]).forEach(function(key) {
+    params += ` --${key} ${combos[i][key]} `
+    folder += `${key}_${combos[i][key]}__`
+  })
+  outputPath = `objects/${ftype}/${folder}/`
+  mkdirp.sync(outputPath)
+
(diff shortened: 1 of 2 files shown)


kaosat-dev:
Thanks for this @jtoy  ! I will take it for a spin today.
Not quite sure if it should be a separate file though, but I will try out.

kaosat-dev:
@jtoy taking a closer look at the code: the script creates a different stl output for every possibly parameter combo right ?

jtoy:
yes, that is how I am able to generate variations versions. I am using this code in my production system now.

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env node
+
+#a simple script that reads an interactive params example and generates every variation
+#to use this:  node multiopenjscad.js /path/to/gears.jscad
+#you must have the packages mkdirp and eval installed to use this.
+const fs = require('fs')
+const path = require('path')
+const {execSync} = require('child_process')
+var _eval = require('eval')
```
kaosat-dev: why import and then not use ?

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
...
+if(args.length < 1){
+  console.log("pass file to parse")
+  process.exit()
+}
+
+inputFile = args[0]
+ftype = inputFile.split("/").slice(-1)[0].split(".")[0]
+console.log("class: "+ftype)
+let src = fs.readFileSync(inputFile, 'UTF8')
+src += "\ngetParameterDefinitions()\n"
```
kaosat-dev: prefer es6 templates

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
...
+    array.push(i)
+  }
+  if(array.length > 0){
+    return array
+  }else {
+    return [json['initial']]
+  }
+}
+
+getGetParameterDefinitions = new Function(src + ";\nreturn getParameterDefinitions();");
```
kaosat-dev: there are already tools in jscad for extracting parameter definitions for CLI : any reason not to use them?

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
...
+console.log("total variations: "+combos.length)
+for(var i=0;  i < combos.length; i++){
+  params = ""
+  folder = ""
+  // TODO sort params so that the sha is always correct
+  Object.keys(combos[i]).forEach(function(key) {
+    params += ` --${key} ${combos[i][key]} `
+    folder += `${key}_${combos[i][key]}__`
+  })
+  outputPath = `objects/${ftype}/${folder}/`
```
kaosat-dev: prefer path creations with require('path').join() as it is cross platform

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env node
+
+#a simple script that reads an interactive params example and generates every variation
+#to use this:  node multiopenjscad.js /path/to/gears.jscad
+#you must have the packages mkdirp and eval installed to use this.
+const fs = require('fs')
+const path = require('path')
+const {execSync} = require('child_process')
+var _eval = require('eval')
+var mkdirp = require('mkdirp');
```
kaosat-dev: add a dependency, can it be avoided ?

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
...
+src += "\ngetParameterDefinitions()\n"
+eval(src)
+function to_array(json){
+  array = []
+  if(json.hasOwnProperty('step')){
+    step = json['step']
+  }else{
+    step = 1
+  }
+  for(var i = json['min'];i <= json['max'];i+= step){
```
kaosat-dev: the script requires min and max values to be specified for all params right ?
in other words : the params must be finite ?

src/cli/multiopenjscad.js
```diff
@@ -0,0 +1,79 @@
...
+}
+combos = Object.keys(json).reduce((acc, key) => {
+  newArray = [];
+
+  json[key].forEach(item => {
+    if (!acc || !acc.length) { // First iteration
+      newArray.push({[key]: item});
+    } else {
+      acc.forEach(obj => {
+        newArray.push({...obj, [key]: item});
```
kaosat-dev: please avoid semicolons , as we use [standard](https://standardjs.com/) styling :)

kaosat-dev:
@jtoy  ok, does that mean that all params need to be finite? (ie, min/max)
Also did you not run into any excessive computation time issues?
This script could be very useful for pre-computing geometry for different param sets to have a 'cached' version (tradeoff of speed vs storage).

I added a review of the code, please do not take the remarks badly!

src/cli/multiopenjscad.js
kaosat-dev: also favor let over var

jtoy:
Yes, the min and max need to be given, probably should write some notes or not run if no min/max given.
no offense taken, I don't write a lot of javascript.

src/cli/multiopenjscad.js
jtoy: hi sorry for the late reply, im working on this now. I figured this existed, where can I find this piece?

z3dev:
@jtoy no offense, but i would like to close the pull request without merging.

if you still need this functionality the please make the same changes again branch V2.
thanks

z3dev:
V1 pull requests are no longer accepted. Please try again with V2 compatible changes.

PR #360, opened 2018-03-05
feat(modules support): add node.js modules support to  cli

kaosat-dev:
This PR adds support for using designs defined as node.js modules (ie with require() & module.exports etc) as inputs to the command line

- it also cleans up & add some common code in the core package
- very minor naming tweak in the web package


z3dev:
I just remembered something... UPPER CASE file extensions don’t work. For example, myfile.STL There are some programs out there that do this still.

kaosat-dev:
which file is that ? (in uppercase ?)

feel free to copy & paste to the wiki of course :) but I like clear and self contained READMEs with all needed information
if you mean the whole part about node modules: that will need to be fleshed out & expanded uppon, and put into the main README & wiki for sure (once I add the 'fake' require() to the web variant in particular, since then we will have 100% useability )

PR #381, opened 2018-04-21
Fix over double quotation symbol (")

hassadee:
Fix over double quotation symbol (") at "autoreload" ID at line 170


z3dev:
Ummm... you changed every single line in the HTML file. How are we suppose to know what changed?

Can you supply the lines numbers of the changes?

hassadee:
Sorry for modify tab indent in every single line of the HTML file.

The line number of the change is 170

z3dev:
Thanks. I saw that change, but was concerned about others.

Will take a serious look now.

z3dev:
A few comments...

You need to respect the intention of each file. We don’t use 4 space intention. Please use 2 space intention, and update the files again.

Second, if this PR is for the Thai language support then change the name, and description.

hassadee:
Hmm, I use 2 space for tab indent in Visual Studio Code, don't know why its appear as 4 space indent here. I think I will create new PR instead.

PR #383, opened 2018-04-24
HTML syntax and performance improvement

hassadee:
1. Small changes with space and missing double quote symbol that skip browser to read the footer path.
2. Add rel="noopener" at <a> element with target="_blank" for performance improving when user open an external link in a new window or tab, more detail: https://developers.google.com/web/tools/lighthouse/audits/noopener

P.S. This time should be fine. 😆


z3dev:
Super! Looks good.

kaosat-dev:
Hi @hassadee , thanks for the contribution ! 
I learned something about links from this :)
Just a heads up : I will merge this but the new version of the jscad-web will be very different from the current one, so your contribution will end up in a bit of different structure (html wise) down the line, hope you do not mind !

PR #425, opened 2019-01-09
feat(reuseable Regl renderer): 

kaosat-dev:
Work in progress DO NOT MERGE ! 
This PR adds an updated Regl Renderer (improved upon csg viewer's) as a reusable package.

- reduced dependencies, simplified internals and much more readable API
- works for web, desktop and *cli* ! 
- includes example code for cli use

### All Submissions:


kaosat-dev:
Hi @danmarshall !
Would love to have your feedback about the general API, please : (I will still likely do a few minor tweaks but it is 99% there I think)
https://github.com/jscad/OpenJSCAD.org/blob/regl-renderer/packages/utils/regl-renderer/demo-cli.js
Is this something you could use easily for the gallery (or any other use case) ?
I will also add a demo/example for web use , for completeness sake, though it will be very similar.

danmarshall:
Hi @kaosat-dev - this is super awesome!! I suppose I would do it from within a node.js script, and not the command line. So, I would just use your code as a starting point. Very excited about this!

kaosat-dev:
Hi @danmarshall oh , I mis-remembered, so you where running the image generation in node ? my bad!
Then yes, this is even closer to your use case :) 
Do you see anything off/ weird, annoying in the example ? The main thing I am not sure off is if the various visual attributes (color, visibility etc) should perhaps be regrouped into a sub object like geometry & transforms.

danmarshall:
@kaosat-dev No worries, even I needed to refresh my own memory 🤓.
I'm not sure if these options:
```
  drawCommands: {
    // draw commands should bootstrap themselves the first time they are run
    drawGrid: require('./src/rendering/drawGrid/index.js'),
    drawAxis: require('./src/rendering/drawAxis'),
    drawMesh: require('./src/rendering/drawMesh/index.js')
  },
```
will be available from a different package?

kaosat-dev:
@danmarshall  haha I am saved then  :)
Very good point about the draw commands
The idea was to have a very small amount of built in ones  I guess they could either:
- be exposed as part of the module's api as exports , which would be something like 
`{
  render,
  cameras,
  drawCommands: {
     drawGrid,
     drawAxis,
      drawMesh
  } 
}`

with this you can compose your draw commands, using the built ins if you need : 
```javascript
const makeMyDrawCmd = (regl, params) => {
const drawMesh = makeDrawMesh()
const drawAxis = makeDrawAxis()
const drawFoo = makeDrawFoo()
  return () => {
      drawMesh(),
      drawAxis()
      drawFoo()
  }
}
```


- be built in but hidden ie not part of the api, so you would be able to use them for your entities's drawing but not have a reference to them .
This makes for less exposed objects, but that would mean you cannot compose the built in draw commands

BarbourSmith:
I know this is still very alpha and subject to change, but I am really excited about it. 

Is there a way that I can play around with https://github.com/jscad/OpenJSCAD.org/blob/regl-renderer/packages/utils/regl-renderer/demo-web.js in the browser? What would the HTML side look like?

kaosat-dev:
@BarbourSmith sorry that is still old code from a previous version, updating things asap :)

kaosat-dev:
@BarbourSmith check the files now, the web demo is updated :) 
you can run the demo micro server with `npm run dev` 
This will not have a standalone build to include directly as a script tag, as it only renders a single frame, so you would want a combo of this + camera controls I think ?

BarbourSmith:
You are a hero! This project is so exciting, and the pace at which it is moving forwards is astounding. Thank you!

PR #456, opened 2019-07-27
Japanese Language Support (only)

z3dev:
This pull request adds 'japanese' to the list of languages (core), as well as, provides a japanese locale file (web). Adjustments to the existing locales were made to support this language as well.

This pull request slightly adjusts the web views (options and shortcuts) improving both the layout and locale support.

### All Submissions:


packages/web/locales/de.json
```diff
@@ -1,41 +1,65 @@
...
-  "zoom to fit on design load":"Alles Zoom beim laden von Designs",
-  "settings storage path":"Dateipfad für Einstellungen",
-  "3d viewer": "3d viewer",
-  "(not settable)": "(nicht verränderbar)",
-  "theme": "Thema",
+
   "STL (Binary)":"STL (binär)",
   "STL (ASCII)":"STL (ASCII)",
-  "AMF (experimental)":"AMF (experimentell)",
+  "AMF (experimental)":"AMF",
```
kaosat-dev: it should still say experimental in German :)

packages/web/locales/en.json
```diff
@@ -1,41 +1,65 @@
...
+  "options.filehandling.converttoscript":"Convert external formats to JSCAD scripts",
+
+  "options.shortcuts.h3":"Keyboard Shortcuts",
+
+  "command":"Command",
+  "keybinding":"Key Stroke",
+  "when":"When",
+  "typeandenter":"type and hit enter",
+
+  "notsettable":"(not settable)",
```
kaosat-dev: parens

packages/web/locales/en.json
```diff
@@ -1,41 +1,65 @@
...
-  "german":"german",
-  "french":"french",
+  "language":"english",
+
+  "english":"English",
+  "german":"German",
+  "french":"French",
+  "japanese":"Japanese",
+
+  "options.language.h3":"Languages",
```
kaosat-dev: ok so after going through this a few time I found what seems 'off' : translations are for UI text strings, the whole 'options' thing  seems like a sort of json in string and does not seems odd.
Can you explain your reasoning please ?

packages/web/locales/fr.json
```diff
@@ -1,41 +1,65 @@
...
+  "options.filehandling.converttoscript":"Convert external formats to JSCAD scripts",
+
+  "options.shortcuts.h3":"Racourcis Clavier",
+
+  "command":"Commande",
+  "keybinding":"Racourci",
+  "when":"Quand",
+  "typeandenter":"type and hit enter",
+
+  "notsettable": "(non modifiable)",
```
kaosat-dev: parens

packages/web/locales/ja.json
```diff
@@ -0,0 +1,66 @@
+{
```
kaosat-dev: This is so awesome  ! :) Thanks for adding this!

packages/web/locales/ja.json
```diff
@@ -0,0 +1,66 @@
...
+  "right":"右側",
+
+  "setProjectionType":"View type",
+  "perspective":"",
+  "orthographic":"",
+
+  "reEvaluateCode":"Reload script",
+
+  "always":""
+}
```
kaosat-dev: hmm is the language prefix for japanese not **jp** and no **ja** (I mean for the file name ja.json)

packages/web/src/ui/views/options.js
```diff
@@ -6,60 +6,60 @@ module.exports = function options (state, i18n) {
...
-    <legend> <h3> ${i18n`language`}</legend>
+    <legend> <h3> ${i18n`options.language.h3`} </h3> </legend>
     <select id='languageSwitcher'>
       ${languages}  
     </select>
   </fieldset>
 
   <fieldset>
-    <legend> <h3> ${i18n`theme`} </h3> </legend>
+    <legend> <h3> ${i18n`options.theme.h3`} </h3> </legend>
```
kaosat-dev: not repeating , but same question as above about the use of xxxDOTxxxDOT for translation strings

packages/web/src/ui/views/shortcuts.js
```diff
@@ -7,7 +7,7 @@ module.exports = function shortcuts (state, i18n) {
 
     // if we are in the midst of assigning a keypress (inProgress) display 'type & hit enter'
     // if we already have a temporary key use that one instead
-    const placeholder = (tmpKey && tmpKey.length > 0) ? tmpKey : i18n.translate('type & hit enter')
+    const placeholder = (tmpKey && tmpKey.length > 0) ? tmpKey : i18n.translate('typeandenter')
```
kaosat-dev: same question as previously, to me default translation 'keys' where mean to be human centric , and not 'variables' but I might have been wrong, got very little experience

packages/web/src/index.js
```diff
@@ -48,7 +48,8 @@ async function makeJscad (targetElement, options) {
     translations: {
       en: require('../locales/en.json'),
       fr: require('../locales/fr.json'),
-      de: require('../locales/de.json')
+      de: require('../locales/de.json'),
+      ja: require('../locales/ja.json')
```
kaosat-dev: yeaaah ! :)

packages/web/locales/de.json
z3dev: Why experimental? It's totally supported.
Or are we saying that AMF is experimental?
Maybe we should say "obsolete"?

packages/core/sideEffects/i18n/index.js
```diff
@@ -59,7 +60,9 @@ const makei18nSideEffect = (options) => {
     getAvailableLanguages: command => {
       const availableLanguages = Object.keys(translations)
         .map(code => {
-          const fullName = longNames[code] ? longNames[code] : 'placeholder'
+          let fullName = longNames[code] ? longNames[code] : 'placeholder'
+          let translation = translations[code]
+          if (translation && 'language' in translation) fullName = translation.language
```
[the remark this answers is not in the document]
z3dev: Actually, there’s no need for the hard coded list of languages. The list can be read from the locale files. I added logic to use the ‘language’ found in the locale file.

packages/web/locales/de.json
```diff
@@ -1,41 +1,65 @@
 {
-  "language": "Sprache",
+  "language":"german",
+
   "english":"englisch",
   "german":"deutsch",
   "french":"französich",
+  "japanese":"",
+
+  "options.language.h3":"Sprache",
```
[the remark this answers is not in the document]
z3dev: HEADER

To be consistent, we should split a few ‘keys’ now.

The major keys would be ...
- ‘error’
- ‘option’ or ‘setting’
- ‘viewer’ or something

Basically, the keys are context, not text. If a single word/phrase is translated then just a simple ‘key’

packages/web/locales/ja.json
z3dev: Nope. It’s ‘ja’

https://en.m.wikipedia.org/wiki/List_of_ISO_639-1_codes

packages/web/src/ui/views/options.js
z3dev: I’m using the ‘.’ to separate words, as this reads easier.

packages/web/src/ui/views/shortcuts.js
z3dev: I’ve seen both. But as soon as you add three options (previously two) then the default ‘key’ becomes meaningless. And, you have to change both code and locale files. This is a HUGE problem for maintenance, and always causes mistakes.

CONTEXT is more important for the ‘keys’

packages/web/locales/de.json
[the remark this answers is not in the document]
kaosat-dev: Hmm I see what you are shooting for but I honestly find it adds complexity for little gain: do you have an example where the same key would have a different string based on 'context' ?

packages/web/locales/de.json
kaosat-dev: hmm good point: I thought the AMF parser was never finsihed ?
I am ok with dropping the 'experimental' , and AMF is not obsolete, there is just the 3MF format which is very similar, but AMF is still in use

packages/core/sideEffects/i18n/index.js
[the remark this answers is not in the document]
kaosat-dev: so the hardcoded 'longNames' can be completely removed ? very nice ! go for it ! :)

packages/web/locales/ja.json
kaosat-dev: ok, I stand corrected :)

packages/web/src/ui/views/shortcuts.js
z3dev: ok. let's compromise a little.

the keys (strings) inside the code should be very simple:
- short as possible english phrases
- no special characters (parens, exclamations, etc)

the locales (key value pairs) can contain anything
- the best phrase for translation
- can include parens, exclamations, etc

z3dev:
REBOOT!

I changed the keys to be short english phrases. The phrases do not contain any special characters or special phrases, as those can be added to the locales if required to communicate.

@kaosat-dev Please veiw the changes to keys in src/ui/views.

packages/web/locales/de.json
```diff
@@ -1,44 +1,44 @@
...
 
-  "options.language.h3":"Sprache",
-  "options.theme.h3":"Thema",
+  "Languages":"Sprache",
+  "Themes":"Thema",
 
-  "options.generation.h3":"Generieren von Geometrie",
-  "options.generation.caching":"Experimenteles caching von Geometrie,",
-  "options.generation.timeout":"Abfallzeit für generieren von Soliden (ms)",
+  "Generation":"Generieren von Geometrie",
```
kaosat-dev: nice !

kaosat-dev:
Awesome @z3dev ! Looks good to me ! Any other changes or can I merge ?

z3dev:
@kaosat-dev please merge

PR #458, opened 2019-07-27
Fix (CORE) transforming of external formats to scripts

z3dev:
This fixes issues when converting external formats (STL, DXF, etc) to JSCAD scripts.

### All Submissions:

### DIFF
--- a/packages/core/code-loading/transformSources.js
+++ b/packages/core/code-loading/transformSources.js
@@ -12,24 +12,41 @@ const modulifyTransform = (options, entry) => {
   return Object.assign({}, entry, { source })
 }
 
-const translateToJscad = (options, entry) => {
-  // console.log('translateToJscad', entry)
-  const { apiMainPath } = options
-  const deserializeStl = require('@jscad/io').stlDeSerializer.deserialize
-  const source = deserializeStl(entry.source, entry.name, options)
-  // const ext = 'js'
-  // const name = entry.name.split('.') + ext
-  // const fullPath = entry.fullPath.split('.') + ext
-  return Object.assign({}, entry, { source })
+const transformDxfToJscad = (options, entry) => {
+  // console.log('***** transformDxfToJscad',options,entry)
+  const deserialize = require('@jscad/io').dxfDeSerializer.deserialize
+  const source = deserialize(entry.source, entry.name, options)
+
+  const ext = 'jscad'
+  const name = entry.name.substring(0, entry.name.lastIndexOf('.') + 1) + ext
+  const fullPath = '/' + name
+
+  return Object.assign({}, entry, { ext, name, fullPath, source })
+}
+
+const transformStlToJscad = (options, entry) => {
+  console.log('***** transformStlToJscad',options,entry)
+  const deserialize = require('@jscad/io').stlDeSerializer.deserialize
+  const source = deserialize(entry.source, entry.name, options)
+
+  const ext = 'jscad'
+  const name = entry.name.substring(0, entry.name.lastIndexOf('.') + 1) + ext
+  const fullPath = '/' + name
+
+  return Object.assign({}, entry, { ext, name, fullPath, source })
 }
 
 const transformSources = (options, filesAndFolders) => {
-  // console.log('transformSources', options, filesAndFolders)
+  console.log('***** transformSources', options, filesAndFolders)
   const transformsPerFormat = {
     'js': [modulifyTransform],
     'jscad': [modulifyTransform],
-    'stl': [translateToJscad, modulifyTransform],
-    'amf': [translateToJscad, modulifyTransform]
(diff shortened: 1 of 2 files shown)


z3dev:
FYI, there still some changes for the other external formats required. But I would like feedback on these changes first.

packages/core/code-loading/transformSources.js
```diff
@@ -12,24 +12,63 @@ const modulifyTransform = (options, entry) => {
...
-  const deserializeStl = require('@jscad/io').stlDeSerializer.deserialize
-  const source = deserializeStl(entry.source, entry.name, options)
-  // const ext = 'js'
-  // const name = entry.name.split('.') + ext
-  // const fullPath = entry.fullPath.split('.') + ext
-  return Object.assign({}, entry, { source })
+/*
+ * Create a new entry for a script (JSCAD) from the given entry and source
+ */
+const createJscadEntry = (entry, source) => {
```
kaosat-dev: very nice  !

packages/core/code-loading/transformSources.js
```diff
@@ -12,24 +12,63 @@ const modulifyTransform = (options, entry) => {
...
+  const ext = 'jscad'
+  const name = entry.name.substring(0, entry.name.lastIndexOf('.') + 1) + ext
+  const fullPath = '/' + name
+
+  return Object.assign({}, entry, { ext, name, fullPath, source })
+}
+
+const transformAmfToJscad = (options, entry) => {
```
kaosat-dev: these are all great !
With a big catch, please read my comment in the overral review

z3dev:
> * do we use CSG based deserialization ?

Wow! I’m not sure... so I suggest we start by allowing the user to choose.

In the website, I guess this means another ‘option’.

Maybe this also all applies to the application.

For the CLI, maybe there could be more granularity, allowing the user to choose which formats to convert into what.

z3dev:
By the way. I like the ‘registration’ of extensions in the core, which is sourced from the IO library. But it’s hard coded to ‘csg’ now. Maybe this is where the new ‘option’ can be useful, allowing ‘jscad’ or ‘csg’

kaosat-dev:
> > * do we use CSG based deserialization ?
> Wow! I’m not sure... so I suggest we start by allowing the user to choose.
> In the website, I guess this means another ‘option’.
> Maybe this also all applies to the application.
> For the CLI, maybe there could be more granularity, allowing the user to choose which formats to convert into what.

I will be honest Jeff, the issue is even more complex than that, a simple option is not enough because in certain cases it would not work (one file vs multiple, external vs internal editor etc). For now I think we should leave it as is, and add a note/ issue for it, because it is HARD to solve (I went through quite a few iterations trying to solve it, and none was satisfactory).
The best solution I found so far:
- if you drop or work on a **SINGLE** file, then do a text/jscad conversion and display that
- if you drop **multiple files or a FOLDER** then use csg as you are likely (as a user) to want to work on a full project where you can import your stl/amf/whatever files as CSG using something like ```shape = require('./mySTL.STL')

kaosat-dev:
> By the way. I like the ‘registration’ of extensions in the core, which is sourced from the IO library. But it’s hard coded to ‘csg’ now. Maybe this is where the new ‘option’ can be useful, allowing ‘jscad’ or ‘csg’

sorry not sure I follow ?

z3dev:
> > By the way. I like the ‘registration’ of extensions in the core, which is sourced from the IO library. But it’s hard coded to ‘csg’ now. Maybe this is where the new ‘option’ can be useful, allowing ‘jscad’ or ‘csg’
> 
> sorry not sure I follow ?

OK. I looks like the webrequire is calling a routine call 'registerAllExtensions' found in core/io

This seems like the correct way to go, rather then having each application build another 'registration' of formats.

I really think this should be part of the IO library though, as that library is the source of the deserializers / serializers. It would be trivial to make this the source of the external 'format' information as well.

kaosat-dev:
> I really think this should be part of the IO library though, as that library is the source of the deserializers / serializers. It would be trivial to make this the source of the external 'format' information as well.

Sorry but no :) this USES the code from IO to handle the format, but the registration process is specific to the 'core'/ main script evaluation process.

z3dev:
> Sorry but no :) this USES the code from IO to handle the format, but the registration process is specific to the 'core'/ main script evaluation process.

Understood. The core decides how to handle each request to process the external format. But the information about each format should come from the IO library.

```
  amf: {
    displayName: 'AMF (experimental)',
    description: 'Additive Manufacturing File Format',
    extension: 'amf',
    mimetype: 'application/amf+xml',
    convertCSG: true,
    convertCAG: false },
```

kaosat-dev:
The format description that you mention is already in IO :) (formats.js)

PR #473, opened 2019-09-20
V2 STL Serializer

z3dev:
This pull request updates the STL serializer to use CSG V2 API. The goals of this pull request are:
- [X] Update to use CSG V2 API
- [X] Update code up to JSCAD coding standards
- [X] Add test suites to complete coverage, 98.02%

### All Submissions:


z3dev:
@kaosat-dev please review. any comments would be welcome.

packages/io/stl-serializer/CSGToStla.js
```diff
@@ -1,50 +1,46 @@
-const { ensureManifoldness } = require('@jscad/io-utils')
-const {isCSG} = require('@jscad/csg')
+const { geometry } = require('@jscad/modeling')
 
-// objects must be an array of CSG objects (with polygons)
+// objects must be an array of 3D geomertries (with polygons)
```
kaosat-dev: Small note: perhaps in the future we should specify that they are of type Geom3 (or any similar things)

packages/io/stl-serializer/CSGToStla.js
```diff
@@ -1,50 +1,46 @@
...
+  polygons.forEach((polygon, i) => {
+    result.push(convertToFacet(polygon))
   })
   return result.join('\n')
 }
 
-const vector3DtoStlString = (v) => {
-  return `${v._x} ${v._y} ${v._z}`
-}
+const vector3DtoStlString = (v) => `${v[0]} ${v[1]} ${v[2]}`
```
kaosat-dev: perhaps since now we use array-like data, it would make more sense for a string representation to also match that ?

packages/io/stl-serializer/README.md
```diff
@@ -1,16 +1,14 @@
 ## @jscad/stl-serializer
 
-> stl serializer for the jscad project (from CSG)
+> Serializer of CSG geometries to STL mesh
```
kaosat-dev: no more CSG :)

packages/io/stl-serializer/index.js
```diff
@@ -1,59 +1,43 @@
...
-  objects.forEach(function (object, i) {
-    if (isCSG(object) & object.polygons.length > 0) {
-      csgs.push(ensureManifoldness(object))
-    }
-  })
-  return options.binary ? serializeBinary(csgs, options) : serializeText(csgs, options)
+  objects = utils.flatten(objects)
+
+  // convert only 3D geometries
+  objects = objects.filter((object) => geometry.geom3.isA(object))
```
kaosat-dev: smart !
Should we also emit a warning if anything other than a geom3 was passed ?
something like '5 geom2, 2 path2 objects ignored'

packages/io/stl-serializer/index.js
```diff
@@ -1,59 +1,43 @@
...
+
+  if (objects.length === 0) throw new Error('the given objects cannot be converted')
```
kaosat-dev: A bit more information would be helpfull in the message imho:
'No valid 3d geometry found to be serialized to stl data`

packages/io/stl-serializer/tests/ascii.test.js
```diff
@@ -1,39 +1,19 @@
 const test = require('ava')
-const { CSG } = require('@jscad/csg')
-const serializer = require('./index.js')
 
-test('serialize csg to stl (binary)', function (t) {
-  const csg1 = new CSG.cube()
-  const observed = serializer.serialize({ binary: true }, csg1)
+const { primitives, transforms } = require('@jscad/modeling')
```
kaosat-dev: where did the binary test go ? :)

kaosat-dev: aah never mind, saw the other file

kaosat-dev: why split them though ?

packages/io/stl-serializer/tests/progress.test.js
```diff
@@ -0,0 +1,17 @@
+const test = require('ava')
```
kaosat-dev: nice ! 
Very good addition
although I believe 3 files for 3 tests is overkill , mostly as they are small tests

packages/io/stl-serializer/tests/ascii.test.js
z3dev: The both ascii and binary test need improvement, which should increase the number of test cases, and the lines of code. So, I just split now in preparation.

packages/io/stl-serializer/CSGToStla.js
z3dev: I'll leave this for now. There's still room for improvements in the code.

packages/io/stl-serializer/index.js
z3dev: Maybe, but I'm not a big fan of verbose logging of warnings.
If so then all the serializer / deserializer functions should be consistent.

kaosat-dev: the problem is that silence failure is even worse: if the user does not know what happened or why something did not have the expected output, it is hard for them to debug

packages/io/stl-serializer/tests/ascii.test.js
kaosat-dev: not sure I follow ? why do we need more tests ?

z3dev: As the note says in the binary test...

// TODO: VERY shallow testing ... improve

packages/io/stl-serializer/index.js
z3dev: @kaosat-dev back to you. I added a warning if some objects were not serialized. How does that look?

packages/io/stl-serializer/index.js
```diff
@@ -33,11 +33,12 @@ const serialize = (options, ...objects) => {
   objects = utils.flatten(objects)
 
   // convert only 3D geometries
-  objects = objects.filter((object) => geometry.geom3.isA(object))
+  let objects3d = objects.filter((object) => geometry.geom3.isA(object))
 
-  if (objects.length === 0) throw new Error('only 3D geometries can be serialized to STL')
+  if (objects3d.length === 0) throw new Error('only 3D geometries can be serialized to STL')
```
kaosat-dev: Hmm this works, although I wonder if we should throw an error or not ?
I keep wondering if returning Error or Warning objects instead of throwing might be be better for these kind of things (also replacing console.warn etc), this way you always get some useful data out of serializers ??
Meh, let's keep it like you did for now, we can always change later

packages/io/stl-serializer/tests/ascii.test.js
kaosat-dev: fair point

PR #478, opened 2019-09-29
Changed image-utils to use `writeFileSync` instead of piping to a str…

johnwebbcole:
…eam.

Streams are async, and without returning the string,
a promise or using a callback, you cannot wait on
the stream response.

I'm using this in unit tests, and need to wait for
the image to be written before comparing it to
a snapshot image.

Should this need to be an async function, it can be
fixed differently.

@jscad/img-utils and @jscad/regl-renderer need to be published to NPM.

### All Submissions:

I think the tests will pass once these packages are published.

### DIFF
--- a/packages/utils/img/package.json
+++ b/packages/utils/img/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jscad/img-utils",
-  "version": "1.0.0",
+  "version": "0.0.0",
   "description": "",
   "main": "src/index.js",
   "scripts": {},
--- a/packages/utils/img/package.json
+++ b/packages/utils/img/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jscad/img-utils",
-  "version": "0.0.0",
+  "version": "1.0.0",
   "description": "",
   "main": "src/index.js",
   "scripts": {},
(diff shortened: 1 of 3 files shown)


kaosat-dev:
Hi @johnwebbcole , thanks for the fix ! 
Two aspects of this:
- the change to writeFileSync makes sense in this case, agreed ! 
- the packages however are still in pre-alpha and are not supposed to be published to NPM quite yet:
> @jscad/img-utils and @jscad/regl-renderer need to be published to NPM. did you mean this for the unit tests to pass on CI, or in general for usability ?

johnwebbcole:
I think that changing the relative links to package links breaks the tests right now during instal (the img-utils package is not found).

Do you think it's possible to publish a 0.0.1-1 pre-release version?  As long as sem-ver is followed when updated, it should be ok for dependent packages to start using parts of the V2 system.

z3dev:
Coming very soon... I asked the same. Hopefully, we can publish an ‘alpha’ version soon.

kaosat-dev:
I am ok with trying to release a 0.0.1 version, but the error will not be solved, as this seems to indicate a lerna issue , as local deps should be resolved correctly, npm or not
Will take it for a spin locally.
Oh btw , '**latest**', should not be used in dependency specifications, that causes issues (we have been bitten by that quite a lot in the past :) ), we now always use exact versions (works well, is reliable and controllable)

PR #484, opened 2019-11-22
fix(V2): small tweaks and fixes for V2

kaosat-dev:
Small sets of fixes for blocker bugs before further changes :)

### All Submissions:

### DIFF
--- a/packages/utils/regl-renderer/src/geometry-utils/entitiesFromSolids.js
+++ b/packages/utils/regl-renderer/src/geometry-utils/entitiesFromSolids.js
@@ -56,7 +56,7 @@ const entitiesFromSolids = (params, solids) => {
     const visuals = {
       drawCmd: 'drawMesh',
       show: true,
-      color: meshColor,//[0.8, 0.5, 0.7, 1.0],
+      color: meshColor,
       transparent: geometry.isTransparent, // not sure
       useVertexColors: true
     }
--- a/packages/utils/regl-renderer/src/geometry-utils/entitiesFromSolids.js
+++ b/packages/utils/regl-renderer/src/geometry-utils/entitiesFromSolids.js
@@ -56,7 +56,7 @@ const entitiesFromSolids = (params, solids) => {
     const visuals = {
       drawCmd: 'drawMesh',
       show: true,
-      color: [0.8, 0.5, 0.7, 0.1],
+      color: meshColor,//[0.8, 0.5, 0.7, 1.0],
       transparent: geometry.isTransparent, // not sure
       useVertexColors: true
     }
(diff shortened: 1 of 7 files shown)


packages/utils/regl-renderer/src/geometry-utils/entitiesFromSolids.js
```diff
@@ -56,7 +56,7 @@ const entitiesFromSolids = (params, solids) => {
     const visuals = {
       drawCmd: 'drawMesh',
       show: true,
-      color: [0.8, 0.5, 0.7, 0.1],
+      color: meshColor,//[0.8, 0.5, 0.7, 1.0],
       transparent: geometry.isTransparent, // not sure
```
z3dev: I think this is the correct change. Why leave the comment with the old color?

packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
```diff
@@ -71,7 +71,7 @@ const drawMesh = function (regl, params = { extras: {} }) {
 
       func: { src: 'src alpha', dst: 'one minus src alpha' }
     },
-    primitive: (context, props) => props && props.primitive ? props.primitive : 'triangles'
+    primitive: (context, props) => props && props.primitive ? props.primitive : (type === '3d' ? 'triangles' : 'lines')
```
z3dev: Hmm... I understand the change but wonder if the real correction should be made when primitives is called, I.e making sure that props.primitive is set correctly.

packages/web/src/ui/flow/design.js
```diff
@@ -103,7 +103,7 @@ const reducers = {
  * @returns {Object} the updated state
  */
   setDesignContent: (state, payload) => {
-    console.log('design: set content', state.design, payload)
+    console.log('design: set content', state, state.design, payload)
```
z3dev: I think console messages should be commented out for the final PR.

packages/utils/regl-renderer/src/geometry-utils/entitiesFromSolids.js
kaosat-dev: agreed

packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
kaosat-dev: that is a different thing, props.primitive is an override, and this sets the **default** for both 2D & 3D

packages/web/src/ui/flow/design.js
kaosat-dev: agreed, but that should be done at a later stage once we are truely in cleanup mode ;)

PR #485, opened 2019-11-30
refactor(): V2 csg integration

kaosat-dev:
WORK IN PROGRESS DO NOT MERGE !!


z3dev:
Why keep those old unused code in some strange directory?
We can recover piece by piece if require from the original CSG repository.

lerna.json
```diff
@@ -1,11 +1,12 @@
...
   "packages": [
+    "packages/core",
+    "packages/modeling",
     "packages/cli",
     "packages/desktop",
     "packages/web",
     "packages/examples",
-    "packages/core",
     "packages/io/*",
     "packages/utils/*"
```
z3dev: This should also have "packages/ui/*" added.

z3dev:
After reading the NPM and the LERNA documentation on managing version numbers, I think we should change everything to prerelease versions (like 0.0.0-alpha). This will help us manage the versions, and make sure we don't accidentally publish new versions of existing V1 packages.

lerna.json
kaosat-dev: no, they are not included for a good reason: some of the dependencies in there are NOT compatible with CI (travis or others), and cannot be included this way.
I spent ages finding workarounds, but nothing worked so far.
We can revisit later.

kaosat-dev:
> After reading the NPM and the LERNA documentation on managing version numbers, I think we should change everything to prerelease versions (like 0.0.0-alpha). This will help us manage the versions, and make sure we don't accidentally publish new versions of existing V1 packages.
> 
> @kaosat-dev thoughts?

I would agree IF we were not using automated versioning :)
The whole 'upgrading versions of packages' is automated based on commit messages + very specific release commands and should NEVER be done manually.
The risk of mistakes is very low, we will just need to add BREAKING Changes in the final V2 merge (jscad uses the **conventional commits system**, and even then, there is a dry-run operation to ensure no accidental publishing happens.

z3dev:
> should NEVER be done manually

Well... someone added the ‘-alpha’ suffix to several modules. And that was a manual change.
I’m suggesting that we do this to all modules until a release (of some sort) is ready.

As for CCS, I will start to follow the guidelines for commit messages. However, as the CCS FAQ mention, making a mistake is not the end of the world.

PR #492, opened 2019-12-26
V2 : Cleanup of Test Scripts

z3dev:
This is a small set of changes to enable automated testing across all modules.

The changes are:
- update dependencies to latest V1 and V2 (there were some old ones)
- change the timeout to two minutes(2), as there is a lot of concurrency
- changes to json-serializer (still V1) to complete testing

And a some tweaks to the project package to excute all tests. Or to excute one test.

### All Submissions:


z3dev:
Node.js 13 build is NOT building. I don't think this can be fixed. Why include Node.js 13 now?

z3dev:
Node.js 6 is NOT building. I think we should drop Node.js 6 now.

packages/io/package.json
```diff
@@ -1,11 +1,10 @@
 {
-  "name": "@jscad/io",
+  "name": "@jscad/io-base",
```
kaosat-dev: we should not be changing package names unless there is a good reason to do so :)

packages/io/amf-deserializer/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "devDependencies": {
     "ava": "^0.25.0",
-    "nyc": "^10.3.2"
+    "nyc": "^10.3.2",
+    "@jscad/sample-files": "github:jscad/sample-files"
```
kaosat-dev: I understand the logic but this has a serious side effect: it means that every time you install 
amf-serializer => IO => jscad you pull in the whole samples repo/package which is hugge

package.json
```diff
@@ -4,9 +4,11 @@
   "description": "",
   "repository": "https://github.com/Spiritdude/OpenJSCAD.org",
   "scripts": {
-    "test": "npm i && npm run bootstrap && lerna run test",
+    "test-one": "lerna run --stream --scope '@jscad/json-deserializer' test",
```
kaosat-dev: what is the purpose of this one ?
is this a temporary thing ?

packages/io/amf-deserializer/package.json
z3dev: I was trying to find a simple way to initiate tests for a single module. This would be very handy for making bug fixes to any of the modules.

Of course, the tests can be initiated from the specific directory. But this feels more ‘built-in’.

Not sure as to which is the best approach.

package.json
z3dev: I was trying to find a simple way to initiate tests for a single module. This would be very handy for making bug fixes to any of the modules.

Of course, the tests can be initiated from the specific directory. But this feels more ‘built-in’.

Not sure as to which is the best approach.

packages/io/amf-deserializer/package.json
z3dev: I agree. I’m still scratching my head on this one.

The tests for DXF deserializer run just fine.

Do you see any differences?

packages/io/package.json
z3dev: I understand. But both packages/io and packages/io/io have the same name. It’s very confusing, so I changed the name here, as this is a ‘private’ module.

Thoughts?

package.json
kaosat-dev: initialise tests ? as in running all the boilerplate / setup code ?

packages/io/amf-deserializer/package.json
kaosat-dev: hmm I need to double check this, because you are right AND we should not have this dependency, will dig deeper in the next few days, have a bit more time available.

packages/io/package.json
kaosat-dev: err no , packages/io is a folder
packages/io/io is the root package

z3dev: Good. I’m beginning to understand.
So, why does this file exist? (packages/io/package.json)

packages/io/amf-deserializer/package.json
z3dev: The latest CI build works! I may have found a solution. :)
I'll continue to monitor as there are a few more changes to post.

package.json
z3dev: removed

packages/io/amf-deserializer/package.json
z3dev: Seems to work just fine now. :)

packages/io/package.json
kaosat-dev: I double checked and I **think** you are right, that file should not exist ! 
I hope I have not missed anything, but I see no reason to keep it :) Good find !

z3dev: removed

z3dev:
@kaosat-dev please reveiw again. i believe this is ready to merge now, but let me know if further changes should be made.

kaosat-dev:
@z3dev looking good ! 
Going to merge it, and tackle the node V13 issues :)

PR #494, opened 2020-01-08
V2 : Fixes to CLI

z3dev:
These are some minor fixes to the CLI, as not all supported formats were enabled.

Also, an additional error message is logged if a project entry point is not found.

### All Submissions:

### DIFF
--- a/packages/cli/src/determineOutputNameAndFormat.js
+++ b/packages/cli/src/determineOutputNameAndFormat.js
@@ -1,11 +1,13 @@
+const extensionReg = /\.(amf|dxf|js|jscad|obj|stl|svg|x3d)$/
+const formatReg = /(amf|dxf|js|jscad|stl|stla|stlb|svg|x3d)/i
 
 const determineOutputNameAndFormat = (outputFormat, outputFile, inputFile) => {
-  if (!outputFormat && outputFile && outputFile.length && outputFile.match(/\.(jscad|js|stl|amf|dxf|svg)$/)) { // output filename set
+  if (!outputFormat && outputFile && outputFile.length && outputFile.match(extensionReg)) { // output filename set
     outputFormat = RegExp.$1
   } else if (!outputFormat && outputFile && outputFile.length) { // output filename isn't valid
     console.log('ERROR: invalid output file <' + outputFile + '>')
     process.exit(1)
-  } else if (outputFormat.match(/(jscad|js|stl|stla|stlb|amf|dxf|svg)/i)) { // output format defined?
+  } else if (outputFormat.match(formatReg)) { // output format defined?
     let ext = RegExp.$1
     if (!outputFile) { // unless output filename not set, compose it
       ext = ext.replace(/stl[ab]/, 'stl') // drop [ab] from stl
@@ -13,7 +15,7 @@ const determineOutputNameAndFormat = (outputFormat, outputFile, inputFile) => {
       outputFile = outputFile.replace(/\.([^\.]+)$/, '.' + ext) // compose output filename
     }
   } else {
-    console.log(`ERROR: invalid output format <${outputFormat}`)
+    console.log(`ERROR: invalid output format <${outputFormat}>`)
     process.exit(1)
   }
   return { outputFormat, outputFile }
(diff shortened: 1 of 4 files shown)


packages/cli/src/determineOutputNameAndFormat.js
```diff
@@ -1,19 +1,21 @@
+const extensionReg = /\.(amf|dxf|js|jscad|obj|stl|svg|x3d)$/
+const formatReg = /(amf|dxf|js|jscad|stl|stla|stlb|svg|x3d)/i
 
 const determineOutputNameAndFormat = (outputFormat, outputFile, inputFile) => {
-  if (!outputFormat && outputFile && outputFile.length && outputFile.match(/\.(jscad|js|stl|amf|dxf|svg)$/)) { // output filename set
+  if (!outputFormat && outputFile && outputFile.length && outputFile.match(extensionReg)) { // output filename set
```
kaosat-dev: much better ! thanks !

packages/cli/src/parseArgs.js
```diff
@@ -1,21 +1,19 @@
...
-  //
-  // process.argv.forEach(function (val, index, array) {
-  //  console.log(index + ': ' + val)
-  // })
   if (args.length < 1) {
     console.log('USAGE:\n\nopenjscad [-v] <file> [-of <format>] [-o <output>]')
-    console.log('\t<file>  :\tinput (Supported types: folder, .jscad, .js, .scad, .stl, .amf, .obj, .gcode, .svg, .json, .dxf)')
-    console.log('\t<output>:\toutput (Supported types: folder, .jscad, .stl, .amf, .dxf, .svg, .json)')
-    console.log("\t<format>:\t'jscad', 'stla' (STL ASCII, default), 'stlb' (STL Binary), 'amf', 'dxf', 'svg', 'json'")
+    console.log('\t<file>  :\tinput (Supported types: folder, .jscad, .js, .amf, .dxf, .obj, .stl, .svg)')
```
kaosat-dev: since we have a flat list of formats would it not make sense to auto generate these console outputs from the list ?

packages/cli/src/parseArgs.js
```diff
@@ -66,24 +62,27 @@ const parseArgs = args => {
...
-      console.log('OpenSCAD Compatibility (' + version + ')')
     } else {
       inputFile = args[i]
       if (fs.statSync(inputFile).isDirectory()) {
         inputIsDirectory = true
-        inputFile = args[i]
         // get actual design entry point if applicable (if passed a folder as input etc)
         inputFile = getDesignEntryPoint(fs, inputFile)
+        if (!inputFile) {
+          console.log("ERROR: could not determin entry point of project.")
```
kaosat-dev: small typo

packages/cli/src/parseArgs.js
```diff
@@ -66,24 +62,27 @@ const parseArgs = args => {
...
+          console.log("Verify main or index exists")
+          process.exit(1)
+        }
         inputFormat = require('path').extname(inputFile).substring(1)
       } else {
         console.log('ERROR: invalid file name or argument <' + args[i] + '>')
-        console.log("Type 'openjscad' for help")
+        console.log("Type 'openjscad' for a list of supported types")
```
kaosat-dev: hmm does the cli always return a list of supported types by defaults ?

packages/cli/src/parseArgs.js
z3dev: done

packages/cli/src/parseArgs.js
z3dev: i took a shot are generating lists of input extensions, output extensions, and output formats from the formats. see core/io/formats.js

the list of formats now controls what is deserializable as input, as well as what can be serialized as output.

the cli now uses those lists to log those lists, as well as build regular expressions.

packages/cli/src/parseArgs.js
z3dev: it does after the latest changes. :)

z3dev:
by the way, i want to drop gcode support as well. this deserializer has never worked well, and crashes almost always. i have never touched the code as well. we should just KILL GCODE

kaosat-dev:
well done @z3dev ! much more coherent , about gcode: we can just flag it as not update/ unuseable so somebody can/ will pick it up & overhaul it sole day ?

z3dev:
> about gcode: we can just flag it as not update

Ok. It’s disabled in the new list of formats. But the IO module needs a little face lift to use V2, I think. I’ll put that on my list of things to do.

PR #495, opened 2020-01-08
V2 : Fix AMF Deserializer

z3dev:
UG.. Missed this one!

These changes correct translate() to provide V2 require and export statements.

### All Submissions:

### DIFF
--- a/packages/io/amf-deserializer/index.js
+++ b/packages/io/amf-deserializer/index.js
@@ -19,6 +19,7 @@ History:
 //
 // //////////////////////////////////////////
 
+const version = require('./package.json').version
 const translate = require('./translate')
 const instantiate = require('./deserialize')
 
@@ -35,7 +36,7 @@ const instantiate = require('./deserialize')
 const deserialize = (input, filename, options) => {
   const defaults = {
     output: 'jscad',
-    version: '0.0.0',
+    version: version,
     addMetaData: true
   }
   options = Object.assign({}, defaults, options)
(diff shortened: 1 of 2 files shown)


packages/io/amf-deserializer/index.js
```diff
@@ -35,7 +36,7 @@ const instantiate = require('./deserialize')
 const deserialize = (input, filename, options) => {
   const defaults = {
     output: 'jscad',
-    version: '0.0.0',
+    version: version,
```
kaosat-dev: in modern js, you do not need to repeat the keys, so this should be 
```
{ output: 'jscad',
 version,
addMetaData: true
  }
```

packages/io/amf-deserializer/index.js
```diff
@@ -19,6 +19,7 @@ History:
 //
 // //////////////////////////////////////////
 
+const version = require('./package.json').version
```
kaosat-dev: hmm this is nice & simple but it changes the significance of the 'version'  compared to previous versions (ie jscad version versus deserializer version), I think it is ok though :)

kaosat-dev: we should just make that clear

packages/io/amf-deserializer/index.js
z3dev: Understood. Let's catch these via lint. There are plenty.

packages/io/amf-deserializer/index.js
z3dev: good idea. i changed the metadata information about the producer, specifically saying JSCAD IO library. i'll update the other IO libraries as well.

PR #499, opened 2020-01-16
V2 : Rewrite of reTesselateCoplanarPolygons

z3dev:
**WORK IN PROGRESS. DO NOT MERGE.**

This is a total rewrite of the reTesselateCoplanarPolygons() function. This version does not create new points, instead merges opposing polygons using the original points.. As such, the precision of all points is respected (This also means the other boolean functions must improve.)

Improved test suites for reTesselateCoplanarPolygons() are included.

### All Submissions:


z3dev:
@kaosat-dev this is a BIG one. please look over the code when you have time. all comments are welcome.

kaosat-dev:
@z3dev sure ! looks hefty ! will take a close look

kaosat-dev:
first very broad comments: (I only had a very superficial glance so far)
- looks very clean, kudos !
- > This version does not create new points, instead merges opposing polygons using the original points

I understand the logic behind what you are shooting for, but I  think that could cause some serious issues: normally a **(re)tesselation** whole purpose is to generate **more / different triangles,** and thus points. Merging multiple pre-existing points/polygons is a different operation. (I cannot recall the name right now, but there was something like this in V1 I think ?)

kaosat-dev:
I could be wildly off based on the function name, so I need to dig deeper though

z3dev:
I can see why the name of retesselate was chosen. It’s because the old algorithm transformed all the points to 2D, and then built polygons by merging edges. This is the definition of tessellation.

However, any transform of points causes precision errors. And going from 3D to 2D to 3D causes bad things to happen.

z3dev:
I'm closing this pull request, as this can be addressed AFTER the release of V2.

PR #509, opened 2020-01-28
V2: move convertToBlob from core/io to io-utils for coherence & practicality

kaosat-dev:
### All Submissions:

@z3dev : this is a small one, but long overdue, and fixes a few issues I had locally


kaosat-dev:
ah dang, based off of wrong branch (includes examples changes), will redo

kaosat-dev:
@z3dev this is ready for review : 
a bit of background : 
- it moves the bulk of core/io (particularly convertToBlob & related code) to io itself
- since it created circular dependencies (core<->io<->core), I extracted all array utilities into a new package, that also allowed to get rid of a ton of duplicate code !
- related cleanups

packages/utils/array-utils/index.js
```diff
@@ -30,4 +30,20 @@ const nth = (index, data) => {
   return data[index]
 }
 
-module.exports = {toArray, head, flatten, nth}
+/* function flatten (array) {
```
z3dev: See the flatten in modeling, which is recursive. I think that’s a better implementation.

z3dev:
I found another package using a specific toArray() function... svg-deserializer

kaosat-dev:
> can the movement from ‘core to IO’ be split from the new array package (I like this)?
No, sadly, because as stated above just the move results in circular dependencies.

kaosat-dev:
> I found another package using a specific toArray() function... svg-deserializer

good catch ! will update

packages/utils/array-utils/index.js
kaosat-dev: Ah yes, btw I did not touch the array things in modeling because there are other people doing changes in modeling, and we can always make modeling import array utils.
Btw about recursive vs non recursive : perhaps we should have both: I remember there was a case where a single level of flattening was expected (I cannot recall the exact location, sorry :( )

We could provide a 'flattenDeep' or 'flattenNested' or a parameter ?

kaosat-dev:
@z3dev should be all good

z3dev:
> @z3dev should be all good

I didn’t have time to review in full. I’ll finish tomorrow morning, which should be while you are sleeping. 😃

packages/io/io/package.json
```diff
@@ -52,6 +52,8 @@
     "@jscad/stl-serializer": "1.0.0-alpha.d0fb3056",
     "@jscad/svg-deserializer": "1.0.0-alpha.d0fb3056",
     "@jscad/svg-serializer": "1.0.0-alpha.d0fb3056",
-    "@jscad/x3d-serializer": "1.0.0-alpha.d0fb3056"
+    "@jscad/x3d-serializer": "1.0.0-alpha.d0fb3056",
+    "@jscad/modeling": "0.0.1",
```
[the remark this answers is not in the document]
kaosat-dev: formats.js has been moved from core to io (same coherence) , and that one needs modeling

z3dev:
@kaosat-dev nice last commit. 👍🏻 

You can merge. I’m very satisfied with this.

kaosat-dev:
thanks @z3dev ! Merging !

PR #510, opened 2020-02-04
V2 : Remove GCODE support

z3dev:
These changes removed support for GCODE conversions, formats, etc.

There are also a few fixes to enable CLI and WEB
- WEB : updated code to require array utilities from @jscad/array-utils
- CLI : fixed export of prepareOutput

### All Submissions:


z3dev:
FYI, I have not removed the GCODE deserializer code. The old code is kind of worthless, as it doesn't meet any of the V2 standards, etc, etc.

kaosat-dev:
@z3dev lots of good things, will review later tonight.
I think it makes sense to KEEP the code around, but add a big note in the README, something like 'deprectated, if you feel like doing a new version, please do ' :D

PR #512, opened 2020-02-05
V2 remove center on primitives

garrlker:
### All Submissions:

* [X] Does your submission pass tests? -- Mostly, 3 tests fail


packages/modeling/src/primitives/roundedCuboid.js
```diff
@@ -4,7 +4,8 @@ const {vec2, vec3} = require('../math')
 
 const {geom3, poly3} = require('../geometry')
 
-const createCorners = (center, size, radius, segments, slice, positive) => {
```
z3dev: You need to change this back, as createCorners() is an internal function only. The center parameter is key to the generation of the polygons.

packages/modeling/src/primitives/star.js
```diff
@@ -12,7 +12,8 @@ const getRadiusRatio = (vertices, density) => {
   return 0
 }
 
-const getPoints = (vertices, radius, startAngle, center) => {
+const getPoints = (vertices, radius, startAngle) => {
```
z3dev: Please revert this change as getPoints() is an internal routine. The center is used when creating the 2D points.

packages/modeling/src/primitives/roundedCuboid.js
garrlker: Done, this did fix the broken test as well.

packages/modeling/src/primitives/star.js
garrlker: Done, this did fix the broken test as well.

z3dev:
Good. Good. Those changes helped.

The tests are only failing for expand. I’ll take a detailed long at the code tomorrow.
(FYI, I live in Japan)

z3dev:
@garrlker please adjust the code in operations/expansions/expandShell.js, line 200
this function uses sphere() to build as set of corners for the 3D expansion, and uses the center option. a change is required.

```
    let corner = sphere({
      center: [vertex[0], vertex[1], vertex[2]],
      radius: delta,
      segments: segments,
      axes: [xaxis, yaxis, zaxis]
    })
```

z3dev:
@garrlker great stuff. I think we are getting close.

I suspect that the dxf-deserializer is using center. This maybe a little more difficult.

garrlker:
I did find where dxf was using a center param and I changed it to use the function but it's still failing.

I'll look at it some more tomorrow.

z3dev:
See...

packages/io/dxf-deserializer/instantiate.js
packages/io/dxf-deserializer/translate.js

z3dev:
Also SVG deserializer... See...

packages/io/svg-deserializer/shapesMapJscad.js
packages/io/svg-deserializer/shapesMapGeometry.js

kaosat-dev:
@garrlker thanks for the awesome hard work ! Getting there :)
@z3dev & @garrlker , I feel stupid for pointing this out now so late, but our center() function's api is horrible : ```center({center: value}, operands)``` should be ```center(value, operands)``` (or even better ```centerAt(value, operands)```
Since that is another seperate change It is perhaps best to leave it for another PR though ?

z3dev:
> center() function's api is horrible : `center({center: value}, operands)` should be `center(value, operands)` (or even better `centerAt(value, operands)`
> Since that is another seperate change It is perhaps best to leave it for another PR though ?

Another PR, I think.

z3dev:
@garrlker still there?

DXF-deserializer and svg-deserializer need changes. Can you finish this week?

garrlker:
@kaosat-dev 
No problem!
I think another PR would be best :smile: 

@z3dev 
Sorry, I had a busy weekend and wasn't able to hop on.
I've pushed those changes, but the `path2toDxf.test.js` LWPOLYLINE test is still failing.

When I skip that test all others pass, so I think there is a good shot of finishing this up by the end of this week!

z3dev:
> I've pushed those changes, but the `path2toDxf.test.js` LWPOLYLINE test is still failing.

Ah... polylines... ok. There’s probably a small change required. I’ll try to locate the lines needing change today..

z3dev:
Some changes are required to io/dxf-serializer/tests/path2ToDxf.test.js 

This test uses center()

garrlker:
I think I already converted that file's tests to use the correct `center` API.

z3dev:
OK. The problem is that arc() creates an arc which has a center at the [x,y] provided. This is NOT the same as centering the shape at [x,y].  So, these two statements are NOT the same.

```
  let path2 = primitives.arc({ center: [5, 5], endAngle: 45 })
  let path2 = transforms.center({ center: [5,5,0] }, primitives.arc({ endAngle: 45 }));
```

@kaosat-dev any thoughts about this use of the center option?

garrlker:
That makes sense.

I'll roll back the `arc` changes.

kaosat-dev:
@z3dev & @garrlker : agreed & my 2 cents is, for arc, the meaning of center is not quite the same so it is ok to keep the center option there, and roll that part back (sorry @garrlker !)

garrlker:
It's no problem!
I rolled back the arc changes and all tests are passing.

z3dev:
> I rolled back the arc changes and all tests are passing.

Awesome. And thanks!
We will do a quick review.

z3dev:
@kaosat-dev please review.

PR #514, opened 2020-02-08
V2 : Enable coverage reports

z3dev:
These changes enables coverage reports for packages with test suites.

The coverage reports can be viewed by opening the "coverage_reports.html" file in the project directory.

### All Submissions:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 node_modules/
 components/
-converage/
+coverage/
 .DS_Store
 .nyc_output
 packages/web/examples
--- a/coverage_reports.html
+++ b/coverage_reports.html
@@ -1,0 +1,28 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <title>Links to Coverage Reports</title>
+  </head>
+
+  <body>
+    <h1>Links to Coverage Reports</h1>
+    <p>The following coverage reports are available after running 'npm run coverage'</p>
+    <table class="summary">
+      <tr><td class="link"><a href="packages/core/coverage/index.html">core</a></td></tr>
+      <tr><td class="link"><a href="packages/io/amf-deserializer/coverage/index.html">io/amf-deserializer</a></td></tr>
+      <tr><td class="link"><a href="packages/io/amf-serializer/coverage/index.html">io/amf-serializer</a></td></tr>
+      <tr><td class="link"><a href="packages/io/dxf-deserializer/coverage/index.html">io/dxf-deserializer</a></td></tr>
+      <tr><td class="link"><a href="packages/io/dxf-serializer/coverage/index.html">io/dxf-serializer</a></td></tr>
+      <tr><td class="link"><a href="packages/io/io-utils/coverage/index.html">io/io-utils</a></td></tr>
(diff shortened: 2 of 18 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 node_modules/
 components/
-converage/
+coverage/
```
kaosat-dev: doh ! nice catch!

package.json
```diff
@@ -35,6 +36,7 @@
     "@jscad/sample-files": "github:jscad/sample-files#master",
     "ava": "1.4.0",
     "lerna": "3.19.0",
+    "nyc": "^15.0.0",
```
kaosat-dev: we should always use exact versions (so without the caret "^" symbol); as not exact versions of dependencies can lead to some issues (breakeage, unanonced and not correct api changes etc)

coverage_reports.html
```diff
@@ -0,0 +1,28 @@
+<!doctype html>
```
kaosat-dev: hmm why do we need this html file ? or even the coverage_reports folder at all ? (and we even .gitignore the coverage folder ?)

z3dev: I added the generation of HTML to the coverage reports, which allows some very nice browsing of coverage per file. However, the HTML is located in the package specific ‘coverage’ folder.

That’s why I added this index to the packages. I thought you may have questions about this. 😃 

Can you try out generating the reports, and browse a little?

package.json
z3dev: done

coverage_reports.html
kaosat-dev: haha I'll take it for a spin locally to generate reports and take a look tomorow then !

kaosat-dev:
@z3dev tried to run this today, but sadly, as I feared , it does not work unless you add nyc as a dependency of all the sub packages (I think it worked for you because you had a version with nyc already installed in there, you will likely have the same issue if you run **lerna clean** before running the coverage command).

I understand why you wanted to remove the dependency from each individual package, but I fear it is not that trivial (I tried to do this a while back with ava , which is also a similar devdependency) and failed

kaosat-dev:
there is a lerna command to achieve something like that, trying to solve it now 
https://github.com/lerna/lerna/blob/master/doc/hoist.md

kaosat-dev:
aha ! Found a solution ! Shall I edit your PR a tiny bit @z3dev  ? it is just a minor change in the main package.json

kaosat-dev:
@z3dev  ?

z3dev:
> aha ! Found a solution ! Shall I edit your PR a tiny bit @z3dev ?

@kaosat-dev Please amend.

z3dev:
@kaosat-dev looks good.

I thought you might use hoist. Do you know if lerna.json can be used for hoist as well?

kaosat-dev:
@z3dev , hmm good question, the docs are not clear on the subject at all : does it matter if it is in the command or lerna config ?

z3dev:
Nope. I’m fine. We are both learning a lot. 😃

z3dev:
@kaosat-dev please merge.

PR #521, opened 2020-02-16
V2 : fix transforms

z3dev:
These changes allow various length arrays to be passed to:
- mirror
- rotate
- scale
- translate

NOTE: center() was not changed, as that should be done seperately.

### All Submissions:


z3dev:
@kaosat-dev please review. let me know if anything should be changed.

z3dev:
@kaosat-dev Can this be merged?

I reviewed again, and think this is ready.

kaosat-dev:
@z3dev reviewing now, sorry for the delay !

packages/modeling/src/operations/transforms/mirror.js
```diff
@@ -27,26 +27,41 @@ const mirror = (options, ...objects) => {
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
   const planeOfMirror = plane.fromNormalAndPoint(normal, origin)
```
kaosat-dev: good & clean updates ! nice !
One question about mirror itself that I had not thought about before though : can you not achieve the same thing as giving a specific origin with a translated version of the object you want to mirror ? aka why do we need both normal & origin ?

packages/modeling/src/operations/transforms/rotate.js
```diff
@@ -15,11 +15,14 @@ const { geom2, geom3, path2 } = require('../../geometry')
  */
 const rotate = (angles, ...objects) => {
   if (!Array.isArray(angles)) throw new Error('angles must be an array')
-  if (angles.length !== 3) throw new Error('angles must contain X, Y and Z values')
 
   objects = flatten(objects)
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
+  // adjust the angles if necessary
+  angles = angles.slice() // don't modify the original
```
kaosat-dev: nice !

kaosat-dev: hmm looking at the examples: can we perhaps add a toArray(angles) , so that even single values can be pushed into angle arrays ?
aka rotate(Math.PI/2, geometry) would also work ?
(I am less 100% sure about this api wise)

packages/modeling/src/operations/transforms/scale.js
```diff
@@ -15,12 +15,16 @@ const { geom2, geom3, path2 } = require('../../geometry')
...
 const scale = (factors, ...objects) => {
   if (!Array.isArray(factors)) throw new Error('factors must be an array')
-  if (factors.length !== 3) throw new Error('factors must contain X, Y and Z values')
-  if (factors[0] <= 0 || factors[1] <= 0 || factors[2] <= 0) throw new Error('factors must be positive')
 
   objects = flatten(objects)
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
+  // adjust the factors if necessary
+  factors = factors.slice() // don't modify the original
```
kaosat-dev: same as with rotate about using toArray() 
In this case it would also make it closer to V1 api and allow scale(0.5, objects), which is very useful

packages/modeling/src/operations/transforms/translate.js
```diff
@@ -5,22 +5,25 @@ const mat4 = require('../../math/mat4')
...
-  if (offsets.length !== 3) throw new Error('offsets must contain X, Y and Z values')
+const translate = (offset, ...objects) => {
+  if (!Array.isArray(offset)) throw new Error('offset must be an array')
 
   objects = flatten(objects)
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
-  const matrix = mat4.fromTranslation(offsets)
+  // adjust the offset if necessary
+  offset = offset.slice() // don't modify the original
```
kaosat-dev: same as above

packages/modeling/src/operations/transforms/mirror.js
z3dev: Normal and origin (point) are required to create an arbritary plane, i.e. plane.fromNormalAndPoint(). And, the plane is used to create a matrix, i.e. mat4.mirrorByPlane()

Of course, origin is optional, with the default being 0,0,0.

Also, V1 API only allowed mirror about 0,0,0.

packages/modeling/src/operations/transforms/rotate.js
z3dev: Hmm... the aliases have single values, e.g. rotateX(Math.PI/2, geometry)

So, I'm not so sure about this change. It would be a rare case to rotate objects the same about X,Y,Z.

Looking at all of the transforms, the only one that might make sense is scale() by a single factor.

packages/modeling/src/operations/transforms/mirror.js
kaosat-dev: Ok thanks for clarifying :) 
But my question still stands, can you not achieve the same effect by using a translation and a mirror about 0,0,0 ?

packages/modeling/src/operations/transforms/rotate.js
kaosat-dev: ok, so perhaps it makes sense to go about it the opposite way: why allow 1d or 2d arrays when when need 3d arrays all the time ? ie, perhaps we can save a bit of complexity and have some additional clarity by ONLY allowing 3 component arrays

packages/modeling/src/operations/transforms/mirror.js
z3dev: Sure. Sounds possible.

packages/modeling/src/operations/transforms/rotate.js
z3dev: I think these changes were requested because transforming 2D objects should only require 2 values. That makes sense.

My changes don’t care... any size array is fine. Is that his confusing to users?

kaosat-dev: very good point ! hmm let me think about it until tomorrow please.

kaosat-dev: meh you are right, your changes are good !

packages/modeling/src/operations/transforms/mirror.js
kaosat-dev: in that case why not have the simpler api ? ie just having to provide the normal ?

z3dev: Let me try this... I’ll try some different scenarios of mirror using mirror about 0,0'0, and translate.

z3dev: Well... It turns out that mirror (normal, origin) is not the same as translate(origin) + mirror(normal)

```
const { color, geometry, primitives, transforms } = require('@jscad/modeling')

const main = (params) => {
  let shape1 = primitives.cuboid({size: [10, 15, 5]})
  let shapeA = transforms.center({center: [-20,30,0]}, shape1)
  let shapeB = transforms.center({center: [-20,-30,0]}, shape1)
  let shapeC = transforms.center({center: [20,30,0]}, shape1)
  let shapeD = transforms.center({center: [20,-30,0]}, shape1)
  color.color([1, 0, 0, 1], shapeA, shapeB, shapeC, shapeD)
  let shapes = [shapeA, shapeB, shapeC, shapeD]

  let mirrorsA = transforms.mirror({normal: [15,15,15], origin: [15,15,15]}, shapes)
  geometry.geom3.toPolygons(mirrorsA[0])
  geometry.geom3.toPolygons(mirrorsA[1])
  geometry.geom3.toPolygons(mirrorsA[2])
  geometry.geom3.toPolygons(mirrorsA[3])
  color.color([0, 1, 0, 1], mirrorsA)

  let mirrorsB = transforms.translate([15,15,15], shapes)
  mirrorsB = transforms.mirror({normal: [15,15,15], origin: [0,0,0]}, mirrorsB)
  geometry.geom3.toPolygons(mirrorsB[0])
  geometry.geom3.toPolygons(mirrorsB[1])
  geometry.geom3.toPolygons(mirrorsB[2])
  geometry.geom3.toPolygons(mirrorsB[3])
  color.color([0, 0, 1, 1], mirrorsB)

  return [...shapes, ...mirrorsA, ...mirrorsB]
}

module.exports = { main }
```

P.S. Comment out one of those toPolygons(), and the shape will rotate as the camera moves.

z3dev:
@kaosat-dev please merge.

z3dev:
Time to push on. Thanks, @kaosat-dev for the review.

Issue #528, opened 2020-03-04
V2 : Cannot find module 'most' (npm run dev in 'packages/web')


hrgdavor:
## Expected Behavior
Web running without errors

## Actual Behavior

> @jscad/web@1.10.0 dev /opt/OpenJSCAD.org/packages/web
> budo demo.js --port=8081 --css demo.css --live -- -b

[0000] info  Server running at http://136.243.90.99:8081/ (connect)
[0000] info  LiveReload running
Error: **Cannot find module 'most'** from '/opt/OpenJSCAD.org/packages/web/src/sideEffects/state'
    at /opt/OpenJSCAD.org/packages/web/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:17
    at process (/opt/OpenJSCAD.org/packages/web/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/opt/OpenJSCAD.org/packages/web/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/opt/OpenJSCAD.org/packages/web/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/opt/OpenJSCAD.org/packages/web/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /opt/OpenJSCAD.org/packages/web/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqCallback.oncomplete (fs.js:166:21) {
  stream: [Labeled]
}

## Steps to Reproduce the Problem

  1. follow eyrly adopter document https://openjscad.org/dokuwiki/doku.php?id=early_v2
  2. run the web: `npm run dev` in packages/web
  3.

## Specifications

  - Version: V2
  - Platform: Debian, and Windows
  - Environment: (browser, local server, node.js etc)
     nodejs 12.16.1 LTS

z3dev:
@hrgdavor Super. And thanks for the quick feedback. :)

hrgdavor:
I am using https://openjscad.org/ quite a lot lately, and have moved from OpenSCAD since OpenJSCAD i more live (development wise) ... and I love JS. So looking forward to your help fixing these initial steps, so I can test the V2 as early adopter.

z3dev:
[...]

PR #529, opened 2020-03-06
V2 : update to most versions

z3dev:
These changes standardize on the latest V1 version of most; 1.8.0

In addition, the versions for most accessories are hardened to avoid unwanted changes to behaviour.

This fixes issue #528 

### All Submissions:


z3dev:
@kaosat-dev please reveiw.

I'm not sure the hardening is required but there was an issue with a missing dependency. See #528

kaosat-dev:
@z3dev pinning dependencies is always good !
I am a bit baffled by the missing dependency issue though, no matter the version, there should still be an install of Most

Btw I have a future update that replaces most with something waaay smaller (tiny, tiny), but that is for quite later :)

PR #532, opened 2020-03-15
V2 : Updated to require specific files, eliminating circular dependencies

z3dev:
Background... During the creation of some designs, I ran into a circular dependency on sphere(), which is used internally by expandShell() when expanding 3D geometries.

Solution... In order to eliminate (as much as possible) circular dependencies, all general require statements that use indexes have been replaced with specific require statements.

```
-const {vec2, vec3} = require('../math')
+const vec2 = require('../math/vec2')
+const vec3 = require('../math/vec3')
 
-const {geom3, poly3} = require('../geometry')
+const geom3 = require('../geometry/geom3')
+const poly3 = require('../geometry/poly3')

```

Only the modeling module has been updated.

### All Submissions:

### DIFF
--- a/packages/modeling/src/connectors/transformationBetween.js
+++ b/packages/modeling/src/connectors/transformationBetween.js
@@ -1,4 +1,7 @@
-const { mat4, plane, vec2, vec3 } = require('../math/')
+const mat4 = require('../math/mat4')
+const plane = require('../math/plane')
+const vec2 = require('../math/vec2')
+const vec3 = require('../math/vec3')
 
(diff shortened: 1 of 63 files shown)


kaosat-dev:
this is a very smart fix @z3dev !

z3dev:
> this is a very smart fix @z3dev !

Thanks. It wasn’t possible before because we were still debating names space for functionality.

z3dev:
@kaosat-dev please review and merge if approve.

PR #535, opened 2020-03-17
V2 : Snap 2D geometry using dynamic EPS

z3dev:
These changes adjust the precision of 2D geometry using a dynamicially calculated EPS value.

Due to the 2D to 3D to 2D conversion that happens in 2D intersect/subtract/union, the precision of vertices are slightly different. This causes other parts of the library to fail, such as geom2.toOutlines.

The dynamic EPS values are calculated based on a well know [paper on BSP trees for 2D](http://media.steampowered.com/apps/valve/2014/DirkGregorius_ImplementingQuickHull.pdf).

NOTE: The dynamic EPS calculations do not guarentee perfect integers, and may actually adjust the precision of integer values. The goal is not integers. The goal is pefectly aligned vertices.

### All Submissions:


packages/modeling/src/operations/booleans/measureEpsilon.js
```diff
@@ -0,0 +1,71 @@
...
+ * Measure the epsilon of the given (geom3) geometry.
+ * @return {Float} the epsilon (precision) of the geometry
+ */
+const measureEpsilonOfGeom3 = (geometry) => {
+  let bounds = measureBounds(geometry)
+  let x = bounds[1][0] - bounds[0][0]
+  let y = bounds[1][1] - bounds[0][1]
+  let z = bounds[1][2] - bounds[0][2]
+
+  let epsilon = (x + y + z) / 3 * EPS
```
MatzeS: Refering to Dick Gregorius Quick Hull (http://media.steampowered.com/apps/valve/2014/DirkGregorius_ImplementingQuickHull.pdf) page 30.

The epsilon implementation in the document is different form yours. Did you want to implement it the Dick Gregorius QuickHull way?
As reference implementation we may consider:
https://github.com/mrdoob/three.js/blob/219e3c1bc24163e2c31ab6f5cf4bb66f81b689bc/examples/jsm/math/ConvexHull.js#L495

Unfortunately, I could not find any background information why Dick Gregorius presents this formula or on what geometric/mathematical insight it is based :( 

I am totally fine with your implementation, just pointing out that it is different.

If you would like to change to the reference implementation I could adjust the code for you.

z3dev: The background is choosing edges (or faces) in the same plane. Very closely aligned edges should be considered within the same plane. Look at page 29.

Yeah. It’s a slightly different calculation but basically the same math. I used our constant EPS as the basis for 1 X 1 X 1. We know that this EPS works for designs up to 10 X 10 or more. Any larger design needs a larger EPS. And the same applies for really small designs.

The EPS being calculated in three.js hull is for checking equality of points. It’s slightly different then this use case. And way too small, really.

MatzeS: Oh, I didn't see you were using a custom EPS value.

Unfortunately, I also have to confirm that calculations based on floating point epsilon yield way to small numbers. You are totally right.

z3dev: No problem. I like the feedback. It makes me THINK about what I’m trying to fix. :)

z3dev:
@MatzeS thanks for the approval.

Have you tried some of your examples using this branch?
You should be able to use the CLI.

MatzeS:
;) I already have and it is working

kaosat-dev:
@z3dev I am ok with merging this , but there is a small conflict , do you want to fix it , or shall I as part of the merge ?

z3dev:
@kaosat-dev Oh! Thanks!
I’ll correct the conflict. Hold on...

z3dev:
@kaosat-dev the conflict has been resolved. all tests pass. please merge.

PR #544, opened 2020-03-28
V2 : enhanced polyhedron to accept an array of colors

z3dev:
These changes add support for colors (arrays) to polyhedron. The colors are applied to each face (polygon) of the polyhedron (geom3).

The reason behind these changes is that the several deserializers use polyhedron to produce geometries.

### All Submissions:

### DIFF
--- a/packages/modeling/src/primitives/polyhedron.js
+++ b/packages/modeling/src/primitives/polyhedron.js
@@ -39,7 +39,9 @@ const polyhedron = (options) => {
     if (!Array.isArray(colors)) {
       throw new Error('colors must be an array')
     }
-    if (colors.length !== faces.length) colors = undefined
+    if (colors.length !== faces.length) {
+      throw new Error('faces and colors must have the same length')
+    }
   }
 
   // invert the faces if orientation is inwards, as all internals expect outwarding facing polygons
@@ -47,12 +49,11 @@ const polyhedron = (options) => {
     faces.forEach((face) => face.reverse())
   }
 
-  let polygons = faces.map((face) => poly3.fromPoints(face.map((idx) => points[idx])))
-
-  if (colors) {
-    // add color to each polygon
-    polygons.forEach((polygon, idx) => { if (colors[idx]) polygon.color = colors[idx] })
-  }
+  let polygons = faces.map((face, findex) => {
+    let polygon = poly3.fromPoints(face.map((pindex) => points[pindex]))
+    if (colors && colors[findex]) polygon.color = colors[findex]
+    return polygon
+  })
 
   return geom3.create(polygons)
 }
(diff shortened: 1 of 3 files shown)


kaosat-dev:
I am generally ok with this but unsure about two points:
- the api itself: so you can now pass multiple arrays for different data, it is good because you can omit some of the data, but would an interlaced data structure not also be practical ? (I would say no to be honest, but it has perks): ie : [[posX, posY, posZ, colR, colG, colB], ...]
- would we not want color information per point and not just per face ? (gradients & co ?)

z3dev:
> * would we not want color information per point and not just per face ? (gradients & co ?)

I think you are correct but there would be major changes require throughout the packages. We are already pushing the limits of JavaScript and WeGL, so I’m not sure this will be practical for a few years.

packages/modeling/src/primitives/polyhedron.js
```diff
@@ -32,13 +35,25 @@ const polyhedron = (options) => {
   if (faces.length < 1) {
     throw new Error('one or more faces are required')
   }
+  if (colors) {
+    if (!Array.isArray(colors)) {
+      throw new Error('colors must be an array')
+    }
+    if (colors.length !== faces.length) colors = undefined
```
kaosat-dev, line 42: I think this should throw an error, ie as a user you need to be able to know why the colors are not being displayed :)

packages/modeling/src/primitives/polyhedron.js
```diff
@@ -32,13 +35,25 @@ const polyhedron = (options) => {
...
   // invert the faces if orientation is inwards, as all internals expect outwarding facing polygons
   if (orientation !== 'outward') {
     faces.forEach((face) => face.reverse())
   }
 
   let polygons = faces.map((face) => poly3.fromPoints(face.map((idx) => points[idx])))
+
+  if (colors) {
+    // add color to each polygon
+    polygons.forEach((polygon, idx) => { if (colors[idx]) polygon.color = colors[idx] })
```
kaosat-dev, line 54: hmm this could get very expensive: 
- the faces are mapped over (1 iteration)
- then the polygons are iterated over again to add color
Perhaps adding poly3.fromPointsAndColor makes sense ?
I think it is not essential to do this right now, but at least we should have TODO note in the code to add this later

packages/modeling/src/primitives/polyhedron.js
z3dev: Done

packages/modeling/src/primitives/polyhedron.js
z3dev: Good catch. I merged to two loops.
Done

z3dev:
@kaosat-dev requested changes complete. :)

PR #546, opened 2020-03-31
V2 : Upgrade JSON Deserializer

z3dev:
This pull request updates the JSON deserializer to use V2 modeling and array utils. The goals of this pull request are:
- [X] Update to use V2 apis
- [X] Update code up to JSCAD coding standards
- [X] Add test suites to complete coverage
  * 100% Statements
  * 100% Branches
  * 100% Functions
  * 100% Lines

### All Submissions:


packages/io/json-deserializer/index.js
```diff
@@ -1,110 +1,100 @@
...
-  return ''
-};
+  options = Object.assign({}, defaults, options)
+
+  // convert the JSON notation into anonymous object(s)
+  let objects = JSON.parse(input)
+
+  // cleanup the objects
+  objects = flatten(toArray(objects))
+  objects = objects.filter((object) => geometry.geom3.isA(object) || geometry.geom2.isA(object) || geometry.path2.isA(object))
```
kaosat-dev, line 46: should we not leave non specific jscad object 'as is' ? we have filtering in place in a ton of places, so the data should be able to be kept ?

packages/io/json-deserializer/index.js
```diff
@@ -1,110 +1,100 @@
...
-    var csg = CSG.fromObject(obj)
-    return toSourceCSG(csg)
-  }
-  if (obj.type && obj.type === 'cag') {
-    var cag = CAG.fromObject(obj)
-    return toSourceCAG(cag)
+const version = require('./package.json').version
+
+/**
+ * Deserialize the given JSON notation (string) into either a script or an array of geometry.
```
kaosat-dev, line 23: so all of this is very cool !
BUT , we still need some specific code to deal with TypedArrays
Good news is, the vtree V2 code has that, bad news is , I am not done with it :D

packages/io/json-deserializer/index.js
```diff
@@ -1,110 +1,100 @@
...
+}
+
+const translateToObjects = (objects) => {
+  return objects.reduce((script, object, index) => script + translateToObject(object, index), '')
+}
+
+// translate the given object to JSON notation (AGAIN)
+// NOTE: this implies that the original JSON was correct :)
+const translateToObject = (object, index) => {
+  return `const json${index} = ${JSON.stringify(object)}\n`
```
kaosat-dev, line 92: Hmm is this not the deserializer ? ie JSON => JSCAD  ? so JSON.stringify seems wrong? should be JSON.parse so we get useable JS objects ?

z3dev: The 'object' parameter is that produced by JSON.parse(), see above. Sadly, the translation needs a string so once again JSON.stringify() is called to produce a string from the object.

packages/io/json-deserializer/index.js
z3dev: I added filtering in order to ensure that only JSCAD objects are returned.

I guess that non-JSCAD objects could be returned. Is this the point of your question?

packages/io/json-deserializer/README.md
```diff
@@ -1,13 +1,13 @@
 ## @jscad/json-deserializer
 
-> json deserializer for the jscad project
+> Deserializer of JSON notation to JSCAD geometries
```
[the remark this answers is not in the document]
z3dev: Done

packages/io/json-deserializer/README.md
```diff
@@ -1,13 +1,13 @@
...
 
 [![npm version](https://badge.fury.io/js/%40jscad%2Fjson-deserializer.svg)](https://badge.fury.io/js/%40jscad%2Fjson-deserializer)
 [![Build Status](https://travis-ci.org/jscad/io.svg)](https://travis-ci.org/jscad/json-deserializer)
 
 ## Overview
 
-This deserializer converts raw json data to jscad code (that can be evaluated to CSG/CAG).
+This deserializer converts JSON notation data (files) to JSCAD scripts or geometries.
```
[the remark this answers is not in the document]
z3dev: Done

packages/io/json-deserializer/index.js
kaosat-dev: yes, in particular for JSON, it is important to be able get more data sometimes than just 3D/2D geometry, so we not filter out data unless we really need to imho :)

packages/io/json-deserializer/index.js
kaosat-dev: ah my bad , totally missed that, then yes, it makes sense

packages/io/json-deserializer/index.js
z3dev: @kaosat-dev do we do the same for JSON serializer as well? I’m also filtering there as well.

z3dev: Done

z3dev:
@kaosat-dev do we do the same for JSON serializer as well? I’m also filtering there as well.

kaosat-dev:
@z3dev yes good call , It makes sense to have it work without filtering in both cases !

PR #561, opened 2020-04-19
V2 fix core

z3dev:
This pull request contains a couple of fixes to the core library.

**instantiateDesign()** was properly converting returned solids toArray() but designs can return arrays, or even arrays of arrays. flatten() was added to handle these cases.

**transformSource()** was transforming everything found in the filesAndFolders. all the children (STL, etc. files) were being converted to scripts, causing designs to fail. the transforming of children was removed, as specific file types can be converted directly to geometry via require().

sadly, test cases are not possible for the core library. sigh.

### All Submissions:

### DIFF
--- a/packages/core/code-evaluation/instanciateDesign.js
+++ b/packages/core/code-evaluation/instanciateDesign.js
@@ -4,7 +4,7 @@ const isGeom2 = require('@jscad/modeling').geometry.geom2.isA
 const isGeom3 = require('@jscad/modeling').geometry.geom3.isA
 const isPath2 = require('@jscad/modeling').geometry.path2.isA
 
-const { toArray } = require('@jscad/array-utils')
+const { flatten, toArray } = require('@jscad/array-utils')
 
 // const toCompactBinary = require('./toCompactTest')
 const isResultSolid = (rawResults) => (rawResults.length > 0 && (isGeom3(rawResults[0]) || isGeom2(rawResults[0]) || isPath2(rawResults[0]) ))
@@ -60,7 +60,7 @@ const instanciateDesign = (rootModule, parameterValues, options) => {
   const { vtreeMode, inputLookup, inputLookupCounts, serialize } = options
   // deal with the actual solids generation
   let solids
-  let rawResults = toArray(rootModule.main(parameterValues))
+  let rawResults = flatten(toArray(rootModule.main(parameterValues)))
 
   const forcedNOVtreeMode = false // FIXME: disabling Vtree mode for now until more V2 progress is done
   if (forcedNOVtreeMode) {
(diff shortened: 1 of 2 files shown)


kaosat-dev:
@z3dev I agree about the addition of `flatten` but unless I am mistaken, removing the transform of children has a lot of side effects:
- moduyfing the old type scripts would not work with folders etc (so unuseable), (we may want to remove even allowing to create scripts without explicit imports of the API, so perhaps not so bad)
- stl/etc files in folders would not work anymore (if I recall the way children work correctly)
- Could you please elaborate a bit on 'causing designs to fail.' ? (I am genuinely curious, I cannot see a concrete case, as I have lots of designs with stls in the same folder, and did not used to have issues on V2 branch)
- last but not least, this also directly ties in with to script vs csg object for drag & drop & co, and I still do not feel smarter about that

Oh and tests are possible :) I will send you something in that area tomorrow

z3dev:
>     * moduyfing the old type scripts would not work with folders etc (so unuseable), (we may want to remove even allowing to create scripts without explicit imports of the API, so perhaps not so bad)

Good point.

We definitiely need to discuss how to support V1 projects, if possible.

First, I thought support for the old 'include' was removed, so designs need to be changed to use 'require', and 'module.exports'. Is this the case?

z3dev:
>     * Could you please elaborate a bit on 'causing designs to fail.' ? (I am genuinely curious, I cannot see a concrete case, as I have lots of designs with stls in the same folder, and did not used to have issues on V2 branch)

Here the failing project.
- project3 (folder)
  - main.js
  - test.stl
  - test.dxf

kaosat-dev:
> Is any kind of support for V1 designs possible?
Good question, and it is a bit tricker than that even : there are V1 designs with include() (we dropped support for that, giant mess), there is also V1 (a more recent one) with require() support...
My first instinct is to say 'drop support, perhaps add tutorials/ converter scripts down the line' ?

z3dev:
> Good question, and it is a bit tricker than that even : there are V1 designs with include() (we dropped support for that, giant mess), there is also V1 (a more recent one) with require() support...
> My first instinct is to say 'drop support, perhaps add tutorials/ converter scripts down the line' ?

Ah... Designs created via the desktop application.
How about adding a section to V2 Changes document? We could write some tips (with example?) on converting to V2.

kaosat-dev:
I agree, and would love to work on that (I still have the examples overhaul in the pipeline).
Btw This 'intermediary' version thing makes me also think that V1.5 of sorts would also need to be on the new website (I am almost done with a draft version of multi version website, essentially jscad.xyz/V1/docs , jscad.xyz/V1/playground etc, using the same url schema)

z3dev:
@kaosat-dev please merge these changes as further testing is not possible without these.

Also, I added 'how to convert V1 to V2' to the V2 checklist, so we don't forget.

PR #566, opened 2020-04-23
V2 : fix color() to return new geometries

### LINKED ISSUE #564: v2: Unexpected color api
Most jscad functions that take an object return a new object with the changes applied. I was surprised to discover that the color function modifies the original object. What's the reasoning behind this? Shouldn't the color function mach the rest of the api?

z3dev:
This pull request changes the behaviour of the color() function, returning NEW geometries.

Due to the behaviour change, some of the IO deserializers needed changes to apply colors to objects (instantiate) and translate colors into proper scripts.

### All Submissions:


packages/modeling/src/color/color.js
```diff
@@ -1,10 +1,37 @@
 const flatten = require('../utils/flatten')
 
+const geom2 = require('../geometry/geom2')
+const geom3 = require('../geometry/geom3')
+const path2 = require('../geometry/path2')
+const poly3 = require('../geometry/poly3')
+
+const colorGeom2 = (color, object) => {
+  const newgeom2 = geom2.create(geom2.toSides(object))
```
kaosat-dev, line 9: would it not make more sense to just use the clone() function ?

packages/modeling/src/color/color.js
```diff
@@ -1,10 +1,37 @@
...
+  newgeom2.color = color
+  return newgeom2
+}
+
+const colorGeom3 = (color, object) => {
```
kaosat-dev, line 14: Imho, I think each of these belongs inside their specific geometries folder (ie geom2/color, geom3/color) but that is a change that can be done later, that I would love to tackle :)

packages/modeling/src/color/color.js
```diff
@@ -1,10 +1,37 @@
...
+const colorPoly3 = (color, object) => {
+  object.color = color
+}
+
 /**
  * Apply the given color to the given objects.
  * @param {Array} color - RGBA color values, where each value is between 0 and 1.0
  * @param {Object|Array} objects - the objects of which to color
- * @returns {Object|Array} the same objects with an additional attribute 'color'
+ * @returns {Object|Array} new geometry with an additional attribute 'color'
```
kaosat-dev, line 34: yay ! :)

packages/modeling/src/color/color.js
```diff
@@ -1,10 +1,37 @@
...
+}
+
+const colorPath2 = (color, object) => {
+  const newpath2 = path2.create(path2.toPoints(object))
+  newpath2.color = color
+  return newpath2
+}
+
+const colorPoly3 = (color, object) => {
+  object.color = color
```
kaosat-dev, line 27: this is mutating the original still , so the same logic as above would work on the otherhand :)

packages/modeling/src/color/color.js
z3dev: Alas, clone is a DEEP clone, which makes full copies of all polygons, and vertices. Very expensive.

z3dev: Actually, thinking a little more about this. The clone functions should be SHALLOW copies, as geometry is 'immutible' by design. But there may be some side effects of changing the clone functions.

@kaosat-dev do you think this is worth correcting now?

RedHatter:
@z3dev Did you mean to commit the torus fix to this branch?

packages/modeling/src/color/color.js
kaosat-dev: @z3dev sorry for the slow reply. This has been nagging me for a few days, I am a bit baffled, but I agree this needs a closer attention to the clone() functions : what do you mean by 'deep clone' ? we just want a fresh copy of the data, and that already seems to be the case (I looked at Geom2.clone(), Geom3.clone() etc)

packages/modeling/src/color/color.js
kaosat-dev: hmm but you want a new geometry containing the **new** polygon with a different color (immutable) right ? 
Ie, if you want to color an individual polygon (like in this case), it should return a new polygon imho

packages/modeling/src/color/color.js
z3dev: DEEP CLONE.... all data is replicated... new arrays

SHALLOW CLONE... the polygons and the transforms are passed on...

packages/modeling/src/color/color.js
z3dev: Ah... colors are just an attribute at the geometry level. No more. We decided to support this long back.

`let newgeom = color([1,1,1,1], circle()) // geom2.color =[…]`

So, adding color to a polygon is a special case.

z3dev: @kaosat-dev I'm fine changing colorPoly3 to return a new polygon. Do you want this changed?

kaosat-dev: Yes please, I think it makes sense to keep it all coherent (I agree that performance wise it is not the best, but it is an edge case, and immutability has pros & cons)

z3dev: Done

packages/modeling/src/color/color.js
z3dev: I create issue #573 to debate this further.

z3dev:
@kaosat-dev please review again. i believe this is ready to merge.

z3dev:
@kaosat-dev please do not merge, as the BIG PR for lint changes should probably go first.

kaosat-dev:
@z3dev are you sure about that ? This seemed like a much smaller PR , and the lint one, still needs to merge V2 back into it ?

z3dev:
> @z3dev are you sure about that ? This seemed like a much smaller PR , and the lint one, still needs to merge V2 back into it ?

It's probably the same either way, but I'd prefer to get those lint changes done and dusted. I have a few more small set of changes in the pipeline, which would benefit from those lint changes.

z3dev:
@kaosat-dev this is ready for merging now.

PR #569, opened 2020-04-25
V2: Missing variable declaration

RedHatter:
Missing variable declaration causing failure in strict mode.

### DIFF
--- a/packages/modeling/src/math/mat4/clone.js
+++ b/packages/modeling/src/math/mat4/clone.js
@@ -8,7 +8,7 @@ const create = require('./create')
  * @returns {mat4} clone of the given matrix
  */
 const clone = (...params) => {
-  let out
+  let out, a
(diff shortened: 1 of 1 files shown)


z3dev:
@RedHatter how did you catch that? Did you run node with strict mode?

RedHatter:
I'm running it in a program built with rollup. In particular I found this error by running in chrome with 'use strict' at the top of the file.

RedHatter:
Sorry, wrong button

z3dev:
Hanks. There are probably other oversights in the code. It’s probably better to fix all the lint errors. But that will be a slightly larger effort.

kaosat-dev:
Thanks for the fix @RedHatter !
@z3dev this is small & fast to merge, shall we ?

z3dev:
@kaosat-dev sure. Looks fine.

PR #583, opened 2020-05-17
V2 : fix rebuild geometry for cli

### LINKED ISSUE #504: V2: cli not finding require'd modules
I can't seem to use the same code in the cli and web when requiring a local module. 
When I require a module in a sub directory and render with a local web instance, it works.
I expect webRequire.js is saving me from doing 'bad things' with untrusted code over the web, but cli.js should just trust my includes, shouldn't it.
Using a full path in the require() returns the same error but with the full path in the error message with '/home/...' and not 'home/...' when using cli.js.

## Expected Behavior
`const { extrudeText } = require('./fonts/index.js')`
makes exported function in file ./fonts/index.js available when using localhost web app and cli
## Actual Behavior
gets error whe using cli:
`Error: No file home/chris/Projects/3DPrinting/MiscDesigns/fonts/index.js found`
`    at _require (/home/chris/lib/JSCAD-dev/OpenJSCAD.org/packages/core/code-loading`/webRequire.js:149:13)
notice file path is expanded to be relative as  'home/..'

  1. create module in subdir in design project
  2. drag and drop to web app - it works
  3. run it with cli.js - gets the above error

## Specifications

  - Version: commit 5abe983 (HEAD -> V2, origin/V2)
  - Platform: Linux (ubuntu 19.10)
  - Environment:  (firefox 72.0.1, localhost server, nodejs v8.16.1)

z3dev:
These changes adjust the logic in rebuildGeometryCli() to ONLY use webRequire() when a conversion to JSCAD source happens, e.g. STL translation. The opposite is also true, the logic ONLY uses require() in normal cases.

Fixes #504 for the CLI only.
Fixes conversions from STL to DXF, etc. (test case has been enabled)

I always wondered why webRequire() was used by rebuildGeometryCli(), and now I understand. There's a special case when external formats are converted in-memory, which webRequire() handles nicely.

### All Submissions:

### DIFF
--- a/packages/core/code-evaluation/rebuildGeometryCli.js
+++ b/packages/core/code-evaluation/rebuildGeometryCli.js
@@ -3,31 +3,31 @@ const path = require('path')
 const { toArray } = require('@jscad/array-utils')
 const requireDesignFromModule = require('../code-loading/requireDesignFromModule')
 const getAllParameterDefintionsAndValues = require('../parameters/getParameterDefinitionsAndValues')
-const transformSources = require('../code-loading/transformSources')
 const makeWebRequire = require('../code-loading/webRequire')
-const makeFakeFs = require('../code-loading/makeFakeFs')
-const { registerAllExtensions } = require('../io/registerExtensions')
 
 const rebuildSolids = (data) => {
   const defaults = { vtreeMode: false, serialize: false }
-  let { mainPath, vtreeMode, parameterValues, inputIsDirectory } = Object.assign({}, defaults, data)
+  let { mainPath, vtreeMode, parameterValues, useFakeFs } = Object.assign({}, defaults, data)
   const apiMainPath = vtreeMode ? '../code-loading/vtreeApi' : '@jscad/modeling'
   // we need to update the source for our module
   let requireFn = require
 
   // source came from conversion, i.e. not from file system
-  if (data.source) {
-    let filesAndFolders = [
+  if (useFakeFs) {
+    const pathParts = path.parse(mainPath)
+    const fakeName = `${pathParts.name}.js`
+    const fakePath = `/${pathParts.name}.js`
+    const filesAndFolders = [
       {
         ext: 'js',
-        fullPath: './index.js',
-        name: 'index.js',
+        fullPath: fakePath,
+        name: fakeName,
         source: data.source
       }
     ]
     requireFn = makeWebRequire(filesAndFolders, { apiMainPath })
 
-    mainPath = './index.js' // and use the alias as the entry point
+    mainPath = fakePath // and use the alias as the entry point
   }
 
   // rootModule should contain exported main and getParameterDefinitions functions
--- a/packages/core/code-evaluation/rebuildGeometryCli.js
+++ b/packages/core/code-evaluation/rebuildGeometryCli.js
@@ -10,34 +10,27 @@ const { registerAllExtensions } = require('../io/registerExtensions')
 
 const rebuildSolids = (data) => {
   const defaults = { vtreeMode: false, serialize: false }
-  const { mainPath, vtreeMode, parameterValues, inputIsDirectory } = Object.assign({}, defaults, data)
+  let { mainPath, vtreeMode, parameterValues, inputIsDirectory } = Object.assign({}, defaults, data)
   const apiMainPath = vtreeMode ? '../code-loading/vtreeApi' : '@jscad/modeling'
   // we need to update the source for our module
   let requireFn = require
-  if (!inputIsDirectory) {
+
+  // source came from conversion, i.e. not from file system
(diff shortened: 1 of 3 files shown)


z3dev:
@kaosat-dev please review.

by the way, if we want to enforce 'projects' then we should block single file scripts, or warn if single file scripts are used. what do you think?

packages/core/code-evaluation/rebuildGeometryCli.js
```diff
@@ -10,34 +10,27 @@ const { registerAllExtensions } = require('../io/registerExtensions')
...
-    const fakeFs = makeFakeFs(filesAndFolders)
-    requireFn = makeWebRequire(filesAndFolders, { apiMainPath, fakeFs })// hasRequire() ? require : makeWebRequire(filesAndFolders, { apiMainPath })
-    // register all extension formats
-    registerAllExtensions(fakeFs, requireFn)
+    mainPath = './index.js' // and use the alias as the entry point
   }
 
-  // console.log('transformed sources', filesAndFolders)
-  // now check if we need fake require or not
-  // FIXME: we need to come up with a way to intercept node 'require' calls to be able to apply transformSources on the fly
```
kaosat-dev, line 37: Unless this is fixed, please leave as is :)
Removing console.log() debug statements is good, but TODO/FIXMEs are there for a reason

packages/core/code-evaluation/rebuildGeometryCli.js
```diff
@@ -10,34 +10,27 @@ const { registerAllExtensions } = require('../io/registerExtensions')
...
   const defaults = { vtreeMode: false, serialize: false }
-  const { mainPath, vtreeMode, parameterValues, inputIsDirectory } = Object.assign({}, defaults, data)
+  let { mainPath, vtreeMode, parameterValues, inputIsDirectory } = Object.assign({}, defaults, data)
   const apiMainPath = vtreeMode ? '../code-loading/vtreeApi' : '@jscad/modeling'
   // we need to update the source for our module
   let requireFn = require
-  if (!inputIsDirectory) {
+
+  // source came from conversion, i.e. not from file system
+  if (data.source) {
```
kaosat-dev, line 19: A few questions for clarity please:
- can we not reuse the inputIsDirectory in this case ? is there any difference ?
- why is mainPath overwritten ? ie does this get triggered *only* if there is a single input script ?

packages/core/code-evaluation/rebuildGeometryCli.js
```diff
@@ -10,34 +10,27 @@ const { registerAllExtensions } = require('../io/registerExtensions')
...
       }
     ]
-    filesAndFolders = transformSources({ apiMainPath }, filesAndFolders)
-    // console.log('filesAndFolders', filesAndFolders)
+    requireFn = makeWebRequire(filesAndFolders, { apiMainPath })
 
-    const fakeFs = makeFakeFs(filesAndFolders)
-    requireFn = makeWebRequire(filesAndFolders, { apiMainPath, fakeFs })// hasRequire() ? require : makeWebRequire(filesAndFolders, { apiMainPath })
-    // register all extension formats
-    registerAllExtensions(fakeFs, requireFn)
```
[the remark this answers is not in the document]
z3dev: CLI generateOutputData() make the same call using NODEJS require(), so i removed this unnecessary call.

packages/core/code-evaluation/rebuildGeometryCli.js
z3dev: I added the source versus prevsource to determine if the conversion took place. And adjusted rebuildGeometryCli() to use webRequire() if a source (converted source) is present.

The use of inputIsDirectory for single files was causing issues when *.stl, etc. files were passed in. For example, ./cli.js my.stl -of dxf

The use of data.source is not ideal. It would better be better to use a real option, i.e. isFromConversion.

For the webrequire to work properly, both fullPath and mainPath need to be the same. We could derive a 'in-memory' name from the original mainPath, but I don't see the reason.

packages/core/code-evaluation/rebuildGeometryCli.js
z3dev: > FIXME: we need to come up with a way to intercept node 'require' calls to be able to apply transformSources on the fly

I think is is fixed. The call to registerAllExtensions() makes this possible. And a very cool solution as well.

packages/core/code-evaluation/rebuildGeometryCli.js
z3dev: @kaosat-dev please review again.

i changed the logic slightly. there's a new option to rebuildGeometryCli called 'useFakeFs'. This option requestes rebuildGeometryCli to create a fake file system, and of course use webRequire as well.

PR #613, opened 2020-07-02
V2 : remove planes in poly3

z3dev:
This set of changes remove planes from poly3. poly3 now only consists of vertices. This has several affects:
- lower memory requirements, and less allocations of memory
- faster performance for functions that do not require planes; primitives, extrudes, etc.

Performace testing has also shown that the removal of planes does not adversely affect other functions; booleans, extrusions, etc.

Test cases have been updated to remove comparisons of planes as well.

Testing with the WEB UI has also been performed, and a small change to regl-renderer was required.

### All Submissions:


packages/modeling/src/operations/booleans/trees/Node.js
```diff
@@ -105,7 +106,8 @@ Node.prototype = {
         index = Math.floor(polygontreenodes.length / 2)
         // index = polygontreenodes.length >> 1
         // index = Math.floor(Math.random()*polygontreenodes.length)
-        const bestplane = polygontreenodes[index].getPolygon().plane
+        const bestpoly = polygontreenodes[index].getPolygon()
+        const bestplane = poly3.plane(bestpoly)
         node.plane = bestplane
```
z3dev, line 111: Just assign it node.plane

PR #625, opened 2020-07-16
Adding maths/bezier object for curve-based easing and extruding.

SimonClark:
Adding a bezier object for curve-based easing.  Created a V2 directory in the examples module, and there are 2 examples of using it in there.

It behaves a little differently than other objects, in that it has functions you call to get the result you need, rather than being acted upon by methods in the modules.  I felt this was reasonable, as it is a helper object, but I am happy to discuss.


z3dev:
@SimonClark thanks! This looks really interesting, so let’s discuss some points. I’ll separate out the discussions as it it make dialogues easier.

z3dev:
Where does this belong?

In general, the maths data types are using arrays for data storage, and have a set of functions that manipulate the data storage, or calculate values. These are used to calculate and create geometry.

In general, the geometries are anonymous objects, and have a set of functions that produce a new geometry after applying some ‘transformation’. These are input parameters to operations, which also produce new geometry. And used by IO modules for conversions.

The new bezier functionality kind of looks like both. But maybe fits better into geometry because...
- it’s an object
- operations could accept a bezier object, e.g. offset({delta: 5}, mybezier)
- IO modules could convert to/from bezier objects
- etc

z3dev:
> It behaves a little differently than other objects, in that it has functions you call to get the result you need, rather than being acted upon by methods in the modules. I felt this was reasonable, as it is a helper object, but I am happy to discuss.

I wouldn’t consider the bezier functionality as a helper object. It would be far more useful as a fully integrated piece of JSCAD, with a fully exposed and useful API for users. :)

Long back, we decided to drop the CLASS API of V1. By flattening the API, we have simple anonymous objects with a set of simple common functions. And, users can then assemble higher level objects/classes easily. Simple is best.

The bezier object is just fine but can getValueAt() and getTangentAt() be brought out as functions work upon bezier objects? I don’t think it will be that different, and the code may be easier to understand.

z3dev:
API questions:

- can the control points be expanded?

let newbezier = bezier.addControlPoints([…], oldbezier)

- can bezier produce a set of 2D/3D points?

let allpoints = bezier.toPoints({options}, mybezier)

SimonClark:
> The new bezier functionality kind of looks like both. But maybe fits better into geometry because...
> 
> * it’s an object
> * operations could accept a bezier object, e.g. offset({delta: 5}, mybezier)
> * IO modules could convert to/from bezier objects
> * etc

It's a weird one, because there is not a 1:1 mapping of object to concept.  The easing function could be used to represent a curve, fed into various transforms as scaling or rotations, etc. I see it as powerful, but weird, in that nothing else in jscad has quite the same flexibility of interpretation.

Interesting point about the offset.  I guess you could arguably transform it like other geometries, and I could see value in that.  Lends credence to making it a geometry.

SimonClark:
> > It behaves a little differently than other objects, in that it has functions you call to get the result you need, rather than being acted upon by methods in the modules. I felt this was reasonable, as it is a helper object, but I am happy to discuss.
> 
> I wouldn’t consider the bezier functionality as a helper object. It would be far more useful as a fully integrated piece of JSCAD, with a fully exposed and useful API for users. :)
> 
> Long back, we decided to drop the CLASS API of V1. It was just too much maintenance to keep both CLASS and FUNCTIONAL APIs. By flattening the API, we now have simple anonymous objects with a set of simple common functions. And, users can then assemble higher level objects/classes easily. And we have operations that can accept any of the geometries. Simple is best.
> 
> The bezier object is just fine but can getValueAt() and getTangentAt() be brought out as functions that work upon bezier objects? I don’t think it will be that different, and the code may be easier to understand.

Yeah, alright.  You've convinced me :). I'm concerned that we could apply transforms to objects and beziers, but only getValueAt(), etc to beziers.  I can get over that, but it messes with the part of my brain that appreciates OOP.

SimonClark:
> * can the control points be expanded? 
> let newbezier = bezier.addControlPoints([…], oldbezier)

Yes, with probably very little extra code. I'm not sure I can see a compelling use case.  Is this something you'd like to see to meet some sort of internal consistency?

> * can bezier produce a set of 2D/3D points? (This is for IO conversions, etc.)
> let allpoints = bezier.toPoints({options}, mybezier)

Yeah, of course. There's no guarantee of the dimensionality or format of the points, other than they will match what you put in.

z3dev:
> Yeah, alright. You've convinced me :). I'm concerned that we could apply transforms to objects and beziers, but only getValueAt(), etc to beziers. I can get over that, but it messes with the part of my brain that appreciates OOP.

Ha ha ha! I understand totally. I’m from a deep OOP background as well. The underlying theme is... stay lean and low level... let users build the OOPS

z3dev:
> > The new bezier functionality kind of looks like both. But maybe fits better into geometry because...
> > 
> > * it’s an object
> > * operations could accept a bezier object, e.g. offset({delta: 5}, mybezier)
> > * IO modules could convert to/from bezier objects
> > * etc
> 
> It's a weird one, because there is not a 1:1 mapping of object to concept. The easing function could be used to represent a curve, fed into various transforms as scaling or rotations, etc. I see it as powerful, but weird, in that nothing else in jscad has quite the same flexibility of interpretation.
> 
> Interesting point about the offset. I guess you could arguably transform it like other geometries, and I could see value in that. Lends credence to making it a geometry.

Yeah. I’m still kind of wondering. The bezier implementation is very close to our V3 concept, where the primitives are ‘descriptive’. For example, circle() just returns a ‘descriptive’ object that contains origin and radius. Then, the descriptive is converted to geometry when necessary, or rendered when necessary, or converted to SVG when necessary, or...

This might be something new! :)

packages/modeling/src/maths/bezier/create.js
```diff
@@ -0,0 +1,154 @@
...
+		out.getValueAt = function(t) {
+			return this.bezierFunction(this.points, t);
+		};
+		out.getTangentAt = function(t) {
+			return this.bezierTangent(this.points, t);
+		};
+	} else {
+		out.dimensions = points[0].length;
+		out.getValueAt = function(t) {
+			var result = new Float32Array(this.dimensions);
```
z3dev, line 129: Not sure why Float32Array is being used. Experience has taught us that those extra 32 bits of precision are super important.

SimonClark: Take them to Float64Array? I ran into the Float32Array elsewhere in the codebase, took it in as my own.

z3dev: See PR #624, which replaces Float23Array with Number. Huge performance and precision differences, which is required in this library. :)

jsdoc.json
```diff
@@ -3,13 +3,14 @@
     "include": [
       "packages/modeling/src/colors",
       "packages/modeling/src/connectors",
-      "packages/modeling/src/geometry",
-      "packages/modeling/src/geometry/geom2",
-      "packages/modeling/src/geometry/geom3",
-      "packages/modeling/src/geometry/path2",
-      "packages/modeling/src/geometry/poly2",
-      "packages/modeling/src/geometry/poly3",
+      "packages/modeling/src/geometries",
```
z3dev, line 6: Please revert as these changes (fixes for documentation) should be part of another PR.

z3dev:
> > The new bezier functionality kind of looks like both. But maybe fits better into geometry because...
> > 
> > * it’s an object
> > * operations could accept a bezier object, e.g. offset({delta: 5}, mybezier)
> > * IO modules could convert to/from bezier objects
> > * etc
> 
> It's a weird one, because there is not a 1:1 mapping of object to concept. The easing function could be used to represent a curve, fed into various transforms as scaling or rotations, etc. I see it as powerful, but weird, in that nothing else in jscad has quite the same flexibility of interpretation.
> 
> Interesting point about the offset. I guess you could arguably transform it like other geometries, and I could see value in that. Lends credence to making it a geometry.

@SimonClark I hope your still around. :)

I've been researching a few concepts, including NURBS. And, it seems that this is something new, but also another necessary piece of to form complex solid geometries. The JSCAD 'geometries' are basically the solids, which are produced today by 'primitives' and 'operations'.

Extending this further, this new piece called 'bezier' can also be used to create geometry. Basically, we don't have a concept of 3D lines/arc/curves today. Your use case is basically using the curves for extrusions (which I have also done using 2D beziers as part of path2). However, the 'bezier' functionality can also create 3D 'paths'? (I'm not sure what the new geometry would be called yet.)

Here something that I ran across from Autodesk Fusion, which explains their concepts of 'model' and 'geometry', which produce 'solids'. https://help.autodesk.com/view/fusion360/ENU/?guid=GUID-1C3FFADB-52C4-49BB-8981-4A448FFE4442

I'm also attaching the Fusion object model. See the section at the bottom left call 'Geometry'.

z3dev:
@SimonClark just wondering... this implementation looks very similar to NURBS curves. What do you think?

https://en.m.wikipedia.org/wiki/Non-uniform_rational_B-spline

SimonClark:
> @SimonClark I hope your still around. :)
Still around, just busy with my day job and family life.  Don't worry, I don't give up that easily :)

> Here something that I ran across from Autodesk Fusion, which explains their concepts of 'model' and 'geometry', which produce 'solids'. https://help.autodesk.com/view/fusion360/ENU/?guid=GUID-1C3FFADB-52C4-49BB-8981-4A448FFE4442

I'll take a look at these.  Admittedly the bezier object as I implemented it was to solve a specific requirement I had.  NURBS are great, too, especially when the control points are required to intersect the curve.  That's something the bezier curves don't do well.  

I do see a ton of use for a NURBS surface mesh.  I need to understand the mathematics of surface to solid modelling a bit better before I consider tackling that.

What would be your preference. Submit the bezier code as is, and explore that further later, or get it fully functional with bezier and NURBS options first?

z3dev:
> 
> What would be your preference. Submit the bezier code as is, and explore that further later, or get it fully functional with bezier and NURBS options first?

Let’s just get the bezier integrated to a point where others can use find and use it. It’s the fist of its kind.

I’ve been thinking that there’s a new category of functions called ‘curves’ or something similar. Then all the various curves can be organized together and maintain similar APIs (if possible).

z3dev:
@SimonClark how about putting this in 'curves'? i think this is the first of a new class of objects.

FYI, you need to updated to the latest V2 as well as some name spaces have been moved; math => maths, geometry => geometries.

SimonClark:
> @SimonClark how about putting this in 'curves'? i think this is the first of a new class of objects.
Sounds reasonable.  Hoping to get to these changes in the next day or so.

jsdoc.json
z3dev: The latest change looks good. did you try to generate documentation?

`npm run docs`

packages/modeling/src/primitives/index.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Primitives provide the building blocks for complex parts.
+ * Primitives provide the building blocks for complex parts.xx
```
z3dev: Is this change expected? Awaiting some additional comments?

z3dev:
@SimonClark this is looking better.

FYI, we are using a coding style called [Standard](https://standardjs.com/), which is pretty popular now. You may not like some of the styles but all code needs to comply. You can run lint on the code using...

`npm run lint` 

You can temporarily tweak package.json to only lint the curves code.

SimonClark:
> @SimonClark this is looking better.
> 
> FYI, we are using a coding style called [Standard](https://standardjs.com/), which is pretty popular now. You may not like some of the styles but all code needs to comply. You can run lint on the code using...

👍

packages/modeling/src/curves/bezier/create.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+const getPointType = function (points) {
+  let firstPointType = null
+  points.forEach(point => {
+    let pType = ''
+    if (typeof (point) === 'number') {
+      pType = 'float_single'
+    } else {
+      point = new Float32Array(point)
```
z3dev, line 34: Is this 'conversion' required? I think the point (and values and types) are being checked below.

packages/modeling/src/curves/bezier/create.js
```diff
@@ -0,0 +1,73 @@
...
+      if (point) {
+        point.forEach(val => {
+          if (isNaN(val)) {
```
z3dev, line 37: If I understand correctly then only Number values should be allowed. If so then the global isNaN() is probably not the correct choice. You should probably use Number.isFinite()

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite

packages/modeling/src/curves/bezier/create.js
```diff
@@ -0,0 +1,73 @@
...
+ * @example
+ * const b = bezier.create([0,10]) // a linear progression from 0 to 10
+ * const b = bezier.create([0, 0, 10, 10]) // a symmetrical cubic easing curve that starts slowly and ends slowly from 0 to 10
+ * const b = bezier.create([0,0,0], [0,5,10], [10,0,-5], [10,10,10]]) // a cubic 3 dimensional easing curve that can generate position arrays for modelling
+ *
+ * @param {Array} points An array with at least 2 elements of either all numbers, or all arrays of numbers that are the same size.
+ * @returns {bezier} a new bezier data object
+ * @alias module:modeling/curves/bezier.create
+ */
+const create = (points) => {
```
z3dev, line 14: How about using rest parameters here? Then it's possible to create using just a series of numbers or a series of arrays.
```
const b = bezier.create( 0,10 ) // a linear progression
const b = bezier.create( 0, 0, 10, 10 ) // a symmetrical cubic easing curve
const b = bezier.create( [0,0,0], [0,5,10], [10,0,-5], [10,10,10] ) // a cubic 3 dimensional easing curve
```
Here's a good explanation of the rest parameters. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

SimonClark: Sure.  I wasn't sure if that was a JSCAD thing for creation objects.  Most analogous would be slice/fromPoints, which does not offer this.  if this is a direction you want jscad to go in, I'm happy to do it though.

packages/modeling/src/curves/bezier/create.js
SimonClark: Hah, cool.  I looked at isNaN and isFinite, but both would say that isFinite([5]) is a valid number.  However isFinite() behaves a bit differently than Number.isFinite().  Thanks for introducing me to that.

packages/modeling/src/curves/bezier/create.js
z3dev: JSCAD API for transforms, booleans, etc. are already using rest parameters, so no issues.

Try it out. I think it makes sense.

SimonClark: Sure, but those are saying "Apply this to these independent entities".  

bezier.create(), slice.fromPoints(), geom2.fromPoints(), etc are all "create something from this set of tightly related data".  I would argue that it makes sense not to use rest parameters for those, especially if the data set that you are passing in may be generated algorithmically.

packages/modeling/src/curves/bezier/create.js
z3dev: Welcome. I came across those differences in another library.

packages/modeling/src/curves/bezier/create.js
z3dev: Sounds fine then. Let’s stay with the current parameters.

packages/modeling/src/curves/bezier/tangentAt.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Calculates the position at a specific point along a bezier easing curve.
```
z3dev, line 2: Umm... it seems that the documention has been reversed in tangetAt and valueAt.

SimonClark: hah!  thanks

packages/modeling/src/curves/bezier/index.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @example
+ * const b = bezier.create([0,10]) // a linear progression from 0 to 10
+ * const b = bezier.create([0, 0, 10, 10]) // a symmetrical cubic easing curve that starts slowly and ends slowly from 0 to 10
+ * const b = bezier.create([0,0,0], [0,5,10], [10,0,-5], [10,10,10]]) // a cubic 3 dimensional easing curve that can generate position arrays for modelling
+ * Usage:
+ * let position = bezier.valueAt(t,b) // where 0 < t < 1
+ * let tangent = bezier.tangentAt(t,b) // where 0 < t < 1
+ *
+ * @typedef {Object} bezier
```
z3dev, line 16: Can you add some information about the properties (attributes of the object)? For example...

` * @property {Array} sides - list of sides, each side containing two points`

packages/modeling/src/curves/index.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Curves are n-dimensional mathematical constructs that define a path from point 0 to point 1
```
z3dev, line 2: Nice!

packages/modeling/src/curves/bezier/index.js
SimonClark: 👍

z3dev:
By the way, if you want to discuss something then feel free to contact me at z3-dev@gfnews.net

PR #630, opened 2020-07-27
V2 : tune retesselate

z3dev:
These changes tune the comparison of plane normals, which assists retesselate.

There are also changes to the measurements, caching calculated values as object attributes. For example, object.volume, etc. These are expected to improve the performace of boolean operations.

Additional testing was completed with both CLI and WEB applications.

### All Submissions:

### DIFF
--- a/packages/modeling/src/measurements/measureArea.js
+++ b/packages/modeling/src/measurements/measureArea.js
@@ -23,10 +23,12 @@ const measureAreaOfPath2 = () => 0
  * @returns {Number} area of the geometry
  */
 const measureAreaOfGeom2 = (geometry) => {
+  if (geometry.area) return geometry.area
+
   const sides = geom2.toSides(geometry)
-  let area = sides.reduce((area, side) => area + (side[0][0] * side[1][1] - side[0][1] * side[1][0]), 0)
-  area *= 0.5
-  return area
+  const area = sides.reduce((area, side) => area + (side[0][0] * side[1][1] - side[0][1] * side[1][0]), 0)
+  geometry.area = area * 0.5
+  return geometry.area
 }
 
 /*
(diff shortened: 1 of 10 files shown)


packages/modeling/src/measurements/measureArea.js
```diff
@@ -23,10 +23,12 @@ const measureAreaOfPath2 = () => 0
  * @returns {Number} area of the geometry
  */
 const measureAreaOfGeom2 = (geometry) => {
+  if (geometry.area) return geometry.area
```
SimonClark, line 26: Is this caching dangerous at all?  If the geometry is altered at all, does area either get invalidated or recalculated?

z3dev: A fair question. All operations on geometry produce a new object, as each object is considered immutable. For example, transform() creates a new object with a new matrix and a copy of the polygons. Everything else is discarded, i.e. the measurements.

SimonClark: Good point. 👍

packages/modeling/src/operations/booleans/retessellate.js
```diff
@@ -5,11 +5,21 @@ const poly3 = require('../../geometries/poly3')
 
 const reTesselateCoplanarPolygons = require('./reTesselateCoplanarPolygons')
 
+// Normals are directional vectors of length 0 to 1.0
```
SimonClark, line 8: Is this true in JSCAD? 

Mathematically, normals have a length of exactly 1, although each dimension is between 0 and 1.

z3dev: Good point.

z3dev: Updated comment.

PR #636, opened 2020-08-04
V2 : Web : serialize object using JSON

z3dev:
These changes update the serialization of objects in the WEB UI.

The advantages are:
- all objects are serialized, not just JSCAD objects
- all attributes are serialized, not just JSCAD object data structures
- rendering of polygons with individual colors

The disadvatages are:
- increased memory useage, as JSON strings require more memory

### All Submissions:


z3dev:
@SimonClark wow. thanks for the quick look at the changes.

yeah. there are still lots of tests to create for core function, but that will probably come about as part of some reorganization.

PR #644, opened 2020-08-10
Fixes #520. Warning: this is a substantial breaking change!

### LINKED ISSUE #520: V2 : Transforms are not applied in order
## Expected Behavior

Transforms (scale, rotate, translate, etc) are applied in order, resulting in expected shapes. The following code should produce three shapes, like this.

good

```
const { color, geometry, primitives, transforms } = require('@jscad/modeling')

const main = (params) => {
  let p0 = primitives.rectangle({center: [60.0000, -20.0000], size: [119.8, 39.8]})
  color.color([1, 0, 0, 1], p0)

  let p1 = primitives.roundedRectangle({center: [30.0000, -20.0000], size: [40, 20], roundRadius: 5})
  color.color([0, 1, 0, 1], p0)

  let p2 = primitives.roundedRectangle({center: [20.0000, -10.0000], size: [40, 20], roundRadius: 5})
  p2 = transforms.rotateZ(0.5235987755982988, p2)
  p2 = transforms.translate([70, -21, 0], p2)
  color.color([0, 0, 1, 1], p0)

  return [p0, p1, p2]
}

module.exports = { main }
```
## Actual Behavior

The transforms are accumulated, rather then being applied to the shape. This results in an unexpected shape.

bad

  1. Drap and drop the given code into the web version of V2

## Specifications

  - Version: V2
  - Platform: ALL
  - Environment: ALL

hjfreyer:
Fixes #520. Warning: this is a substantial breaking change!

New matrices applied via the transform() functions of geom2, geom3, and
path2 should multiply with the new matrix on the left. That is, instead
of geom.transform = geom.transform * new_transform, it should be
geom.transform = new_transform * geom.transform.

Before this change, calls to transform() are effectively applied
Last-In-First-Out (stack order). This can lead to strange circumstances
like where a caller tries to move a shape long the X axis, but since
previous code already pushed a "rotate around the Z axis" matrix onto the
"stack", the call to translateX() actually translates along the Y axis,
which is clearly not the user's intent.

Fixed tests, and also changed them a bit to translate differenly along
different axes. This makes the strangness of the behavior more obvious.

This is a breaking change for any code that applies multiple transforms
to a single geometry (except for the special case where they commute).

### All Submissions:

* [Y] Have you followed the guidelines in our Contributing document?
* [Y] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [Y] Does your submission pass tests?

### DIFF
--- a/packages/modeling/src/geometries/geom2/transform.test.js
+++ b/packages/modeling/src/geometries/geom2/transform.test.js
@@ -27,15 +27,15 @@ test('transform: adjusts the transforms of geom2', (t) => {
   t.true(compareVectors(another.transforms, expected.transforms))
 
   // expect lazy transform, i.e. only the transforms change
-  expected.transforms = [0, 1, 0, 0, -1, 0, 0, 0, 0, 0, 1, 0, -5, 5, 5, 1]
-  another = transform(mat4.fromTranslation([5, 5, 5]), another)
+  expected.transforms = [0, 1, 0, 0, -1, 0, 0, 0, 0, 0, 1, 0, 5, 10, 15, 1]
+  another = transform(mat4.fromTranslation([5, 10, 15]), another)
   t.true(comparePoints(another.sides[0], expected.sides[0]))
   t.true(comparePoints(another.sides[1], expected.sides[1]))
   t.true(comparePoints(another.sides[2], expected.sides[2]))
   t.true(compareVectors(another.transforms, expected.transforms))
 
   // expect application of the transforms to the sides
-  expected.sides = [[[-6, 5], [-5, 5]], [[-5, 5], [-5, 6]], [[-5, 6], [-6, 5]]]
+  expected.sides = [[[4, 10], [5, 10]], [[5, 10], [5, 11]], [[5, 11], [4, 10]]]
   expected.transforms = mat4.identity()
   toSides(another)
   t.true(comparePoints(another.sides[0], expected.sides[0]))
@@ -44,8 +44,8 @@ test('transform: adjusts the transforms of geom2', (t) => {
   t.true(compareVectors(another.transforms, expected.transforms))
 
   // expect lazy transform, i.e. only the transforms change
-  expected.transforms = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 5, 5, 5, 1]
-  another = transform(mat4.fromTranslation([5, 5, 5]), another)
+  expected.transforms = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 5, 10, 15, 1]
+  another = transform(mat4.fromTranslation([5, 10, 15]), another)
   t.true(comparePoints(another.sides[0], expected.sides[0]))
   t.true(comparePoints(another.sides[1], expected.sides[1]))
   t.true(comparePoints(another.sides[2], expected.sides[2]))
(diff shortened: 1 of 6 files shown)


z3dev:
@hjfreyer thanks for the pull request.

i tried these changes with some designs ported from V1 to V2, as well as a new V2 design. Previously, I was forcing the application of the transforms using geom2.toSides() / goem3.toPolygons() / path2.toPoints(). After making the changes, these 'hacks' were no longer required.

Awesome!

packages/modeling/src/geometries/geom2/transform.test.js
```diff
@@ -27,15 +27,15 @@ test('transform: adjusts the transforms of geom2', (t) => {
...
+  expected.transforms = [0, 1, 0, 0, -1, 0, 0, 0, 0, 0, 1, 0, 5, 10, 15, 1]
+  another = transform(mat4.fromTranslation([5, 10, 15]), another)
   t.true(comparePoints(another.sides[0], expected.sides[0]))
   t.true(comparePoints(another.sides[1], expected.sides[1]))
   t.true(comparePoints(another.sides[2], expected.sides[2]))
   t.true(compareVectors(another.transforms, expected.transforms))
 
   // expect application of the transforms to the sides
-  expected.sides = [[[-6, 5], [-5, 5]], [[-5, 5], [-5, 6]], [[-5, 6], [-6, 5]]]
+  expected.sides = [[[4, 10], [5, 10]], [[5, 10], [5, 11]], [[5, 11], [4, 10]]]
```
z3dev, line 38: These sides actually make sense now!

packages/modeling/src/geometries/geom3/transform.test.js
```diff
@@ -28,23 +28,23 @@ test('transform: Adjusts the transforms of a populated geom3', (t) => {
...
-  another = transform(mat4.fromTranslation([5, 5, 5]), another)
+  expected.transforms = [6.123234262925839e-17, 1, 0, 0, -1, 6.123234262925839e-17, 0, 0, 0, 0, 1, 0, 5, 10, 15, 1]
+  another = transform(mat4.fromTranslation([5, 10, 15]), another)
   t.true(comparePolygons(another.polygons[0], expected.polygons[0]))
   t.true(compareVectors(another.transforms, expected.transforms))
 
   // expect application of the transforms to the polygons
   expected.polygons = [
-    { vertices: [[-5, 5, 5], [-5, 6, 5], [-5, 6, 6]] }
+    { vertices: [[5, 10, 15], [5, 11, 15], [5, 11, 16]] }
```
z3dev, line 38: A far better expectation!

packages/modeling/src/geometries/path2/transform.test.js
```diff
@@ -27,23 +27,23 @@ test('transform: adjusts the transforms of path', (t) => {
...
-  another = transform(mat4.fromTranslation([5, 5, 5]), another)
+  expected.transforms = [0, 1, 0, 0, -1, 0, 0, 0, 0, 0, 1, 0, 5, 10, 15, 1]
+  another = transform(mat4.fromTranslation([5, 10, 15]), another)
   t.true(comparePoints(another.points, expected.points))
   t.false(another.isClosed)
   t.true(compareVectors(another.transforms, expected.transforms))
 
   // expect application of the transforms to the sides
-  expected.points = [[-5, 5], [-5, 6], [-6, 5]]
+  expected.points = [[5, 10], [5, 11], [4, 10]]
```
z3dev, line 37: Points that make sense!

hjfreyer:
Awesome! Is there anything else I need to do in order to get it merged in?

z3dev:
> Awesome! Is there anything else I need to do in order to get it merged in?

I’ll do a final review tomorrow. And merge if there are no issues.
Should be just fine.

PR #659, opened 2020-08-23
V2 Pre-release - Phase 2

z3dev:
This is a pre-release of V2, which includes changes from two pull requests. Changes can be found via the CHANGELOG files of the affected packages.

Looking good!

### All Submissions:


packages/web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jscad/web",
-  "version": "2.0.0-alpha.0",
+  "version": "2.0.0-alpha.2",
```
SimonClark, line 3: This intentionally went to alpha.2?

z3dev: Correct.

PR #669, opened 2020-09-02
fix unwanted offset when outerRotation applied

SimonClark:
### All Submissions:

Reversed the order of applied transforms, updated test to catch similar issues.


z3dev (approved):
These changes look fine. Nice fix on the documentation.

PR #675, opened 2020-09-06
V2 : fix json import

z3dev:
These changes fix issues when drag-n-drop of exported JSON files, i.e. designs saved as JSON. There were issues with the handling of single files, as well as a missing deserializer for JSON files.

UNDERNEATH... I've been wanting to refactor how the deserializers were registered in the IO module. The following changes have been made:
- create a list of deserializers by extention (new deserializers.js file)
- correct the index of IO module to include the deserializers
- change core transformSources to use the list of deserializers, simplify redundant verbose code
- change core registerExtensions to use the list of deserializers, simplify redundant verbose code
- change cli generateOutputData to use the list of deserializers

Now, there's just one list which can change, adding or removing a deserializer for an external format.

### All Submissions:


packages/core/code-loading/transformSources.js
```diff
@@ -1,10 +1,13 @@
+const { deserializers } = require('@jscad/io')
+
 const isCommonJsModule = require('./isCommonJsModule')
 const modulifySource = require('./modulifySource')
 
 const passThroughTransform = (options, entry) => entry
 
 /* function to turn old style jscad code with implicit imports
 into code with explicit exports/imports */
+// FIXME wouldn't a javascript error be better then 'hacking' the users code?
```
SimonClark, line 10: Yes!  And if there is a time to do this, it's now.

packages/core/code-loading/webRequire.test.js
```diff
@@ -106,10 +106,11 @@ test('webRequire: should allow using require.extensions like the native node req
 
 test('webRequire: should allow using require.extensions like the native node require (parser)', (t) => {
   const registerStlExtension = (fs, _require) => {
-    const deserializer = require('@jscad/io').stlDeSerializer
+    const { deserializers } = require('@jscad/io')
+    const deserializer = deserializers.stl
```
SimonClark, line 110: I'm assuming there is little point optimizing imports to (in this case) only get the STL deserializer?  It doesn't reload the code each time I assume?

packages/io/io/deserializers.js
```diff
@@ -0,0 +1,17 @@
+const amfDeSerializer = require('@jscad/amf-deserializer')
+const dxfDeSerializer = require('@jscad/dxf-deserializer')
+const jsonDeSerializer = require('@jscad/json-deserializer')
+const objDeSerializer = require('@jscad/obj-deserializer')
+// const scadDeSerializer = require('@jscad/scad-deserializer') //FIXME: upgrade, fix before re-enabling
+const stlDeSerializer = require('@jscad/stl-deserializer')
+const svgDeSerializer = require('@jscad/svg-deserializer')
+
+const deserializers = {}
+deserializers[amfDeSerializer.extension] = amfDeSerializer.deserialize
```
SimonClark, line 10: The repetition here bugs me.  Is something like tis too hacky?
```
const deserializerFormats = ["amf", "dxf", "json", "obj", "stl", "svg"]
const deserializers = {}

deserializerFormats.forEach(format => {
	deserializer = require(`@jscad/${format}-deserializer`);
	deserializers[deserializer.extension] = deserializer.deserialize
})
```

packages/io/io/index.js
```diff
@@ -1,5 +1,4 @@
 const { makeBlob, convertToBlob } = require('@jscad/io-utils')
-const prepareOutput = require('./prepareOutput')
 
 const amfSerializer = require('@jscad/amf-serializer')
```
SimonClark, line 3: Seems weird to have them listed here as well.  Can we have only one place where they get registered?

packages/core/code-loading/webRequire.test.js
z3dev: No reload. Just a quick test case for webRequire.

packages/io/io/deserializers.js
z3dev: Well... yes and no.

The repetition is kind of ugly but only here. And there’s no need to know the extensions, as those are exported by the specific deserializer modules.

The suggestion is good but it means the extensions define the module name. I may have to rename a module.

packages/io/io/index.js
z3dev: Yeah. I almost created a list for serializers as well, but decide to keep this focused on deserializers, and not expand scope.

I’ll add a new issue for the serializers cleanup.

SimonClark (approved):
👍

PR #678, opened 2020-09-09
V2 add align() and measureAggregateBoundingBox functions

SimonClark:
### All Submissions:

* [x ] Have you followed the guidelines in our Contributing document?
* [x ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [x ] Does your submission pass tests?

Check out the examples for usage.


packages/modeling/src/operations/transforms/index.js
```diff
@@ -6,6 +6,11 @@
  * const { center, rotateX, translate } = require('@jscad/modeling').transforms
  */
 module.exports = {
+  align: require('./align').align,
+  alignX: require('./align').alignX,
```
z3dev, line 10: I didn't see these 'aliases' defined in align.js

packages/modeling/src/operations/transforms/align.js
```diff
@@ -0,0 +1,87 @@
...
+    } else if (modes[i] === 'lower') {
+      translation[i] = alignTo[i] - bounds[0][i]
+    }
+  }
+
+  return translate(translation, geometry)
+}
+
+/**
+ * Align the given geometries using the given options.
```
z3dev, line 53: We should be very specific about what align means, as there are many kinds of alignments possible. Something like...

Align the boundaries of the given geometries using the given options.

z3dev (changes requested):
I have a few changes suggested, as well as a question or two.

The new align function looks good but I wonder if some of the options are required. I will try out the new example to get a better understanding on how the options interact.

packages/modeling/src/operations/transforms/align.js
```diff
@@ -0,0 +1,87 @@
...
+const validateOptions = (options) => {
+  const defaults = {
+    modes: ['center', 'center', 'lower'],
+    alignTo: [0, 0, 0],
+    grouped: false
+  }
+  options = Object.assign({}, defaults, options)
+
+  if (options.modes.length !== 3) throw new Error('align(): modes must be an array of length 3')
+  if (options.modes.filter(mode => ['center', 'upper', 'lower', 'none'].includes(mode)).length !== 3) throw new Error('align(): all modes must be one of "center", "upper" or "lower"')
```
z3dev, line 14: I don't mind 'upper' and 'lower' however we should be consistent with our bounding box definition. It think this will make these options more understandable.

https://en.wikipedia.org/wiki/Minimum_bounding_box

packages/modeling/src/measurements/measureAggregateBoundingBox.js
```diff
@@ -0,0 +1,31 @@
...
+ * @return {Array} the min and max bounds for the group of geometry, i.e. [[x,y,z],[X,Y,Z]]
+ * @alias module:modeling/measurements.measureAggregateBoundingBox
+ *
+ * @example
+ * let bounds = measureAggregateBoundingBox([sphere(),cube()])
+ */
+const measureAggregateBoundingBox = (...geometries) => {
+  geometries = flatten(geometries)
+  if (geometries.length === 0) throw new Error('measureAggregateBoundingBox: no geometries supplied')
+  const bounds = measureBoundingBox(geometries)
```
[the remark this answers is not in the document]
SimonClark: I felt it useful.  It adds the context of the method that the user called, helps them debug their script.

packages/modeling/src/measurements/measureAggregateBoundingBox.js
```diff
@@ -0,0 +1,31 @@
...
+  if (geometries.length === 1) {
+    return bounds
+  }
+  const result = [[Number.MAX_VALUE, Number.MAX_VALUE, Number.MAX_VALUE], [-Number.MAX_VALUE, -Number.MAX_VALUE, -Number.MAX_VALUE]]
+  return bounds.reduce((result, item) => {
+    for (let i = 0; i < 3; i++) {
```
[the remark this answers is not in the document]
SimonClark: nice 👍

packages/modeling/src/operations/transforms/align.js
```diff
@@ -0,0 +1,87 @@
+const flatten = require('../../utils/flatten')
+const { measureAggregateBoundingBox } = require('../../measurements')
+const { translate } = require('./translate')
+
+const validateOptions = (options) => {
+  const defaults = {
```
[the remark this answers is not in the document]
SimonClark: yeah, I agree, that's cleaner

packages/modeling/src/operations/transforms/align.js
```diff
@@ -0,0 +1,87 @@
...
+    modes: ['center', 'center', 'lower'],
+    alignTo: [0, 0, 0],
+    grouped: false
+  }
+  options = Object.assign({}, defaults, options)
+
+  if (options.modes.length !== 3) throw new Error('align(): modes must be an array of length 3')
```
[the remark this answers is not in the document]
SimonClark: What's the best example of this, so I can copy the pattern?

Ie, does:
```modes:["center"]``` imply ```modes:["center", "none", "none"]```?
```modes: "lower"``` implies ```modes: ["lower", "lower, "lower"]```?

File: packages/modeling/src/operations/transforms/index.js (hunk 1 above)
Comment (SimonClark) [reply packages/modeling/src/operations/transforms/index.js:10]: yeah, I started on them, and it didn't feel quite right separating out the axes.  Thoughts?
```alignX({mode: "lower", alignTo: 0, grouped:true}, ...objects)```
Worth it or not?

Comment (z3dev) [reply packages/modeling/src/operations/transforms/index.js:10]: Nope

File: packages/modeling/src/operations/transforms/align.js (hunk 3 above)
Comment (SimonClark) [reply packages/modeling/src/operations/transforms/align.js:14]: Not quite sure what changes this implies.  What are you thinking?

File: packages/modeling/src/operations/transforms/index.js (hunk 1 above)
Comment (SimonClark) [reply packages/modeling/src/operations/transforms/index.js]: removed

File: packages/modeling/src/operations/transforms/align.js (hunk 3 above)
Comment (z3dev) [reply packages/modeling/src/operations/transforms/align.js:7]: Not sure.

We could change measureBoundingBox to refer ‘lower’ and ‘upper’.

Or

We could change align to refer ‘min’ and ‘max’

But there may be another term as well.

File: packages/modeling/src/operations/transforms/align.js (hunk 7 above)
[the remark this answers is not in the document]
Comment (z3dev) [reply packages/modeling/src/operations/transforms/align.js:6]: Good point! Here some code from translate()

```
  // adjust the offset if necessary
  offset = offset.slice() // don't modify the original
  while (offset.length < 3) offset.push(0)
```

This allows for arrays of one (probably not useful) or two (for 2D translate).

File: packages/modeling/src/operations/transforms/align.js (hunk 3 above)
Comment (SimonClark) [reply packages/modeling/src/operations/transforms/align.js:7]: let's go min,max,center

File: packages/modeling/src/measurements/measureAggregateBoundingBox.test.js [hunk 8]
```diff
@@ -0,0 +1,35 @@
+const test = require('ava')
+
+const { cube } = require('../primitives')
+const measureAggregateBoundingBox = require('./measureAggregateBoundingBox')
+const measureBoundingBox = require('./measureBoundingBox')
+
+test('measureAggregateBoundingBox (single objects)', (t) => {
+  const aCube = cube({ size: 4, center: [4, 10, 20] })
+  const bounds = measureAggregateBoundingBox(aCube)
+  t.deepEqual(bounds, [[2, 8, 18], [6, 12, 22]], 'Bounds were not as expected: ' + JSON.stringify(bounds))
```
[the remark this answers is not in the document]
SimonClark: actually, I took the bounds out of the output, as it was redundant when using deepEqual

packages/modeling/src/utils/padArrayToLength.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Fill an array with a default value, until it reaches a specified length.
+ * @param {Array} arr - the array to potentially add elements to.
+ * @param padding - the element to add to the array to reach the desired length.
+ * @param {Number} targetLength - The desired length of the return array.
+ * @returns {Array} an array of at least 'targetLength" length
+ * @alias module:modeling/utils.padArrayToLength
+ */
+const padArrayToLength = (arr, padding, targetLength) => {
+  arr = arr.slice()
```
z3dev, line 10: This is making a clone of the array. Is that the desired behavior? The comments seem to say otherwise.

packages/modeling/src/utils/padArrayToLength.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Fill an array with a default value, until it reaches a specified length.
+ * @param {Array} arr - the array to potentially add elements to.
+ * @param padding - the element to add to the array to reach the desired length.
+ * @param {Number} targetLength - The desired length of the return array.
+ * @returns {Array} an array of at least 'targetLength" length
+ * @alias module:modeling/utils.padArrayToLength
+ */
+const padArrayToLength = (arr, padding, targetLength) => {
```
z3dev, line 9: Let’s use ‘array’ instead of ‘arr’

packages/modeling/src/utils/padArrayToLength.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Fill an array with a default value, until it reaches a specified length.
+ * @param {Array} arr - the array to potentially add elements to.
+ * @param padding - the element to add to the array to reach the desired length.
```
z3dev, line 4: Hmm... anything is ok... how to say anything in JSDOC?

z3dev (changes requested):
Looking good.

I just have a couple of nits about the new util function.

packages/modeling/src/utils/padArrayToLength.js
SimonClark: {*} apparently

z3dev (approved):
Looks good! Thanks for his nice piece of functionality!

Issue #681, opened 2020-09-10
V2 – Smoothness/Segments control on linear extrude


MathieuLoutre:
## Expected Behavior
When extruding a bezier path, we should have control over how many segments are created for the curves

## Actual Behavior
There seem to be a default and no options to change that

## Steps to Reproduce the Problem
Following the SVG import example, I've written this script which comverts an SVG into geometry, then converts the path2 into polygons/geom2 and then extrudes it (using `geom2` as target for the deserializer gives an out of memory when extruding). 

```
const svg = svgDeSerializer.deserialize({ output: 'geometry', target: 'path2' }, `<svg xmlns="http://www.w3.org/2000/svg"><path d="M249 108C271 168 292 260 305 403L249 403C243 338 234 259 219 194C201 255 181 330 167 403C213 435 259 471 259 575C259 703 190 708 175 708C136 708 85 674 85 557C85 522 90 473 98 419C54 391 13 354 13 238C13 97 66-2 142-2C158-2 185 1 214 39C231 13 250-3 270-3C279-3 290-2 297 1L302 54C275 53 259 84 249 108ZM146 582C146 629 153 651 171 651C193 651 201 617 201 575C201 512 183 480 157 458C150 503 146 545 146 582ZM145 65C93 65 76 193 76 249C76 299 89 330 109 354C127 262 153 164 184 94C173 76 160 65 145 65Z"/></svg>`)

const poly = svg.map((shape) => primitives.polygon({ points: shape.points }))

let shapes = booleans.union(extrusions.extrudeLinear({ height: 10 }, polyLetter))
shapes = transforms.center({}, transforms.mirrorX(transforms.mirrorY(shapes)))

const stl = Buffer.concat(stlSerializer.serialize({ binary: true }, shapes).map((buff) => Buffer.from(buff)))

fs.writeFileSync(`${glyph.name}.stl`, stl, 'binary')
```

Admittedly, I don't fully understand what's going on but in SCAD I can use $fn to control how smooth the curves are. Is there a way to do that in JSCAD? Where is this going wrong? (I suspect it's the conversion into polygon which might ignore curves but I'm not sure what else to convert it into). I also noticed an "extrudeAlongPath" in the repo but I'm not sure how to approach it.

## Specifications

  - Version: 2.0.0-alpha.4
  - Platform: OSX 10.14.6
  - Environment: Node.js

z3dev:
[...]

PR #682, opened 2020-09-10
Add support for segments in SVG Deserializer

MathieuLoutre:

This adds a `segments` argument to the SVG Deserializer. It gets passed down to any curves, specifically: ellipses, circle, rounded rectangles and bezier curves using the existing `segments` argument in each of the primitives. This works both for the script and path output.

The tests add a case for each of these testing the increase of segments. In the case of ellipses and circles, the default used (32) to be different from the default for bezier curves (16). Now, since there's only one argument to control both and the new default is 16, the ellipses, circles and rounded rectangles created through the SVG deserializer will have `segments=16`. 

This PR comes after the discussion in issue #681

### DIFF
--- a/packages/io/svg-deserializer/index.js
+++ b/packages/io/svg-deserializer/index.js
@@ -38,7 +38,8 @@ const deserialize = (options, input) => {
     output: 'script',
     pxPmm: require('./constants').pxPmm,
     target: 'path', // target - 'geom2' or 'path'
-    version: '0.0.0'
+    version: '0.0.0',
+    segments: 16
   }
   options = Object.assign({}, defaults, options)
   return options.output === 'script' ? translate(input, options) : instantiate(input, options)
@@ -55,7 +56,7 @@ const deserialize = (options, input) => {
  * @return {[geometry]} a set of geometries
  */
 const instantiate = (src, options) => {
-  const { pxPmm, target } = options
+  const { pxPmm, target, segments } = options
 
   options && options.statusCallback && options.statusCallback({ progress: 0 })
 
@@ -67,7 +68,7 @@ const instantiate = (src, options) => {
 
   options && options.statusCallback && options.statusCallback({ progress: 50 })
 
-  const result = objectify({ target }, svgObj)
+  const result = objectify({ target, segments }, svgObj)
 
   options && options.statusCallback && options.statusCallback({ progress: 100 })
   return result
@@ -84,7 +85,7 @@ const instantiate = (src, options) => {
  * @return {string} a string (JSCAD script)
  */
 const translate = (src, options) => {
-  const { filename, version, pxPmm, addMetaData, target } = options
+  const { filename, version, pxPmm, addMetaData, target, segments } = options
 
   options && options.statusCallback && options.statusCallback({ progress: 0 })
 
@@ -105,7 +106,7 @@ const translate = (src, options) => {
 
   options && options.statusCallback && options.statusCallback({ progress: 50 })
 
-  const scadCode = codify({ target }, svgObj)
+  const scadCode = codify({ target, segments }, svgObj)
   code += scadCode
   code += '\nmodule.exports = { main }'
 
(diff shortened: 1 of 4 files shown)


z3dev:
@MathieuLoutre you certainly caught on quickly! Super job!

I’ll review tomorrow.
FYI, I reside in Japan. ;)

packages/io/svg-deserializer/index.js
```diff
@@ -38,7 +38,8 @@ const deserialize = (options, input) => {
     output: 'script',
     pxPmm: require('./constants').pxPmm,
     target: 'path', // target - 'geom2' or 'path'
-    version: '0.0.0'
+    version: '0.0.0',
+    segments: 16
```
z3dev, line 42: Let's use segments 32, as that's the value for all the normal primitives, etc.

Note... this will require changes to the test cases.

packages/io/svg-deserializer/index.js
```diff
@@ -67,7 +68,7 @@ const instantiate = (src, options) => {
 
   options && options.statusCallback && options.statusCallback({ progress: 50 })
 
-  const result = objectify({ target }, svgObj)
+  const result = objectify({ target, segments }, svgObj)
```
z3dev, line 71: Let's change to pass all 'options'
`objectify(options, svgObj)`

I forsee additional options in the future.

packages/io/svg-deserializer/index.js
```diff
@@ -105,7 +106,7 @@ const translate = (src, options) => {
 
   options && options.statusCallback && options.statusCallback({ progress: 50 })
 
-  const scadCode = codify({ target }, svgObj)
+  const scadCode = codify({ target, segments }, svgObj)
```
z3dev, line 109: Same

packages/io/svg-deserializer/shapesMapGeometry.js
```diff
@@ -4,12 +4,12 @@ const { svg2cagX, svg2cagY, cagLengthX, cagLengthY, cagLengthP, reflect } = requ
 // const { cssPxUnit } = require('./constants')
 
 const shapesMapGeometry = (obj, objectify, params) => {
-  const { svgUnitsPmm, svgUnitsX, svgUnitsY, svgUnitsV, svgGroups, target } = params
+  const { svgUnitsPmm, svgUnitsX, svgUnitsY, svgUnitsV, svgGroups, target, segments } = params
```
z3dev, line 7: UG... getting too long

packages/io/svg-deserializer/shapesMapGeometry.js
```diff
@@ -4,12 +4,12 @@ const { svg2cagX, svg2cagY, cagLengthX, cagLengthY, cagLengthP, reflect } = requ
...
 
   const types = {
-    group: (obj) => objectify({ target }, obj),
+    group: (obj) => objectify({ target, segments }, obj),
 
-    rect: (obj, svgUnitsPmm, svgUnitsX, svgUnitsY) => {
+    rect: (obj, svgUnitsPmm, svgUnitsX, svgUnitsY, svgUnitsV, svgGroups, segments) => {
```
z3dev, line 12: UG UG... not ideal

z3dev (changes requested):
Great set of changes.
Sorry to spoil everything, but segments 32 should be the default.

packages/io/svg-deserializer/shapesMapGeometry.js
MathieuLoutre: Yeah... This is not great. Given how much pick and choose there is, it would be nice to have that as an object you can destructure but I imagine we'd have to match that across all deserialisers? Maybe a refactor for another time?

MathieuLoutre:
I've updated to match suggestions

packages/io/svg-deserializer/shapesMapGeometry.js
z3dev: Yeah. Another time. ;)

z3dev (approved):
Awesome!
This is a great first PR. I’m totally impressed.

z3dev:
@MathieuLoutre thanks for the extra effort. Your changes were spot on, great.

Please continue to improve JSCAD.

MathieuLoutre:
Thanks so much! Happy to be able to contribute though my 3D knowledge is pretty basic right now :)

PR #685, opened 2020-09-11
V2 docs improvements

SimonClark:
Changes to make the docs look and work a bit better...  Let me know what you think / like.

image


z3dev:
@SimonClark this is good, but I can see where this is heading.

My first reaction is that anyone should be able to add to documentation, especially non-API documentation. I’m a HUGE fan of WIKI, and think that users can contribute changes or content very easily. I’m not so sure about MD files.

Having said that, I’m open to experimenting.

SimonClark:
> My first reaction is that anyone should be able to add to documentation, especially non-API documentation. I’m a HUGE fan of WIKI, and think that users can contribute changes or content very easily. I’m not so sure about MD files.

Yeah, I think wiki based documentation is good goal.  Eventually we need something more robust than jsdoc, or that works with jsdoc to add comments and wiki functionality. I've got some ideas.

This PR is a vanity piece for me. The default style and navigation bugs the crap out of me.  Adding the tutorials and homepage content seemed like a good thing too.

SimonClark:
Now, that said, there is a "Disqus" option in the docdash theme.  It would be somewhat trivial to add on-page discussions.  But... if/when we moved to something more customized, we might lose all existing comments.

SimonClark:
I think more valuable than user-editable docs, in the short term, would be a user-contributable script repository.  Something like Thingiverse, but for jscad scripts.

jsdoc.json
```diff
@@ -34,6 +34,67 @@
...
+    "removeQuotes": ["none"],
+    "scripts": ["jscad-docs.css", "jscad-docs.js"],
+    "menu": {
+      "openjscad.org (v1)": {
+        "href":"https://openjscad.org",
+        "target":"_blank",
+        "class":"menu-item",
+        "id":"website_link"
+      },
+      "JSCad User Group": {
```
z3dev, line 86: Please add V2 as well. Website is www.jscad.xyz

jsdoc/assets/README.md
```diff
@@ -0,0 +1,116 @@
+# JSCAD V2 - Documentation
```
z3dev, line 1: Can this be cleaned up a little? For example, screen shots don't really apply here.

jsdoc/assets/static/jscad-docs.css
```diff
@@ -0,0 +1,187 @@
+body {
```
z3dev, line 1: HA HA HA... you really don't like the default colors.

packages/examples/old/various/logo.js
```diff
@@ -5,20 +5,27 @@
 // tags       : Logo,Intersection,Sphere,Cube,Big
```
z3dev, line 5: Why this change?

z3dev (changes requested):
I have some suggestions for a few small changes. Otherwise, ready to go.

jsdoc.json
SimonClark: Cool!  Will do.

jsdoc/assets/README.md
SimonClark: Yeah.  I'd not turned an eye towards this file yet. Coming soon...

jsdoc/assets/static/jscad-docs.css
SimonClark: it's, uh, well. No. No I don't. Not since 1995.

packages/examples/old/various/logo.js
SimonClark: Hah, used it to generate the background.  I'll revert

z3dev (approved):
Good to go...

PR #698, opened 2020-09-26
V2 web menus

SimonClark:
### All Submissions:

Improves the html structure of the help, editor and options menus
Updates examples to use proper docblocks
Generates examples menu from examples files.


packages/examples/core/booleans/basicBooleans.js
```diff
@@ -1,11 +1,12 @@
...
-// title       : Basic Booleans Demonstration
-// author      : Simon Clark
-// license     : MIT License
-// description : Demonstrating the basics of boolean shape combinations
-// file        : basicBooleans.js
-// tags        : cube, sphere, union, subtract, difference, subtraction, intersect, intersection
-*/
+/**
+ * Basic Booleans Demonstration
+ * @category Manipulating Shapes
```
z3dev, line 3: Awesome! I hope this works out well.

z3dev (approved):
These changes look great. The generation of the list of examples will be a great step forword towards the integration of examples with documentation and applications.

However, this should be merged AFTER THE V2 PRE-RELEASE as I saw a possible conflict.

packages/web/postInstall.js
```diff
@@ -1,6 +1,66 @@
...
+    sortExamples(examples)
+    fs.writeFile('examples/examples.json', JSON.stringify(examples), function (err) {
+      if (err) return console.log(err)
+    })
+  } else {
+    return console.log('Examples directory does not exist: ' + examplesSrc)
+  }
+}
+
+const processExamplesInDirectory = (dir, examples) => {
```
z3dev, line 25: Hmm... just thinking about this a little further.. can the generation of the JSON file become a function of the examples? possibly generated as part of postinstall? and committed to GIT as well?

The reason for this is easy distribution of a release. It would be great to keep the distribution as easy as 'copy from here to there'.

z3dev:
Very small request... can you add 'version' information to the help popup? i believe the 'state' has this information.

packages/web/postInstall.js
SimonClark: We could, but I'm not sure I agree that we should. 

Moving the generation to postinstall.js of the examples is reasonable, however, examples.json doesn't support the examples, it supports the ui.  If we called it examplesMenuData.json, it clearly belongs where it is.

Also committing something that was autogenerated to git is a bit of a smell. If I had untracked changes after bootstrap, I'd assume the devs forgot to add something to .gitignore.

SimonClark:
> Very small request... can you add 'version' information to the help popup? i believe the 'state' has this information.
> 
> UGG... in fact, there's no 'About JSCAD' information. V1 has a nice little popup but not V2. :(

I'll totally add the version.  That's pretty important.

The 'about jscad popup is probably a issue/PR of its own :)

packages/web/postInstall.js
z3dev: Hmm... OK.

Let’s see what can be done to autogenerate and distribute changes. I was hoping for an easy solution but even the API documentation needs to be generated every release. So, maybe there are some additional steps.

z3dev:
> I'll totally add the version. That's pretty important.

Thanks. It helps when new issues are created.

> The 'about jscad popup is probably a issue/PR of its own :)

Agreed

z3dev (approved):
Looks good. I'm eager to see these changes on the website.

PR #703, opened 2020-09-30
fix(core): Remove connectors from public api

SimonClark:
### All Submissions:

Removing connectors for the time being until we can make them actually work.

### DIFF
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -2,7 +2,6 @@
   "source" : {
     "include": [
       "packages/modeling/src/colors",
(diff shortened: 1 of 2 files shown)


z3dev:
This seems reasonable but I think we should have a plan for connectors, or at least work arounds for users that use connectors.

z3dev:
The documentation for connectors should also be disabled.

z3dev (approved):
Ok. Let’s disable for now.

But we need to discuss further with our users on work around, and a new implementation for V2.

PR #709, opened 2020-10-06
V2 : allow other file types to be loaded into the fake FS of the WEB UI

z3dev:
These changes allow other file types (image and font files) to be read into the fake file system. The 'files' are then accessible, and can be used in other libraries, which allows cool stuff to happen. 👍 

Surfaces can be created from images.

Paths can be created from font outlines.

There's a lot more possible as well, but this is a good starting point.

There's also a fix to the webRequire functions. The 'main' attribute of packages can be either a fixed path or a relative path. This fix corrects the use of relative paths.

### All Submissions:

Extensive testing with the WEB UI was performed.


z3dev:
OK. First question...

There's a white list of allowed file types, composed of supported 3D file formats and binary file formats (new). Other file types (extensions) are skipped, not loaded.

The functionality could load anything... as ArrayBuffer. I'm kind of split on this one.

Thoughts?

z3dev:
Code example of using the fake FS...
```
const {primitives} = require('@jscad/modeling')

const { loadFont, loadFontFromData, textToPaths } = require('./jscad-text')

const fs = require('fs')

const main = (params) => {
  const data = fs.readFileSync('project_text/Osaka.ttf')

  const font = loadFontFromData(data)

  const paths = textToPaths({font}, 'JSCAD!　大好きです!')

  return paths
}

module.exports = { main }
```

z3dev:
@SimonClark sorry. i found a late issue, so please wait for awhile.

z3dev:
> @SimonClark sorry. i found a late issue, so please wait for awhile.

OK. After a TON OF TESTING, the late issue was resolved by enhancing the watcher to compare binary files (ArrayBuffers) as well as test files (Strings). And another small bug was identified and squashed as well.

SimonClark:
Holy crap!  This looks awesome!  Been swamped, but I'll get on this...

packages/web/src/sideEffects/localFs/utils.js
```diff
@@ -1,18 +1,50 @@
...
+
 /**
- * returns list of files whose source has changed compared to the file with the same
- * name in memFs
- * @param {Object} reference - The reference hash of files, organized by name.
- * @param {Array} files - The list of files : need to have a source & name field.
- * @return array of files that changed
+ * Compare the contents of the reference and the new file lists.
+ * @see flattenFiles() below
+ * @param {Array} reference - the reference list of files
```
SimonClark, line 25: Feel like the variable names here could make this more readable:
changedFiles => areFilesChanged or areFilesDifferent
reference => referenceFileList or fileList1
files => newFileList or fileList2

packages/web/src/sideEffects/localFs/utils.js
```diff
@@ -1,18 +1,50 @@
...
+ * @return {Array} a list of the changed files
  */
-const changedFiles = (reference, files) => files.filter((file) => {
-  const matches = reference.filter((ref) => ref.fullPath === file.fullPath && ref.source !== file.source)
-  return matches.length > 0
-})
+const changedFiles = (reference, files) =>  {
+  const working = reference.slice()
+  return files.filter((file) => {
+    const found = working.findIndex((a, i) => sameFile(a, file))
```
SimonClark, line 32: also feel like this could be quite inefficient. It's doing full file comparisons over and over.  Wonder if we can hash and cache, or something similar.

packages/web/src/sideEffects/localFs/utils.js
```diff
@@ -1,18 +1,50 @@
...
-const flattenFiles = (files) => {
+/*
+ * Create a flattened list of files from the given file heiarchy.
+ * @return {Array} a list of file entries
+ */
+const flattenFiles = (hierarchy) => {
   let result = []
-  files.forEach((fileOrFolder) => {
+  hierarchy.forEach((fileOrFolder) => {
     if (fileOrFolder.children) {
```
SimonClark, line 48: Not super familiar with this.  Does it still work with empty folders?

SimonClark (dismissed):
Minor changes suggested.  Merge or update at your discretion.

packages/web/src/sideEffects/localFs/utils.js
z3dev: The list of references is reduced if there’s a match. Not perfect but far better then before. It actually works too. 👍

packages/web/src/sideEffects/localFs/utils.js
z3dev: Good question... I try it but I don’t think an empty folder will initiate a ‘watch’ event.

packages/web/src/sideEffects/localFs/utils.js
z3dev: Actually, changedFiles is correct, as a list of changed files is returned.

I’ll review the other names...

SimonClark: ah, gotcha.  👍

SimonClark (approved):

PR #714, opened 2020-10-15
V2 : io revisited

z3dev:
These changes standardize the IO packages for:
- metadata strings, e.g. Created by JSCAD
- documentation

In addition, the IO utility objects have been re-worked to ES5 classes, and test suites have been created. 👍 

### All Submissions:

Testing with both CLI and WEB UI was conducted as well.


z3dev:
@SimonClark there are changes in the IO packages to create documentation. i haven't completed all the required doc string changes yet, as i would like you to reveiw the generated documentation, only for amf deserializer and serializer.

What do you think?

z3dev:
@SimonClark did you have a chance to look at the generated documentation for the IO libraries?

SimonClark:
Sorry, totally snowed under with work.  Crazy times.  Things are easing up now, and I should get to look at this soon.

z3dev:
> Sorry, totally snowed under with work. Crazy times. Things are easing up now, and I should get to look at this soon.

Thought that might be the case. No problem. This will go into the next release as there are still some changes coming.

packages/io/dxf-serializer/tests/geom2ToDxf.test.js
```diff
@@ -46,7 +46,7 @@ test('2D GEOMETRY to DXF POLYLINE', (t) => {
 })
 
 const empty = `999
-DXF generated by JSCAD
+Created by JSCAD
```
SimonClark, line 49: Lot's of repeat here.  Worth pulling this out into a constant, maybe?

packages/io/io-utils/BinaryReader.js
```diff
@@ -1,51 +1,58 @@
...
-  readUInt16: function () { return this._decodeInt(16, false) },
-  readInt32: function () { return this._decodeInt(32, true) },
-  readUInt32: function () { return this._decodeInt(32, false) },
-
-  readFloat: function () { return this._decodeFloat(23, 8) },
-  readDouble: function () { return this._decodeFloat(52, 11) },
-
-  readChar: function () { return this.readString(1) },
-  readString: function (length) {
+  readInt8 () { return this._decodeInt(8, true) }
```
SimonClark, line 23: Should these be arrow functions to be consistent with the rest of JSCAD?

packages/io/io-utils/Blob.js
```diff
@@ -0,0 +1,132 @@
...
+ */
+class Blob {
+  /**
+   * Returns a newly created Blob object which contains a concatenation of all of the data in the given contents.
+   * @param {Array} contents - an array of ArrayBuffer, or String objects that will be put inside the Blob.
+   */
+  constructor (contents, options) {
+    // make the optional options non-optional
+    options = options || {}
+    // number of bytes
```
SimonClark, line 31: couple of readability comments here...
- Not clear if the comment goes with the line above or below.
- the comments themselves are a smell. Instead of 
```
// media type
this.type = ''
// readability state (CLOSED: true, OPENED: false)
this.isClosed = false
```
why not 
```
this.mediaType = ''
this.isUnreadable = false
```

SimonClark (commented):
some initial comments.  Still more to go through.

packages/io/io-utils/BinaryReader.js
z3dev: Will give it a try but these are ‘methods’ on a ‘class’, which is different then arrow functions which have no ‘this’ context.

It’s a strange language, JavaScript.

packages/io/dxf-serializer/tests/geom2ToDxf.test.js
z3dev: Yeah. Lots of different implementations, so lots of duplication.
The problem is where to put the constant. There is only io/io
Suggestions?

packages/io/io-utils/Blob.js
z3dev: I think those came from the JavaScript specifications. Will double check...

packages/io/io-utils/BinaryReader.js
SimonClark: Damn, yeah, you're right.  *shakes fist* Javascript!

packages/io/io-utils/Blob.js
z3dev: Yup. Those attribute names are from the specification.

https://w3c.github.io/FileAPI/#constructorBlob

I tweaked the doc strings, which hopefully make those values slightly clearer.

packages/io/dxf-serializer/tests/geom2ToDxf.test.js
z3dev: Although, a common value is possible. I'm opting for no change.
The new phrase is shorter! 👍

SimonClark (approved):

PR #717, opened 2020-10-16
V2 rotate/pan throttling change proposal

hrgdavor:
### All Submissions:

changed rotate and pan to use requestAnimationFrame (removed need for cancelAnimationFrame I used in my first attempt at this)

this change also works well when dev console is active in chrome (V1 and V2 on chrome with dev console slow down the rotation significantly )

### DIFF
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -8,7 +8,7 @@ const perspectiveCamera = cameras.perspective
 const orbitControls = rendererStuff.controls.orbit
 
 // params
-const rotateSpeed = 0.003
+const rotateSpeed = 0.002
 const panSpeed = 1
 const zoomSpeed = 0.08
 
@@ -17,7 +17,6 @@ let render
 let viewerOptions
 let camera = perspectiveCamera.defaults
 let controls = orbitControls.defaults
-controls.drag = 1
 let rotateDelta = [0, 0]
 let panDelta = [0, 0]
 let zoomDelta = 0
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -26,7 +26,7 @@ let render
 let viewerOptions
 let camera = perspectiveCamera.defaults
 let controls = orbitControls.defaults
-controls.drag = 1;
+controls.drag = 1
 
 const grid = { // grid data
   // the choice of what draw command to use is also data based
@@ -66,6 +66,7 @@ const viewer = (state, i18n) => {
     let delta = [0,0];
     let deltaPan = [0,0];
     let timer;
+
     // rotate
     gestures.drags
       .forEach((data) => {
@@ -82,7 +83,7 @@ const viewer = (state, i18n) => {
         if(!timer) timer = requestAnimationFrame(doRotatePan);
     })
 
-    function doRotatePan(){
+    const doRotatePan = () => {
       timer = null;
       if(delta[0] || delta[1]){      
         const updated = orbitControls.rotate({ controls, camera, speed: rotateSpeed }, delta)
@@ -96,24 +97,6 @@ const viewer = (state, i18n) => {
         camera.target = updated.camera.target
       }
     }
-
-    // pan
-    // gestures.drags
-    //   .forEach((data) => {
-    //     const ev = data.originalEvents[0]
-    //     const shiftKey = (ev.shiftKey === true) || (ev.touches && ev.touches.length > 2)
-    //     if (shiftKey) {
-    //       const now = Date.now()
-    //       const ms = now - lastPan
-    //       if (ms > (1000 / panRate)) {
-    //         const delta = [data.delta.x, data.delta.y].map((d) => d)
-    //         const updated = orbitControls.pan({ controls, camera, speed: panSpeed }, delta)
-    //         camera.position = updated.camera.position
-    //         camera.target = updated.camera.target
-    //         lastPan = now
-    //       }
-    //     }
-    //   })
 
     // zoom
     gestures.zooms
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -83,19 +83,22 @@ const viewer = (state, i18n) => {
         if(!timer) timer = requestAnimationFrame(doRotatePan);
     })
 
-    const doRotatePan = () => {
+    const doRotatePan = (timestamp) => {
       timer = null;
       if(delta[0] || delta[1]){      
         const updated = orbitControls.rotate({ controls, camera, speed: rotateSpeed }, delta)
         delta = [0,0];
         controls = { ...controls, ...updated.controls }
       }
+      
       if(deltaPan[0] || deltaPan[1]){      
         const updated = orbitControls.pan({ controls, camera, speed: panSpeed }, deltaPan)
         deltaPan = [0,0];
         camera.position = updated.camera.position
         camera.target = updated.camera.target
       }
+
+      updateAndRender(timestamp)
     }
 
     // zoom
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -26,6 +26,7 @@ let render
 let viewerOptions
 let camera = perspectiveCamera.defaults
 let controls = orbitControls.defaults
+controls.drag = 1;
 
 const grid = { // grid data
   // the choice of what draw command to use is also data based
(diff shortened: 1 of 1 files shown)


packages/web/src/ui/views/viewer.js
```diff
@@ -26,6 +26,7 @@ let render
 let viewerOptions
 let camera = perspectiveCamera.defaults
 let controls = orbitControls.defaults
+controls.drag = 1;
```
SimonClark, line 29: first thought is jscad uses standard.js style rules, which specifies no EOL semi-colons, spaces between "}else{", etc.

packages/web/src/ui/views/viewer.js
```diff
@@ -63,39 +63,57 @@ const viewer = (state, i18n) => {
...
+          deltaPan[1] += y;          
+        }else{
+          delta[0] -= x;
+          delta[1] -= y;
         }
-      })
+        if(!timer) timer = requestAnimationFrame(doRotatePan);
+    })
+
+    function doRotatePan(){
```
SimonClark, line 85: also arrow functions by default.

const doRotatePan = () => {

packages/web/src/ui/views/viewer.js
```diff
@@ -63,39 +63,57 @@ const viewer = (state, i18n) => {
...
+      }
+      if(deltaPan[0] || deltaPan[1]){      
+        const updated = orbitControls.pan({ controls, camera, speed: panSpeed }, deltaPan)
+        deltaPan = [0,0];
+        camera.position = updated.camera.position
+        camera.target = updated.camera.target
+      }
+    }
+
+    // pan
```
SimonClark, line 100: Unless there is a firm plan to re-enable this code on a finite timeline, commented out code should be removed.

hrgdavor:
How does this process work ?
I have commited the changes based on your suggestions to the same branch.
Do I need to reapply pull request, or the new changes are no considered relevant for the merge ?

z3dev:
> How does this process work ?
> I have commited the changes based on your suggestions to the same branch.
> Do I need to reapply pull request, or the new changes are now considered as basis for a possible merge ?

Please keep this branch active, and continue to make changes to it.
And of course, discuss the changes with us. 👍

packages/web/src/ui/views/viewer.js
```diff
@@ -63,39 +63,40 @@ const viewer = (state, i18n) => {
...
-            lastPan = now
-          }
+          deltaPan[0] += x;
+          deltaPan[1] += y;          
+        }else{
+          delta[0] -= x;
+          delta[1] -= y;
         }
-      })
+        if(!timer) timer = requestAnimationFrame(doRotatePan);
```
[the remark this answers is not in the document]
z3dev: Having said the above, I can understand that the refresh rate of the rendering is fixed, in the same way using animation frame. So, it may make sense to limit the delta (camera) changes to the same rate.

Does this make sense?

z3dev (commented):
Let’s discuss.
I like simple code, so these changes are interesting.
But there has to be some reasons for using animation frame further.

packages/web/src/ui/views/viewer.js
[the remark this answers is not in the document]
hrgdavor: This code does not skip events (just skips calling rotate/pan for burst of events) so in that regards is better in following the mouse :) .

The reason to use requestAnimationFrame is to let browser decide the rate.
Browser does have own rate if CPU is not very busy, but can also reduce framerate further when under heavy load.

I did try it on my Android phone and seems to be working well, even better than jscad.xyz
also on my laptop touchpad, very smooooth :)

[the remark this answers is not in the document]
z3dev: Good to know that testing is with other devices. 👍 

OK. So, can the two loops be combined? I don't see why two loops would be required.

Also, have you tried removing the 'throttles' in updateAndRender? I'm wondering are necessary now.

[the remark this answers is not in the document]
hrgdavor: I have already combined drag and pan loops since they source same events
```
    gestures.drags
      .forEach((data) => {
```
I can remove throttles in updateAndRender but need to tweak code a bit so zoom/pan does not cause it to be called twice in the same animationFrame while rotation is happening.

just showing a grid on mobile is already hard enough for my phone, an in that case new code is no worse than old.

Also further consideration would be making updateAndRender be called only when needed instead on a timer.

I will also change zoom to use this principle (zoom to fit does not need this).

packages/web/src/ui/views/viewer.js
```diff
@@ -63,39 +63,43 @@ const viewer = (state, i18n) => {
...
+      }
+      
+      if(deltaPan[0] || deltaPan[1]){      
+        const updated = orbitControls.pan({ controls, camera, speed: panSpeed }, deltaPan)
+        deltaPan = [0,0];
+        camera.position = updated.camera.position
+        camera.target = updated.camera.target
+      }
+
+      updateAndRender(timestamp)
```
z3dev, line 101: Hmm... there may be some issues here, as updateAndRender() can be heavy if there are lots of polygons.

hrgdavor: well I have just puched draft commit without throttling based on ms,
and changed zoom to use new concept, and examples\core\primitives\primitives3D works much smoother on my phone now than on current live verzion jscad.xyz.

for this change to be final I would need to clean-up unused variables.

further throttling options could be considered, but browser will not force animation at 60 FPS r whatever if it is not able to catchup.

packages/web/src/ui/views/viewer.js
```diff
@@ -80,10 +82,33 @@ const viewer = (state, i18n) => {
           delta[0] -= x;
           delta[1] -= y;
         }
-        if(!timer) timer = requestAnimationFrame(doRotatePan);
+        if(!timer) {
+          cancelAnimationFrame(renderTimer);
```
hrgdavor, line 86: I am making sure that updateAndRender is called only once in an animationFrame. cancel is called because doRotatePanZoom function will call updateAndRender

packages/web/src/ui/views/viewer.js
hrgdavor: I have tried reversing the logic (did not commit/push this experiment):

 - zoom/pan/roate events are collected only without calling update for next animation frame
 - animation frames are used but throttled (current default 10 FPS)
 - inside animation frame both doRotatePanZoom and updateAndRender are called
   (calling doRotatePanZoom is ok as it will do nothing if no events have been collected)

this throttling did not help at all in situation with too many stuff to render 
(on my phone:  "Show grid" plus "3d primitives example" is enough to make it slow)

when poly count was manageable throttling only made it all look less performant.

hrgdavor: I have noticed that app works much better in landscape.
on my phone resize function is messed up.
so canvas is made too big when in portrait mode.

I will look into this further but will split that into a separate pull request.

Now comparing the performance of new code vs jscad.xyz in landscape ... new code is much smoother.

packages/web/src/ui/views/viewer.js
```diff
@@ -117,10 +108,34 @@ const viewer = (state, i18n) => {
...
+        camera.target = updated.camera.target
+      }
+
+      if(zoomDelta){
+        const updated = orbitControls.zoom({ controls, camera, speed: zoomSpeed }, zoomDelta)
+        controls = { ...controls, ...updated.controls }
+        zoomDelta = 0;
+      }
+
+      updateAndRender(timestamp, true)
```
z3dev, line 132: Afer looking at the code, I think adding updateAndRender() here was a mistake. Sorry.
This could potentially create several concurrent requestAnimationFrame() loops, which would definitely effect performance.

z3dev:
@hrgdavor please use lint to clean up the code once complete. it's fairly easy to adjust the project level lint script to lint a single directory.

`    "lint": "standardx 'packages/**/*.js'",`

Just change the script to (or whatever you want to lint)

`    "lint": "standardx 'packages/web/src/ui/views/**/*.js'",`

z3dev:
@hrgdavor I tried out these changes with Firefox and Chrome on MacOS and iOS. the refresh rates are cleaner and sharper, and very simular to V1 now.

I also tried with a large design and didn't find any issues rotating, panning, or zooming.

Basically, this should be a good set of changes. Please clean up the code as requested.

packages/web/src/ui/views/viewer.js
```diff
@@ -63,50 +63,41 @@ const viewer = (state, i18n) => {
...
-          if (ms > (1000 / rotateRate)) {
-            const delta = [data.delta.x, data.delta.y].map((d) => -d)
-            const updated = orbitControls.rotate({ controls, camera, speed: rotateSpeed }, delta)
-            controls = { ...controls, ...updated.controls }
-            lastRotate = now
-          }
-        }
-      })
-    // pan
+    let delta = [0,0];
```
z3dev, line 66: Please use consistent names, like rotateDelta, panDelta, zoomDelta.

packages/web/src/ui/views/viewer.js
hrgdavor: Yeas, I was trying out ideas to see how the interaction feels, to make it more responsive to mouse inputs.

z3dev (approved):
Yeah!
Those last changes were excellent!

I’ll try out the changes tomorrow. But this looks good to go.
@SimonClark still needs to review.

hrgdavor:
wooot, my first contribution :) ... also with these changes, I feel better creating stuff in V2
(extrudeFromSlices gives me more possibilities, and I have made workflow for myself that runs smooth now with V2 too)
next in line is likely livereload support, as I am most motivated for sth like that (sth I need/want badly :D ).

SimonClark (dismissed):

z3dev:
Hmm... I tried the changes with and without...

`controls.drag = 1`

I tend to prefer without. The inertia is very minimal now. Thoughts?

hrgdavor:
> I tend to prefer without. The inertia is very minimal now. Thoughts?
yes, with default drag and rotateSpeed = 0.002 is definitely better on PC for me too

I tried mobile, and it looks ok too.

So, I agree, drag=1 is not needed with these changes.

z3dev (approved):
Thanks for the late change.

Issue #403, opened 2018-10-01
Include TypeScript Declaration Files


ykomatsu:
## Expected Behavior

OpenJSCAD.org includes updated TypeScript Declaration Files.

VS Code IntelliSence for **JavaScript** uses **TypeScript** Declation Files.

https://code.visualstudio.com/docs/languages/javascript#_intellisense

DefinetelyTyped has TypeScript Declaration Files for OpenJSCAD.org, but they look a bit old.

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/openjscad

## Actual Behavior

OpenJSCAD.org does not include TypeScript Declaration Files.

## Steps to Reproduce the Problem

... snip ...

## Specifications

... snip ...

z3dev:
@ykomatsu thanks for the information.

TypeScript support is not planned for OpenJSCAD, so those definitions at DefinitelyTyped were created by someone else. Maybe you can find the original author via the user group.

danmarshall:
Hi @ykomatsu, I wrote the original typings. Since JSCAD is under active development, I have an interim set of typings on https://github.com/danmarshall/jscad-typings . It depends on which modules you are using. Here's how these are used in Maker.js: https://github.com/Microsoft/maker.js/blob/master/packages/maker.js/tsconfig.json#L7
Let me know if I can provide more info.

z3dev:
@danmarshall Thanks again!

ykomatsu:
@z3dev @danmarshall Thank you!

I just want to write `*.jscad` files with VS Code IntelliSense.
Probably, I need TypeScript Declaration Files of https://github.com/jscad/csg.js.

I wonder if I can convert JavaScript files to TypeScript Declaration Files, like JSDoc:

https://github.com/jscad/csg.js/blob/master/docs/api.md

z3dev:
A few of the hard core TypeScript users are looking at this, but we don’t have the band width to work on TypeScript. Hopefully, future JavaScript standards will bring the things closer. Until then, please feel free to chat on this thread.

@danmarshall May have a few more tips for you.

kaosat-dev:
@danmarshall yes, it is jsdoc generated, but I sadly discovered recently that is not 100% complete (blame it on the remaining mess of V1) , it will be 100% accurate once V2 is done (the actual api / what is exported will be much clearer & cleaner)

z3dev:
[...]

PR #724, opened 2020-10-30
[experiment] v2: generate typescript declarations from jsdoc

ahdinosaur:
hallo! :smiley_cat: 

i'm quietly in love with OpenJSCAD, been eagerly awaiting the new v2 release. :tada: 

i've also become a fan of TypeScript in the last couple years, so would love to help contribute TypeScript types here.

this is an super early pull request to get feedback, more of a conversation starter than anything. :seedling: 

in https://github.com/jscad/OpenJSCAD.org/issues/403 there was a mention about generating types from the (jsdoc) code, so i thought i'd start with that, and that's what i tried here. turns out this is a [recent feature built-in to TypeScript](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html).

there are some caveats though with how the jsdoc needs to be written: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html, specifically with `@typedef`, so i didn't want to go any further until i confirmed this is the preferred approach and that this doesn't break the jsdoc documentation generated. i tried to write enough of an example so you see what i mean.

the alternative approach would be to manually write `.d.ts` files alongside the `.js` files. while this might seem like more work, in many ways i think this could be easier, as TypeScript wouldn't impede on jsdoc, and would mean we could write the types in a more TypeScript fluent way. this is for example how i believe Three.js does their TypeScript types. so another question is whether y'all would be keen to include `.d.ts` files within the code base.

anyways, curious to hear what y'all think. cheers! :purple_heart:

### DIFF
--- a/packages/modeling/src/geometries/geom2/type.d.ts
+++ b/packages/modeling/src/geometries/geom2/type.d.ts
@@ -3,7 +3,7 @@ import Mat4 from '../../maths/mat4/type'
 
 export interface Geom2 {
   sides: Array<[Vec2, Vec2]>
-  transforms?: Mat4
+  transforms: Mat4
 }
 
 export default Geom2
--- a/packages/modeling/src/geometries/geom2/type.d.ts
+++ b/packages/modeling/src/geometries/geom2/type.d.ts
@@ -1,0 +1,9 @@
+import Vec2 from '../../maths/vec2/type'
+import Mat4 from '../../maths/mat4/type'
+
+export interface Geom2 {
+  sides: Array<[Vec2, Vec2]>
+  transforms?: Mat4
+}
+
+export default Geom2
--- a/package.json
+++ b/package.json
@@ -8,6 +8,9 @@
     "coverage": "npm run bootstrap && lerna run --concurrency 1 --stream coverage",
     "test": "npm run bootstrap && lerna run --stream test",
     "lint": "standardx 'packages/**/*.js'",
+    "typescript:build": "tsc --project tsconfig.build.json",
+    "typescript:clean": "rimraf 'packages/**/!(type).d.ts'",
+    "typescript": "npm run typescript:clean && npm run typescript:build",
     "bootstrap": "lerna bootstrap --hoist nyc --hoist ava --ignore-prepublish --no-ci",
     "changed": "lerna changed",
     "clean": "lerna clean",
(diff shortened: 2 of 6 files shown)


z3dev:
@ahdinosaur WOW! Thanks for taking this 'little' project on! I think lots of TS fans will appreciate this contribution.

@danmarshall is a huge fan of TS as well, so let's see if he has a few words of wisdom.

I also say the TS files for Three.js. It looked liked those files are being generated. Is that right?

z3dev:
For reference, there's an open issue against the CSG project (defunct) which talks about TypeScript. Some of the comments are pretty biased but there are some points to consider. csg/#164

ahdinosaur:
> I also say the TS files for Three.js. It looked liked those files are being generated. Is that right?

as far as i can tell, the Three.js `.d.ts` files are not generated, they are manually updated. see also https://github.com/mrdoob/three.js/issues/20299.

z3dev:
> > I also say the TS files for Three.js. It looked liked those files are being generated. Is that right?
> 
> as far as i can tell, the Three.js `.d.ts` files are not generated, they are manually updated. see also [mrdoob/three.js#20299](https://github.com/mrdoob/three.js/issues/20299).

Very interesting indeed. I think the Three.js project is suffering from documentation bloat, as ALL the documention is inside the one project, including API documentation, example documentation, etc. It all adds up quickly and Three.js is now a HUGE PIG repository.

I still support TS definitions for the API, as that makes JSCAD easy to use for those in the TS environments.
But I think some more thought needs to go into documentation for examples, user guides, etc. (Another issue, another day)

packages/modeling/src/geometries/geom2/type.d.ts
```diff
@@ -0,0 +1,9 @@
+import Vec2 from '../../maths/vec2/type'
+import Mat4 from '../../maths/mat4/type'
+
+export interface Geom2 {
+  sides: Array<[Vec2, Vec2]>
+  transforms?: Mat4
```
z3dev, line 6: @ahdinosaur any ideas why transforms is declared as 'optional'?

ahdinosaur: oops, that's my bad, i misread the `create` function!

ahdinosaur:
closing in favor of https://github.com/jscad/OpenJSCAD.org/pull/726

PR #727, opened 2020-10-31
V2 : fix extrude rotate

### LINKED ISSUE #719: v2: hangs when using extrudeRotate, polygon and union
I have a minimal code to reproduce the issue.

```js
const { primitives, booleans, extrusions } = require('@jscad/modeling');
const { union } = booleans;
const { extrudeRotate } = extrusions;
const { polygon, cylinder } = primitives;

module.exports.main = () => {
  const points = [
    [0, 0],
    [10, 0],
    [10, 5],
    [0, 5]
  ]

  const p = extrudeRotate({}, polygon({ points }));
  const c = cylinder({radius: 2, height: 20} );

  // return [p, c];
  return union(p, c)
}
```

without the union(return them in an array), it renders like this:

image

## Expected Behavior
It should render properly (looks the same as the version without union)

## Actual Behavior
It hanges forever and cpu usage is high.

The issue actually happens for all boolean operation on the object returns from extrudeRotate

  1. download the reproduce file 
[reproduce.js.zip](https://github.com/jscad/OpenJSCAD.org/files/5408797/reproduce.js.zip)

  2. unzip it and drag it into v2 website: https://www.jscad.xyz/
  3. you will see that it doesn't render.
  4. replace `return union(p, c)` with `return [p, c]`, it renders properly.

## Specifications

  - Version: v2
  - Platform: mac
  - Environment:  browser Chrome Version 86.0.4240.80

z3dev:
These changes correct extrudeRotate() when extruding shapes that overlap the X axis (-/+ points), and corrects the extruding of walls between slices.

Fixes #719 and more..

### All Submissions:


SimonClark (approved):
Looks good.  Gonna trust the math :)

SimonClark (approved):

z3dev:
> Looks good. Gonna trust the math :)

I had the same thoughts, but feel that the math pieces are really solid.
Thanks for the review!

Issue #620, opened 2020-07-07
Feature/brainstorm: workspace for multipart project with layouts


hrgdavor:
moved to https://github.com/jscad/OpenJSCAD.org/discussions/823

z3dev:
@hrgdavor thanks for these excellent thoughts.

I’m sure that you will find that JSCAD is missing some functionality. Please feel free to add comments here, or open a specific issue.

One area that I recently bumped into was “align”. I wanted to align two parts at the bottom.

PR #729, opened 2020-11-04
V2 feature livereload

hrgdavor:
### All Submissions:

woot my second pull request !!

This is an initial basic implementation of livereload support. 
 - It does not try to reconnect
 - no granularity (just allows regular reload)
 - currently, you need to start lievereload server before drag/dropping files

Code works by using setInterval with 50ms waiting for watcherDelay to pass.
This enables current reload functionality that loads periodically to be the fallback.
A message from livereload resets the delay so project reloading can be done asap.

Another benefit of using livereload is that there is no constant reading of project files,
which can be a win for large projects too, not just a speed boost for small ones.

Implementation was tested using node module livereload, but should not be limited to other livereload implementations.

Additionally a line was commented in ui/flow/design.js:428
```
.delay(1000)// FIXME delay is a hack
```
for my simple case where I load single file reloading still works, and is pretty damn fast :) 
that delay removal does not have to be part of the final version of this pull request(I do not know what it was fixing).


z3dev:
@hrgdavor pretty cool. i can understand why you want this functionality but have a few thoughts.

First, the V2 WEB UI is really built for the website (www.jscad.xyz). So, the auto-reload should work on the website for all design projects. I took a quick look at the livereload site, and from my understanding a 'local' file system is require. Is the livereload usable on the website?

Second, livereload seems to require registration by users. although some of the users might want to use this feature, asking everyone to register won't be feasible. do you know if registration is required?

Third, you mentioned a server. Can you provide some information on how to start and manager the livereload server?

hrgdavor:
@z3dev 

### TLDR 
 - it will work for local instance as well for jscad.xyz 
 - there is a free livereload server in form of a npm module
 - this pull request can be expanded to allow loading only changed files for large projects (more speed :) )
 - I would prefer a separate pull request for the mentioned "more advanced version"

### Longer version.
Livereload is simple but grea tidea from Andrey Tarantsov (I think). Although his website links to Mac app store and it looks like it is only commercial, there are free implementations of the protocol and the one I often use is https://www.npmjs.com/package/livereload

it is simple to install globally, or as dev dependency
```
npm -i -g livereload
```
If you look closely in console when you run openjscad from source `npm run dev` in `packages/web` you are using a builtin Livereload
```sh
OpenJSCAD.org\packages\web> npm run dev
> @jscad/web@2.0.0-alpha.11 dev C:\wrk\OpenJSCAD.org\packages\web
> budo demo.js --port=8081 --css ./css/demo.css --live -- -b

[0004] info  Server running at http://192.168.1.105:8081/ (connect)
[0004] info  LiveReload running
```
Luckily that livereload is not using the default port(`35729`) so it was trivial to use livereload commandline for refreshing the design files.

Steps are really simple for the openjscad users:
 - open shell in the folder with design files 
 - run `livereload` 
 - drag and drop the design into openjscad browser app as before

The reason it works is because livereload uses websockets which are not limited by cross origin access. So, regardless if I run local instance of openjscad or am using jscad.xyz the connection is always to `ws://127.0.0.1:35729/livereload`

This initial version can be finished very quickly bringing the benefit of reload that reacts to file change instead of periodically loading all desing files to check their contents and then refreshing the view.

**I would like to keep/finish this simple version first so it can be pushed soon to V2 and jscad.xyz.**

Next step would be a bigger change that woul be a separate pull request.
A more advanced version that works for big designs with multiple files.
When files change in big design:
 - livereload is leveraged so only changed files are read from fs
 - after the changed files are loaded, view is refreshed

hrgdavor:
@z3dev did you have time to try this out ?

z3dev:
@hrgdavor I've been thinking about this for a while.

**First**, the websites (www.jscad.xyz and www.openjscad.org) are currently housing WEB applications for beginners.
- there's a built-in editor for writing designs
- there's examples to browse and change
- there's links to resources
- etc

I even think these sites have TOO MUCH functionality for beginners.
- auto-reload
- loading of projects (not single file designs)
- experimental caching
- etc

So, my first thought is that the current website should be simplified, and kept simple for starters. There has even been talk about moving the WEB applications to another page, and create a 'real' website for JSCAD with introduction materials, getting started documentation, api documentation, project information, etc.

That brings up the **Second** point. For those managing full blown projects, what are the current options? Well... none, except the CLI, which kind of sucks. There's the V1 DESKTOP application but that hasn't been ported to V2 JSCAD.

So, there's kind of a BIG HOLE now for those wanting to use projects.

I think some thought and effort needs to go into a new application that is focused on projects. And the pieces exist to create something really great.
- the viewer is a module, and can be placed in a web page anywhere
- the core logic is a module, and can be used to load and compile JSCAD designs
- the modeling module can be used together or independently

So, I think this PR would be best as part of the new application.

So, what shall we build? Desktop app? Web app? IDE plugin? All of these?

udif:
Please do not "dumb down" the web application. It is not only for beginners. I find the fact that there is already a hosted webapp extremely useful. I can share designs with other users using it, and I don't have to host my own.

As for projects, a VSCODE plugin seems ideal to me. I think it makes the desktop app redundant.

hrgdavor:
This PR is pretty simple and small, and benefits single file designs for sure, and will likely big projects as well. I wonder if you tried ti with multifile projects. For me a big part of how I feel while using a tool is how quickly my changes are applied.

Learning even few basic things about jscad itself is from my point of view much harder than installing a npm module, and running livereload command in project folder (before drag&drop and starting to change the design files).

It sounds nice to create a new app for full blown projects, but it will take time. It would be nice to have faster loading with what we have here while we wait :) .

If you can give me source of full blown project (best one that you are planing to have open source) I could look into it and get a better idea. I have proposed a second PR after that would focus on loading big projects, because livereload would I think make it much snappier(the bigger the project, more benefit you have from loading only changed files) and it could be done in here.

To have better performance and tighter integration you would need to force a structure on big projects, so the app has more control and insight what has changed. This way reloading re-rendering stuff would happen more often on only parts.

For example, if you separate object design part of workflow from layout part of the workflow, then app would know if design files change or parameters change that object needs rebuilding (extrusions, booleans, etc). Because if yo only allow rotate, scale, translate, colorize in the layout part of the workflow, you can optimize those operations using threejs. I did not dig deep here, but generic performance optimization only goes so far.

I it is an unfortunate part of opensource that the community is in flux, and I think if you want more people on board, you also should think about a feature to simply share designs. Proudly sharing things we make is a good motivator for some(it is for me at least). One simple option without requiring you to have storage would be to integrate with github gists.

.....
in the end I would prefer we have this productivity enhancer sooner than later, because my motivation to contribute also depends on how much I use jscad.

z3dev:
> Please do not "dumb down" the web application. It is not only for beginners. I find the fact that there is already a hosted webapp extremely useful. I can share designs with other users using it, and I don't have to host my own.

The ability to share designs will not vanish. It's a favorite amoung JSCAD users.

> As for projects, a VSCODE plugin seems ideal to me. I think it makes the desktop app redundant.

A good suggestion, thanks.

z3dev:
> I it is an unfortunate part of opensource that the community is in flux, and I think if you want more people on board, you also should think about a feature to simply share designs. Proudly sharing things we make is a good motivator for some(it is for me at least). One simple option without requiring you to have storage would be to integrate with github gists.

Totally agreed. Sharing designs / parts would boost JSCAD immensely. There are a couple of sites that JSCAD needs better integration; jsfiddle and gists.

hrgdavor:
> > I it is an unfortunate part of opensource that the community is in flux, and I think if you want more people on board, you also should think about a feature to simply share designs. Proudly sharing things we make is a good motivator for some(it is for me at least). One simple option without requiring you to have storage would be to integrate with github gists.
> 
> Totally agreed. Sharing designs / parts would boost JSCAD immensely. There are a couple of sites that JSCAD needs better integration; jsfiddle and gists.

for jsfiddle/codepen a browserify build with all jscad modules(or most of them) would be nice I thnik, not 100% sure.

Also, although you have documentation examples, and designs can be shared, what is missing a some kind of listing for newcommers to discover the designs.

Even now I can save a design as gist and create a link that can be shared, it is not too difficult (it is not integrated or user firendly, but it is not too hard to do).

hrgdavor:
**Anyway, to get back to the topic, and this pull request.** I would like to make this livereload integration part of jscad web, and even the more advanced version (in a separate pull request). It is usefull and nothing is stopping us to use that code later for a special version that you were talking  about (a version more suited for big projects).

jscad web works fine for me right now(even more so with livereload), and I am making small things and some bigger ones 
 - puzzle with 15 pieces( with MAGNETS inside :D )
 - knife sharpening tool
 - trinkets

**regarding the more advanced jscad**
I had my own idea what I would like and what it could be , so I created an issue for brainstorm
https://github.com/jscad/OpenJSCAD.org/issues/620

I am interested what you consider a big project.
 - is it something you can share with me (privately at least) 
 - what are the pain points ? 
I am currently motivated to make some improvements to jscad, but unfortunatelly can't guarantee it will last (I am just like that, I get excited about some stuff, then I find other fun stuff to do...  ... well I do 3D print rather often, so there is actually chance II will be around jscad)

z3dev:
>     * what are the pain points ?

OK. If you're set to bring this into JSCAD then here are some points that need to be considered.
- this has to work on all browsers; chrome, firefox, edge, and safari. (Let me know if you need help to test)
- the auto-reload functionality needs to work, so...
  - the live-reload support should be attempted, and if fails then the fail over to the old method
  - the old method needs to work as the default
- auto-reload for single files OR projects

I think that should keep you busy for a little while. 👍

hrgdavor:
> * this has to work on all browsers; chrome, firefox, edge, and safari. (Let me know if you need help to test)

 - tested in Chrome, FF, Edge (I don't have Safari)
 - works on all, and is fast

Firefox has issue unrelated to livereload : reloading only works for projects, not single files (same as current jscad.xyz)

>   * the live-reload support should be attempted, and if fails then use the old method
>   * the old method needs to work as the default

That is how it works, livereload is attempted and reload logic is only changed if websocket conencts to livereload server

> * auto-reload for single files OR projects

Yes, it does not matter what type it is. Current method only affects how the timing for reload is chosen, and is not dependent on the project type (single file / folder)

> * hard-coding of the URL / PORT is not portable, so please add a small JSON config to the web/data/ directory

fixed
 
> I think that should keep you busy for a little while. 👍

DONE 👍

z3dev (changes requested):
So, I finally got around to testing this.

Seems to work fine across the browsers. 👍 

I have a few comments, and a few small code changes requested.

I didn't have to install livereload globally. It can be installed as part of a design project, and then just started by...
```
cd project-directory
npm install livereload
npx livereload
```

packages/web/src/sideEffects/localFs/index.js
```diff
@@ -3,9 +3,43 @@ const { callbackToObservable } = require('@jscad/core').observableUtils
...
+
+    webSocket.onopen = function (event) {
+      console.log('websocket open', event)
+      sendWs({ command: 'hello', protocols: ['http://livereload.com/protocols/official-6', 'http://livereload.com/protocols/official-7'], ver: '2.0.8', ext: 'Chrome', extver: '2.1.0' })
+      webSocketOpen = true
+    }
+
+    webSocket.onmessage = function (event) {
+      const data = JSON.parse(event.data)
+      if (data.command === 'hello') sendWs({ command: 'info', plugins: { less: { disable: false, version: '1.0' } }, url: document.location })
```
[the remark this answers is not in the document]
hrgdavor: These messages have not changed since I first saw livereload, cold be as much as 10 years. It is very unlikely to change.

packages/web/src/sideEffects/localFs/index.js
```diff
@@ -3,9 +3,43 @@ const { callbackToObservable } = require('@jscad/core').observableUtils
...
 const { changedFiles, flattenFiles } = require('./utils')
 
+const localFsOptions = require('../../../data/localFsOptions.json')
+
 const makeLocalFsSideEffect = async (params) => {
+  console.log('params', JSON.parse(JSON.stringify(params)))
   const commandResponses = callbackToObservable()
 
+  let webSocket = null
+  let webSocketOpen = false
```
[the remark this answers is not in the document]
hrgdavor: I switched to websocket=null after it is closed and use that for if(xxx)

packages/web/src/sideEffects/localFs/index.js
```diff
@@ -44,7 +81,13 @@ const makeLocalFsSideEffect = async (params) => {
 
         const { enabled } = options
         if (enabled) {
-          const walkAndCheck = () => {
+          if (!webSocketOpen) startWebSocket()
+
+          watcher = setInterval(() => {
+            const now = Date.now()
+            if (!rawData || (webSocketOpen && lastCheck) || (Date.now() - lastCheck) < watcherDelay) return
```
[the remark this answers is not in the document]
hrgdavor: we also have startMs few lines below, so I moved code a bit and used that

hrgdavor:
> I didn't have to install livereload globally. It can be installed as part of a design project, and then just started by...
> 
> ```
> cd project-directory
> npm install livereload
> npx livereload
> ```

Yes, livereload does not need to be global. It is just my preference :) (for some tools, but local install is always safer :) )

```
npm install livereload --save-dev
```
would add it as dev dependency ;)

packages/web/src/sideEffects/localFs/index.js
```diff
@@ -3,9 +3,41 @@ const { callbackToObservable } = require('@jscad/core').observableUtils
 
 const { changedFiles, flattenFiles } = require('./utils')
 
+const localFsOptions = require('../../../data/localFsOptions.json')
+
 const makeLocalFsSideEffect = async (params) => {
+  console.log('params', JSON.parse(JSON.stringify(params)))
```
z3dev, line 9: GAG. I didn't see this last review. Can you comment out this debug line? thanks

packages/web/src/ui/flow/design.js
```diff
@@ -423,7 +423,7 @@ const actions = ({ sources }) => {
   const setDesignContent$ = most.mergeArray(
     Object.values(sources).filter((x) => x !== undefined && 'source' in x)
   )
-    .delay(1000)// FIXME delay is a hack
+    //.delay(1000)// FIXME delay is a hack
```
z3dev, line 426: Please remove line 426. The comment FIXME is not necessary.

packages/web/src/sideEffects/localFs/index.js
hrgdavor: ok

packages/web/src/ui/flow/design.js
hrgdavor: ok

z3dev (approved):
Looks like a good starting point for livereload support.

hrgdavor:
It would be nice if someone could give me a sample BIG project with multiple includes. I could look into the advanced version of livereload support and measure performance.

z3dev:
> It would be nice if someone could give me a sample BIG project with multiple includes. I could look into the advanced version of livereload support and measure performance.

Best to ask the user community as there are some larger projects out there.

PR #735, opened 2020-11-17
Update orbitControls.js

hrgdavor:
### All Submissions:

a small fix that modifies pan so it is scaled based on current zoom (distance camera<->eye).

There is a comment near my change that somebody left there
```
  // TODO scale by the correct near/far value instead of 1000 ?
  // const planesDiff = camera.far - camera.near
```

but when I inspected near and far, it was the same regardless of zoom. Maybe this is related to some other aspect.

It is a simple improvement, please test if you find problems with the change.

### DIFF
--- a/packages/utils/regl-renderer/src/controls/orbitControls.js
+++ b/packages/utils/regl-renderer/src/controls/orbitControls.js
@@ -248,9 +248,8 @@ const pan = ({ controls, camera, speed = 1 }, delta) => {
   ]
   const unPanStart = unproject([], panStart, viewport, invProjView)
   const unPanEnd = unproject([], panEnd, viewport, invProjView)
-  // TODO scale by the correct near/far value instead of 1000 ?
-  // const planesDiff = camera.far - camera.near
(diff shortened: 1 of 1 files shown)


hrgdavor:
Also works better on mobile, and is more aligned with fingers than current jscad.xyz

packages/utils/regl-renderer/src/controls/orbitControls.js
```diff
@@ -250,7 +250,8 @@ const pan = ({ controls, camera, speed = 1 }, delta) => {
   const unPanEnd = unproject([], panEnd, viewport, invProjView)
   // TODO scale by the correct near/far value instead of 1000 ?
```
z3dev, line 251: I think this fix addresses this TODO, so let's remove the comments.

z3dev (changes requested):
Just a small change requested to remove comments.

I tried this on all browsers on MacOS and iOS. Works really nice. SUPER JOB!

z3dev (approved):
Looks nice!
Thanks for the late change.

Issue #562, opened 2020-04-22
V2 : Performance Ideas


z3dev:
Some ideas on boosting performance.

- [x] change math functions to require an 'out' parameter (like glmatrix)
- [x] poly3.transform() is calling mat4.isMirroring(), expensive on large 3D geometries
- [X] measureBounds() is called multiple times during boolean operations, cache the bounds for later use
- [BAD] create fast array concat (See [discussion](https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki))

## Expected Behavior
Blazing fast with same behaviour

## Actual Behavior
Slighter faster then V1

## Steps to Reproduce the Problem

  1. Run performance suite

## Specifications

  - Version: V2
  - Platform: ALL
  - Environment: ALL

z3dev:
Caching of bounding box and other measurements was implemented via

ahdinosaur:
hi @z3dev, how does someone run the performance suite?

i noticed these caches when trawling through the code, and i'm curious about the performance impact of using a WeakMap to store cached values rather than mutating the object, as i wonder if mutating the object would lead to JavaScript engine de-optimizations due to polymorphism. ([:information_source:](https://www.digitalocean.com/community/tutorials/js-v8-engine))

z3dev:
> hi @z3dev, how does someone run the performance suite?
> 
> i noticed these caches when trawling through the code, and i'm curious if there might be a performance improvement in using a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to store cached values rather than mutating the object, as i wonder if mutating the object would lead to JavaScript engine de-optimizations due to polymorphism. ([ℹ️](https://www.digitalocean.com/community/tutorials/js-v8-engine))

Interesting idea.

I’ve looked at WeakMap before and it seems kind of incomplete. I’m not sure why the standards introduced them. SIGH

I have a performance suite that runs whenever changes are merged into the V2 branch. It actually lives outside of this project, here.

hrgdavor:
Something is fishy in the renderer when geenrating regl code for 1000 cubes takes 4 seconds, but rotating the scene and zooming is snappy and works fine (so likely is at least 60fps).

my knowledge here is limited, but I am sensitive to performance of the tools I use.

It is not unusual for designs to have multiple instances of "same" geometry that is only copied and translated 
to multiple places. 

```js
[...]

PR #738, opened 2020-11-27
cache measurements using WeakMap

ahdinosaur:
here's a pull request to cache measurements using WeakMap references to object rather than mutating object. :cake: 

ref: https://github.com/jscad/OpenJSCAD.org/issues/562#issuecomment-734054731

i haven't run the whole performance suite for comparison yet, so far the numbers seem very slightly better but hard to tell if that's just random performance jitter. i do think this change makes the code cleaner and less surprising, but that's just my personal opinion so not sure if others agree.

### Checklist:

### DIFF
--- a/packages/modeling/src/measurements/measureArea.js
+++ b/packages/modeling/src/measurements/measureArea.js
@@ -5,6 +5,8 @@ const geom3 = require('../geometries/geom3')
 const path2 = require('../geometries/path2')
 const poly3 = require('../geometries/poly3')
 
+const cache = new WeakMap()
+
 /*
  * Measure the area of the given geometry.
  * NOTE: paths are infinitely narrow and do not have an area
@@ -23,12 +25,14 @@ const measureAreaOfPath2 = () => 0
  * @returns {Number} area of the geometry
  */
 const measureAreaOfGeom2 = (geometry) => {
-  if (geometry.area) return geometry.area
+  if (cache.has(geometry)) return cache.get(geometry)
 
   const sides = geom2.toSides(geometry)
-  const area = sides.reduce((area, side) => area + (side[0][0] * side[1][1] - side[0][1] * side[1][0]), 0)
-  geometry.area = area * 0.5
-  return geometry.area
+  const area = 0.5 * sides.reduce((area, side) => area + (side[0][0] * side[1][1] - side[0][1] * side[1][0]), 0)
+
+  cache.set(geometry, area)
+
+  return area
 }
 
 /*
@@ -38,11 +42,14 @@ const measureAreaOfGeom2 = (geometry) => {
  * @returns {Number} area of the geometry
  */
 const measureAreaOfGeom3 = (geometry) => {
-  if (geometry.area) return geometry.area
+  if (cache.has(geometry)) return cache.get(geometry)
 
   const polygons = geom3.toPolygons(geometry)
-  geometry.area = polygons.reduce((area, polygon) => area + poly3.measureArea(polygon), 0)
-  return geometry.area
+  const area = polygons.reduce((area, polygon) => area + poly3.measureArea(polygon), 0)
+
+  cache.set(geometry, area)
+
+  return area
 }
 
 /**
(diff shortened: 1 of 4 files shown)


z3dev:
@ahdinosaur thanks. I was expecting something like this.

The only down side of these changes is that the measurements will not be attributes of the geometry.
Not sure this really matters. 

I’ll run the performance suite with these changes as well. I’m not expecting any change as geometries are not created / destroyed often.

SimonClark:
I understand that weakmap doesn't stand in the way of the reference object's garbage collection. Is there a risk, though, that the cached measurements represent a sort-of memory leak.  When they are stored on the object, they get deleted with the object.  Here, they persist, correct?

hrgdavor:
> I understand that weakmap doesn't stand in the way of the reference object's garbage collection. Is there a risk, though, that the cached measurements represent a sort-of memory leak. When they are stored on the object, they get deleted with the object. Here, they persist, correct?

 - if u store measurement on the object that is deleted, the measurement is deleted
 - if u store measurement in a WeakMap using the object as the key, when object is deleted the key is removed from the WaekMap and associated value is deleted too.

The exact timing may not be the same but in the end memory is released in a timely manner, just maybe not in the same sweep depending on the garbage collector impl.

Finally, I am just not sure why would there be performance benefit ?
Possible performance benefit must be proven by test, as expectations are often not met when we try to optimize for performance.

ahdinosaur:
> The only down side of these changes is that the measurements will not be attributes of the geometry.
Not sure this really matters.

hi @z3dev, the only downside i can tell so far is that if you have different versions of `@jscad/modeling` that aren't de-duplicated, there will be different caches. this might lead to extra memory usage, but on the other hand is kinda more correct since every caches only contains compatible values.

> Here, they persist, correct?

hi @SimonClark, as @hrgdavor says the caches should be garbage collected, similar to before. i learned this idea by reading the internal code of memoization modules, e.g. [`mem`](https://github.com/sindresorhus/mem).

> Finally, I am just not sure why would there be performance benefit ?

hi @hrgdavor, my thinking as commented in https://github.com/jscad/OpenJSCAD.org/issues/562#issuecomment-734054731, is that there might be a performance benefit in keeping objects monomorphic. by storing the cache as an attribute of the object, there is now another shape of the object, which compounds if you cache in different orders, and potentially could de-optimize functions if the object is considered megamorphic. at least that was my hunch, i agree that any hypothetical benefit needs to be proven with a test, otherwise it's just a theory, and my theory might be wrong anyways. :smiley_cat:

SimonClark (commented):
I'm generally fine with this.  

Is it worth adding tests confirming that the caching is working?

packages/modeling/src/measurements/measureArea.js
```diff
@@ -23,12 +25,14 @@ const measureAreaOfPath2 = () => 0
  * @returns {Number} area of the geometry
  */
 const measureAreaOfGeom2 = (geometry) => {
-  if (geometry.area) return geometry.area
+  if (cache.has(geometry)) return cache.get(geometry)
```
z3dev, line 28: It may be faster just to call get(), and then test of a value is returned.

WeakMap.prototype.get(key)
- Returns the value associated to the key, or undefined if there is none.

z3dev (changes requested):
Just a little nit on performance. What do you think?

z3dev:
@ahdinosaur still there? i posted a comment as part of my review. what do you think?

z3dev:
@ahdinosaur can you try the suggestions that I posted?

z3dev:
After reading about Symbols, and various usage scenarios, I have additional questions.

First, should the bounds/volume/center/etc be attached to the geometry? I don’t think that performance has any difference either way. This question is about where the attribute should be attached, and seen by other applications / code. For example, the JSON Serializer will add the attributes to the JSON representation of the geometry. Good? Bad?

Second, should the cached value be hidden? Theses changes hide the value. If Symbols are used then the values will be hidden as well.

I’m kind of leaning towards leaving the current implementation as that leaves the values exposed, and usable. Maybe there’s a design or a library that wants to access these values or save those values.

z3dev:
A PR with the requested changes was submitted, so closing this PR.

Issue #696, opened 2020-09-24
Non-manifold vertices when extruding SVG or subtracting shapes


MathieuLoutre:
## Expected Behavior

Extruding an SVG and subtracting a cylinder from a cube should result in manifold models. See SCAD code provided.

## Actual Behavior

Extruding an SVG creates a model with non-manifold vertices. Same thing when subtracting a 2 cylinders from a cube. See JSCAD code provided. This leads to errors in PrusaSlicer.

## Steps to Reproduce the Problem

The STL produced here is shown as having non-manifold vertices in meshlab.

```js
const fs = require('fs').promises

const { geometries, primitives, booleans, extrusions, transforms } = require('../OpenJSCAD.org/packages/modeling')
const { svgDeSerializer, stlSerializer } = require('../OpenJSCAD.org/packages/io/io')

async function cast () {
	const height = 25.4
	const width = 12
	const depth = 22
	const scale = height/720
	const extrudeHeight = 1.32

	const shapes = svgDeSerializer.deserialize({
		output: 'geometry',
		target: 'path', 
		segments: 100
	}, '<svg PXPMM="1" xmlns="http://www.w3.org/2000/svg"><path d="M258 108C280 168 301 260 314 403L258 403C252 338 243 259 228 194C210 255 190 330 176 403C222 435 268 471 268 575C268 703 199 707 184 707C145 707 94 674 94 557C94 522 99 473 107 419C63 391 22 354 22 238C22 97 75-2 151-2C167-2 194 1 223 39C240 13 259-3 279-3C288-3 299-2 306 1L311 54C284 53 268 84 258 108ZM155 582C155 629 162 651 180 651C202 651 210 617 210 575C210 512 192 480 166 458C159 503 155 545 155 582ZM154 65C102 65 85 193 85 249C85 299 98 330 118 354C136 262 162 164 193 94C182 76 169 65 154 65Z" /></svg>')

	const poly = booleans.union(shapes.map((shape) => primitives.polygon({ points: shape.points })))
	const polyLetter = transforms.scale([scale, scale, 1], poly)
	let letterShape = extrusions.extrudeLinear({ height: extrudeHeight }, geometries.geom2.reverse(polyLetter))

	const body = booleans.subtract(
		booleans.union(
			primitives.cuboid({ 
				size: [width, height, depth],
				center: [0, 0, 0]
			}),
[...]

PR #743, opened 2020-12-15
V2 : adding new class of operations to modeling; modifiers

z3dev:
These changes add a new class of operations to the modeling API, which exposes two functions:
- snap : snap geometries to the precision (epsilon) of the geometry
- generalize : generalize geometries for exporting (including snap, simplify, triangulate, and repair steps)

After working on some changes for Issue #696, it became apparent that some functionality for modifying geometry can be used internally (in preparation for exports), as well as by users. So, there's a new class of operations called 'modifiers'. There are all kinds of functions that simplify, decimate, smooth, etc. geometry. These should be exposed for general use whenever possible.

Why operations called 'modifiers'?
- these are operations as new geometries are produced, just like extrude, union, etc.
- these 'modify' a geometry as a whole using a specific algorithm
- these can be applied to all geometries (in most cases)

### All Submissions:


z3dev:
@SimonClark can you take a quick look at these changes?

this is kind of radical, as the previous versions of JSCAD did not expose these kinds of functions. i think these should be exposed, as there should be no limit on what users can do with the library.

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
+const vec3 = require('../../maths/vec3')
+const line3 = require('../../maths/line3')
+
+const poly3 = require('../../geometries/poly3')
+
+// an edge is an array of two vertices
+const addEdge = (edges, edge, polygon) => {
```
SimonClark, line 7: Feel like this needs a docblock.  "Add an edge to a polygon, and the polygon to the edge's polygon references. If an equivalent edge is found in the array "edges", then use a reference to that edge instead.".

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
...
+  let ei = edges.findIndex((element) => {
+    if (element) {
+      if (vec3.equals(element[0], edge[0]) && vec3.equals(element[1], edge[1])) return true
```
SimonClark, line 10: These comparisons do not use the JSCAD-wide Epsilon to detect edges that are "close enough".  Should they?

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
...
+      if (vec3.equals(element[0], edge[1]) && vec3.equals(element[1], edge[0])) return true
+    }
+    return false
+  })
+
+  if (ei < 0) {
+    edge.polygons = [polygon]
```
SimonClark, line 17: ~~If the edge already has polygons, but is just not found in edges, it will wipe out the edges references.  This seems dangerous if we are exposing this function to unknown uses.~~

Nevermind, this one is not exposed.  Worth annotating as a private method somehow, in general, if a file has multiple exported methods?

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
...
+
+const removeEdge = (edges, edge) => {
+  let ei = edges.findIndex((element) => {
+    if (element) {
+      if (vec3.equals(element[0], edge[0]) && vec3.equals(element[1], edge[1])) return true
+      if (vec3.equals(element[0], edge[1]) && vec3.equals(element[1], edge[0])) return true
+    }
+    return false
+  })
+  if (ei >= 0) {
```
SimonClark, line 33: Should we settle on how we say "found in array". Perhaps it should always be explicitly ```ei == -1``` or ```ei != -1```

SimonClark (commented):
First round of comments, not finished looking but I wanted to publish these...

packages/modeling/src/maths/vec2/snap.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Snaps the coordinates of the given vector to the given epsilon.
+ *
+ * @param {vec2} out - the receiving vector
+ * @param {Number} epsilon - epsilon of precision, less than 0
+ * @param {vec2} vector - the vector to snap
+ * @returns {vec2} the resulting vector
+ * @alias module:modeling/maths/vec2.snap
+ */
+const snap = (output, epsilon, input) => {
```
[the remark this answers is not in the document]
z3dev: There is...

I'm not convinced that 'optional' parameters in the math classes are a good idea. And these may make JS runtime optimizations impossible.

I have a task for V2.1 to try to prove this. If there's no difference then we can keep the current signitures. If there is a performance improvement then changes are required.

The original [gl-matrix library](https://github.com/toji/gl-matrix) does not have optional parameters, it's always fuction(out, operand, in).

packages/modeling/src/maths/vec2/snap.js
```diff
@@ -0,0 +1,16 @@
...
+  output[0] = Math.round(input[0] / epsilon) * epsilon + 0
```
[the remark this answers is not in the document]
z3dev: Basically, calculations can produced either +0 or -0. This little addition makes everything +0.

There's a long long tirad about this on [StackExchange](https://stackoverflow.com/questions/7223359/are-0-and-0-the-same), which discusses all the issues of testing for +/-0 as well as calculations that return different results. It's not a good standard.

packages/modeling/src/maths/vec2/snap.js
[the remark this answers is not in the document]
SimonClark: fair, as long as you're on it :)

SimonClark (approved):
Kudos!  Lot's of tough math in this one.  I've not been able to take the time to understand double check the math itself, just looked for things that smelled off, and it looks pretty good.

Some minor comments about debugging cruft, and such.  I'll approve, but let you decide to merge or address as you see fit.

packages/modeling/src/operations/modifiers/edges.js
z3dev: Done

packages/modeling/src/operations/modifiers/edges.js
z3dev: Yeah. Edges are expected to be precise, which means 'snap' should be performed beforehand.

packages/modeling/src/operations/modifiers/edges.js
z3dev: Got it. Fixed.

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
...
+  addEdge(edges, edge, polygon)
+
+  for (let i = 0; i < (nv - 1); i++) {
+    edge = [vertices[i], vertices[i + 1]]
+    addEdge(edges, edge, polygon)
+  }
+}
+
+const removePolygons = (edges, oldedge) => {
+//console.log('removePolygons',oldedge)
```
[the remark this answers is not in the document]
z3dev: Done

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
...
+    if (polygons.length === 1) {
+      //console.log('open edge: ',edge[0],'<-->',edge[1])
+      edge.distance = vec3.squaredDistance(edge[0], edge[1])
+      openedges.push(edge)
+    }
+  })
+//console.log('**********OPEN*********')
+//console.log(openedges)
+//console.log('**********OPEN*********')
+console.log('open edges:',openedges.length)
```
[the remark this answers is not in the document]
z3dev: Done

packages/modeling/src/operations/modifiers/edges.js
```diff
@@ -0,0 +1,204 @@
...
+  })
+  return polygons
+}
+
+const polygonsToEdges = (polygons) => {
+  let edges = []
+  polygons.forEach((polygon) => {
+    addPolygon(edges, polygon)
+  })
+console.log('edges:',edges.length)
```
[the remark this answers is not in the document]
z3dev: Done

z3dev:
> Kudos! Lot's of tough math in this one. I've not been able to take the time to understand double check the math itself, just looked for things that smelled off, and it looks pretty good.
> 
> Some minor comments about debugging cruft, and such. I'll approve, but let you decide to merge or address as you see fit.

@SimonClark thanks for the review.

the fixing of t-junctions is not straight forward, and requires some tricky algorithms. this isn't perfect yet, and i may try porting the old algorithm as well. SIGH

i cleaned up the code and added comments. it is still kind of weak, so i will add more comments next time through.

SimonClark (approved):
Okay, I'm happy.

PR #751, opened 2020-12-30
V2 : fix extrude

### LINKED ISSUE #749: V2 strange artifact when extruding
## Expected Behavior

just extrusion without bottom square that surrounds it

## Actual Behavior
extrusion with bottom square that surrounds it

  1. save script in file
```

function main(){
	let points = [
		[4,0],[4,5],[6,5],[4,7],
		[-4,7],[-6,5],[-4,5],[-4,0]
	]
	let shape = jscad.geometries.geom2.fromPoints(points)
	return jscad.extrusions.extrudeLinear({height: 2}, shape)
}

module.exports = {main}
```
  2. load in jscad.xyz
  3.

image

## Specifications

  - Version: all
  - Platform: all
  - Environment: browser

z3dev:
These changes correct the calculation of planes for slices, where the slice is a complex shape. Additional tests have been added as well.

Fixes #749 

There's an additional fix to the core library which I found while working on the VUE components. This fix allows parameter values to be used even if the design does not define the parameter. In other words, designs should not be required to define parameters via getParameterDefinitions(). The checks should only be performed IF there's a definition.

### All Submissions:

### DIFF
--- a/packages/core/src/parameters/applyParameterDefinitions.js
+++ b/packages/core/src/parameters/applyParameterDefinitions.js
@@ -10,14 +10,10 @@
 const applyParameterDefinitions = (inputParameters, parameterDefinitions, throwOnNoDefinition = false) => {
   const values = Object.keys(inputParameters).reduce((paramValues, paramName) => {
     let value = inputParameters[paramName]
-    let definition = parameterDefinitions.filter((definition) => definition.name === paramName)
-    definition = definition.length > 0 ? definition[0] : undefined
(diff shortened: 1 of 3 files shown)


hrgdavor:
wee, my example from #749 works on this branch :)

z3dev:
> wee, my example from #749 works on this branch :)

A great example! Thanks, @hrgdavor

SimonClark (approved):
Looks good.

PR #769, opened 2021-02-16
Move web parameters to standalone package

danmarshall:
Hello!
In the sprit of the monorepo, I am proposing that the web UI parameters become their own package. This will allow folks to compose a lightweight UI with just the needed components. I'm thinking of the regl-renderer and these parameters controls.

There is no code change here, just moving files to a new folder in `/packages`.


packages/web-parameters/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@jscad/web-parameters",
+  "version": "2.0.0-alpha.15",
+  "description": "Parameter inputs for Web Application for JSCAD",
+  "repository": "https://github.com/jscad/OpenJSCAD.org",
+  "main": "src/index.js",
+  "scripts": {
+    "preversion": "npm run build && git add dist"
```
danmarshall, line 8: this might not be needed

z3dev:
@danmarshall hmm... so the idea is to lighten up the web UI? There’s a lot of stuff in there so I just wondering what else could be broken out.

z3dev:
@danmarshall sorry. I read this again.

Ok. So, you want to create a small web based UI which packages the core (for compiling), regl-renderer (for viewing), and parameters (for handling options).

I think there’s some good reasoning here but wonder if the parameters mean anything outside of the JSCAD web UI. I’m fairly fluent in VUE now and know that the current code would be kind of worthless. Not sure about REACT based UI components.

Thoughts?

danmarshall:
Yes if someone wanted this in native Vue or React these would be additional. But if you’re in vanilla js then this package is useful.

z3dev:
A few more questions... sorry.

The code in designParameters() relies on a most style call back. I understand that the HTML wrapper is valuable but don’t think the callback should be based on most. Wouldn’t this be more useful split apart?

Also, I came across another critical piece of code for web UI, called wallkFileTree(). And like you, wanted the code separated from the JSCAD web UI. I moved that to core/src/web/
Wouldn’t these pieces for parameters fit into the core package? There’s core/src/parameters/ as well.

danmarshall:
I'm planning on doing all the core file loading offline then webpacking it. So the core won't be a part of the viewer experience. I'm sot sure of the `most` callback, the only dependency is the `nanohtml`: https://github.com/jscad/OpenJSCAD.org/pull/769/files#diff-1b0c5b5ea9251678b787b60cc85e0c8d7c68e762a3368bace32ff3ca939436bcR32

z3dev:
Understood.

I have another tutorial planned for the Vue components that creates an input form from the design parameters. It could reuse this piece as well... hmm...

Can you wait for a week while I try to create the new tutorial?

danmarshall:
Of course, no hurry at all :)

danmarshall:
Hi @z3dev - I just created a new viewer tool called [jscad-now](https://github.com/danmarshall/jscad-now) which currently uses copy-pasted versions of these parameter files. I feel like there is more refactoring to do here, since I had to grab code from both `web` and `core` to get a complete working parameter feedback loop.

z3dev:
> Hi @z3dev - I just created a new viewer tool called [jscad-now](https://github.com/danmarshall/jscad-now) which currently uses copy-pasted versions of these parameter files. I feel like there is more refactoring to do here, since I had to grab code from both `web` and `core` to get a complete working parameter feedback loop.

@danmarshall thanks. I haven’t had time to look yet at the parameters tutorial. I lost my machine on Friday. 😔 but I’m getting back on board again.

I was expecting a few more pieces would be required. If you have suggestions then let me know.

Also, I saw the post on the new app  (?). It looks exciting as JSCAD needs a replacement for the desktop app. Maybe we could collaborate on something.

danmarshall:
@z3dev happy to collaborate! Give jscad-now a try. Also happy to pull it into this repo if you think it's up to that level.

danmarshall:
I've been thinking about this, and think these parameters should be in vanilla javascript. I don't think we can commit to a framework like Vue or React (or angular) because not everybody uses the same framework.

z3dev:
Just an update, I’m working on the new tutorial for VUE now, and have reached the point where I need to create HTML for the parameters. I’ll probably start down the same road, and make a copy of those functions.

I’ll post back when I have some more progress.

z3dev:
@danmarshall here's an update.

i created a new tutorial for the vue components that uses the definitions and the default values provided from the callback. it took a while to get that 'vue' mind set back but i happy to say that all works well.

the only component required was parameterControls. but vue is slightly different, so the conversion of definition to HTML element was the key part; createParamRowFromDefinition(). the whole HTML from createParamControls() would be unusable inside a vue component.

also, i was kind of surprised that the returned HTML element from createParamRowFromDefinition() did not contain the parameter 'name' as a HTML attribute. so, i made a small change to the code to add the 'name' as attribute 'paramenter-name'

z3dev:
Thoughts on a new package...

We have some cleanup to perform on V2 planned, which includes the breakout of 'most' based functions from the core package. I really doubt that the next application will use 'most', so those will probably be moved into the web package.

This will make the core package smaller, and allow further automated testing. It looks like a few of the functions from web need to be moved to core, and facilitate the development of some new applications, like jscad-now. I also hope to rework the code slight to make the 'vtree' pluggable, not compiled in to the core.

So, I'd like to resist this new package, and pursue a more compact 'core'.

Thoughts?

danmarshall:
I'm not quite sure what core does - it seems like a lot of file system stuff? I don't use core in jscad-now. Browserify is what pulls in files. So where will the parameters go? Into core, or a new vue component?

z3dev:
> I'm not quite sure what core does - it seems like a lot of file system stuff? I don't use core in jscad-now. Browserify is what pulls in files. So where will the parameters go? Into core, or a new vue component?

Hmm... If jscad-now isn't using the core then how are projects possible? Basically, rebuildGeometry* functions are the core, but require a whole lot more, like IO packages, etc.

z3dev:
> > I'm not quite sure what core does - it seems like a lot of file system stuff? I don't use core in jscad-now. Browserify is what pulls in files. So where will the parameters go? Into core, or a new vue component?
> 
> Hmm... If jscad-now isn't using the core then how are projects possible? Basically, rebuildGeometry* functions are the core, but require a whole lot more, like IO packages, etc.

I see. So, if the project stays within the limits of the packages included then all should work.

The only missing piece will be the extra support for require() of an external format, e.g. require('./my.stl')

danmarshall:
Hmm it seems even outside of jscad-now, in an ordinary node module, that require('./my.stl') would require that module to probably deserialize in order to use it? At this time I haven't imported stl files , I may have to investigate. Probably a browserify plugin would be needed.

Back to the parameters though, where will they live in the future? Probably in a ui package, which doesnt quite seem like core.

z3dev:
> Back to the parameters though, where will they live in the future? Probably in a ui package, which doesnt quite seem like core.

I’m not sure there’s any advantage, as Browserify will only assemble the pieces required. Wouldn’t a small ‘core’ suffice?

It would be reasonable to make a new package if it’s generally useable by various applications.

For example, the vue components build a library from core, and regl-renderer. I think Browserify does all the work... discarding unused functions... including only necessary functions.

danmarshall:
I just did a quick test. So it depends on how its imported.

```js
const core = require("@jscad/core");
console.log(core.parameters)
```
This bundles the whole core - 47k lines of code

```js
const parameters = require("@jscad/core/src/parameters");
console.log(parameters)
```
This bundles just the parameters, 227 lines of code.

The 2nd option is acceptable. Now the only missing part is the group of files from web. I'm unable to install @jscad/web because it references an unpublished version of most-gestures@0.4.2 - latest is 0.4.1

z3dev:
> The 2nd option is acceptable. Now the only missing part is the group of files from web. I'm unable to install @jscad/web because it references an unpublished version of most-gestures@0.4.2 - latest is 0.4.1

Yeah. I hope @kaosat-dev will accept the changes, and merge into most-gestures

danmarshall:
@z3dev - another reason for the standalone package, is that I've been contemplating that Maker.js use the same format/schema for parameters. For this to work it would make sense that the package be standalone. Other projects too may adopt the format, knowing they can do so in a lightweight way. Just a thought.

z3dev:
@danmarshall Got it!

Just some feedback...

I think those those two files from core should also go into this min-package. It seems those will be required if another website is created.

In addition, some of the code could be broken into smaller pieces, which will allow proper test suites to be added.

z3dev:
Final note... I’m releasing V2 JSCAD on April 1st. So, if this can wait just a little longer then these changes can be merged and released easily. There are still some steps to make the V2 JSCAD release process clean.

z3dev:
@danmarshall are you still interested in creating a small package for the parameter functionality?

if so then let's restart as the 'master' branch is now the working branch for changes. :)

z3dev:
@danmarshall lets restart on these changes. Lots of conflicts as well.

Issue #833, opened 2021-05-12
strange issues with rendering


z3dev:
## Expected Behavior

Load the following design (drag and drop), and the design is rendered.

## Actual Behavior

The screen goes 'white'.

If any rotation is performed then the design is rendered.

This same issue appears in Firefox, Chrome, and Safari.

## Steps to Reproduce the Problem

  1. As described.

## Specifications

  - Version: Latest
  - Platform: All
  - Environment: All browsers

z3dev:
```
const { geometries, primitives } = require('@jscad/modeling')

const { geom2, geom3, path2 } = geometries
const { arc, circle, sphere } = primitives

const main = (p) => {
  // create a HUGE path for testing
  let bigpath = arc({radius: 100, segments: 80000})
console.log('POINTS',bigpath.points.length)

  // return bigpath

  // create a HUGE 2D geometry for testing
  let biggeom2 = circle({radius: 100, segments: 80000})
console.log('SIDES',biggeom2.sides.length)

  // return biggeom2

  // create a HUGE 3D geometry for testing
  let biggeom3 = sphere({radius: 100, segments: 720})
console.log('POLYGONS',biggeom3.polygons.length)

  return biggeom3
}

module.exports = { main }
```

z3dev:
@hrgdavor can you take a look at this issue?

these are BIG designs that take some time to render.

hrgdavor:
I have tried it on https://openjscad.xyz/ and it has the issue that you described.

When I tried it locally, sphere was rendered on the screen just fine after processing was done.

online live version rendered sphere with 720 segments just fine in few seconds, but I had to reduce segments to 155 locally to get the sphere rendered in 3 seconds ... segments larger that that would just hang and not finish.

did you test this both locally and online ?

I also tried it in my new renderer: http://3d.hrg.hr/jscad/three/threejscad.html?uri=examples/render.huge.js
and I am glad it is faster for render initialization (part that takes model and converts to opengl) for this huge example :)

z3dev:
> did you test this both locally and online ?

Both. And across browsers.

it only happens with BIG designs.

hrgdavor:
I am using master branch locally, and could not reproduce it locally in chrome. 
Also I have issue where 720 segments is waay too many (it hangs) on the local version.

did you notice such differences ? I will ned to somehow reproduce it locally to be able to fix :)

z3dev:
[...]

PR #838, opened 2021-05-14
Another attempt at fixing rendering

z3dev:
After repeated attempts to fix issues with BIG designs( #833 ) , I discovered that the orbitControls returns an indication that the camera position was moved.

After correcting that logic, I came up with an alternative set of changes to the WEB UI viewer. The changes seem reasonable, and based on previous changes from @hrgdavor 

### All Submissions:

Tested on Firefox, Chrome, and Safari

### DIFF
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -121,7 +121,7 @@ const viewer = (state, i18n) => {
     const updateAndRender = (timestamp) => {
       doRotatePanZoom()
 
-      if (updateView || controls.autoRotate.enabled) {
+      if (updateView) {
         const updated = orbitControls.update({ controls, camera })
         controls = { ...controls, ...updated.controls }
         updateView = controls.changed // for elasticity in rotate / zoom
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -21,11 +21,7 @@ let rotateDelta = [0, 0]
 let panDelta = [0, 0]
 let zoomDelta = 0
 let zoomToFit = false
-let renderUntil = Number.MAX_VALUE
-
-// set a time to stop rendering with a small time buffer just in case.
-// rotation has some elasticity in movement so this way we let it finish for sure
-const moveRender = () => { renderUntil = Date.now() + 1000 }
+let updateView = true
 
 const grid = { // command to draw the grid
   visuals: {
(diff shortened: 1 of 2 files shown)


z3dev:
@hrgdavor this has been one challenging issue. can you try these changes?

packages/web/src/ui/views/viewer.js
```diff
@@ -91,49 +86,46 @@ const viewer = (state, i18n) => {
...
-      if (doRotatePanZoom() || controls.autoRotate.enabled) {
-        moveRender()
-      }
+      doRotatePanZoom()
 
-      if (renderUntil > Date.now()) {
+      if (updateView || controls.autoRotate.enabled) {
         const updated = orbitControls.update({ controls, camera })
         controls = { ...controls, ...updated.controls }
+        updateView = controls.changed // for elasticity in rotate / zoom
```
hrgdavor, line 127: that was the missing part :) I could not figure out

hrgdavor (approved):
looks good. nice refinement. works smoothly.

packages/web/src/ui/views/viewer.js
```diff
@@ -91,49 +86,46 @@ const viewer = (state, i18n) => {
...
         zoomToFit = false
-        changed = true
+        updateView = true
       }
-      return changed
     }
 
     // the heart of rendering, as themes, controls, etc change
     const updateAndRender = (timestamp) => {
-      if (doRotatePanZoom() || controls.autoRotate.enabled) {
```
hrgdavor, line 130: looks like it is no longer needed to check `|| controls.autoRotate.enabled`
as it is handled below by `updateView = controls.changed`

hrgdavor (changes requested):
remove || controls.autoRotate.enabled

packages/web/src/ui/views/viewer.js
z3dev: i was lucky to find it!

i knew there had to be something as the rotation did not function properly. but if i clicked 'auto zoom' a few times, the rotation continued. then, i understood that orbitcontrols.update() was adjusting the camera position. BINGO!!!

thanks a lot for the assistance!

packages/web/src/ui/views/viewer.js
z3dev: ahhhh! right!!!

z3dev: done

packages/web/src/ui/views/viewer.js
z3dev: P.S. I have another fix coming. You will like this one!

mister-jscad (approved):
Changes look fine. Hope this is the last one of these.

PR #839, opened 2021-05-15
fix(web): replaced JSON comparisons with ID comparisons in viewer

z3dev:
Small change... BIG performance boost, especially with large designs.

### All Submissions:

Tested with Firefox, Chrome, and Safari

### DIFF
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -145,8 +145,7 @@ const viewer = (state, i18n) => {
       const theme = state.themes.themeSettings.viewer
       const color = theme.rendering.meshColor
       const sameColor = prevColor === color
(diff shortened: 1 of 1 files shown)


z3dev:
@hrgdavor How about reviewing these changes?

This is something that I have been working on for a few weeks, and finally decided that the JSON comparisons were inappropriate (and HUGE CPU / MEMORY hogs). Simple comparisons if 'id' is enough.

hrgdavor (approved):
Yes, this will definitely help for large designs, and is better choice here.

This is something I will consider for the new web, as info on what was updated is crucial for the viewer to optimize rendering.

mister-jscad (approved):
Good changes. Hope this helps performance.

Issue #517, opened 2020-02-12
identifying solids 


joebk:
## Expected Behavior
I want to identify how many solids I have in my code, name them and be able to perform tasks on them. For the specific issue i have see the picture, i want yello part underneath the orange.
![jscadissue](https://user-images.githubusercontent.com/59221652/74316336-a736f600-4d79-11ea-8b9c-f572500da96d.jpg)
## Actual Behavior
When I use difference on two objects I end up with a number of solids however they are all referred to by the same name even though they do not touch. 

Hope you can help thanks. 

## Steps to Reproduce the Problem

  1. take a part e.g. lego brick
  2. put it inside a box
  3. use difference(lego brick,box) 

## Specifications

  - Version: 
  - Platform: windows
  - Environment: (browser, local server, node.js etc)web

gilboonet:
I wrote a function called scission3d() that splits such a solid and returns an array with each part.
`function sortNb		(E){ // returns E numerically sorted and deduplicated
	return E.sort(function(a, b) {return a-b}).filter(
	    function(item, pos, ary) {return !pos || item != ary[pos - 1]});
}

function compare3d	(v1, v2){ // returns true if V1 and V2 are at the same position
	return (v1.pos.x == v2.pos.x) && (v1.pos.y == v2.pos.y) && (v1.pos.z == v2.pos.z);
}

function scission3d	(geom){
  var i,Pl, j,i1,j1,ok,ti,tj,z,zz = [], P, RScission, til, tjl, tii1, zzl, zzdl;
// construit table de correspondance entre Polygones (P)
// build polygons lookup table
  P = geom.toPolygons();
  RScission = [];
  Pl = P.length;
  for (i = 0; i < Pl; i++){
	ti = P[i].vertices;
	z = [];
	for (j = 0; j < Pl; j++){
      tj = P[j].vertices;
	  ok = false;
	  for (i1 = 0; i1 < ti.length; i1++){
	    tii1 = ti[i1];
		for(j1 = 0; j1 < tj.length; j1++)
		  if (!ok)ok = compare3d(tii1, tj[j1]);
	  }
	  if (ok)z.push(parseInt(j));
	}
	z = sortNb(z);
	zz.push({e:0, d:z});
  }

// regroupe les correspondances des polygones se touchant
// boucle ne s'arrêtant que quand deux passages retournent le même nb de polygones
// merge lookup data from linked polygons as long as possible
  ok = false;
  nElOk = 0;
  do {
[...]

PR #846, opened 2021-05-26
feat(modeling): new scission() function added to booleans

z3dev:
These changes add a new function to the booleans, called scission(). This new function cuts/divideds/scissions the given geometry into component pieces.

This is useful to pull apart larger designs into pieces, especially STL mesh imports. As well as special times when the results of boolean operations need to be split into pieces. See #517 splitting of solid into pieces 

BIG THANKS to @giboneet for the initial implementation.

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/booleans/scission.js
+++ b/packages/modeling/src/operations/booleans/scission.js
@@ -1,0 +1,43 @@
+const flatten = require('../../utils/flatten')
+
+// const geom2 = require('../../geometries/geom2')
+const geom3 = require('../../geometries/geom3')
+
+// const scissionGeom2 = require('./scissionGeom2')
+const scissionGeom3 = require('./scissionGeom3')
+
+/**
+ * Scission (divide) the given geometry into the component pieces.
+ *
+ * @param {...Object} geometries - list of geometries
+ * @returns {Array} list of pieces from each geometry
+ * @alias module:modeling/booleans.scission
+ *
+ * @example
+ * let figure = require('./my.stl')
+ * let pieces = scission(figure)
+ *
+ * @example
+ * +-------+            +-------+
+ * |       |            |       |
+ * |   +---+            | A +---+
+ * |   |    +---+   =   |   |    +---+
+ * +---+    |   |       +---+    |   |
+ *      +---+   |            +---+   |
+ *      |       |            |    B  |
+ *      +-------+            +-------+
+ */
+const scission = (...objects) => {
+  objects = flatten(objects)
+  if (objects.length === 0) throw new Error('wrong number of arguments')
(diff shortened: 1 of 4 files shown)


packages/modeling/src/operations/booleans/scission.js
```diff
@@ -0,0 +1,43 @@
+const flatten = require('../../utils/flatten')
+
+// const geom2 = require('../../geometries/geom2')
```
mister-jscad, line 3: Hmm... is scission of 2D geometries possible?

packages/modeling/src/operations/booleans/scissionGeom3.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const polygonIndexesPerPoint = new Map()
+  const temp = vec3.create()
+  polygons.forEach((polygon, polygonIndex) => {
+    polygon.vertices.forEach((point) => {
+      insertMapping(polygonIndexesPerPoint, vec3.snap(temp, point, eps), polygonIndex)
+    })
+  })
+
+  const zz = polygons.map((polygon, polygonIndex) => {
```
mister-jscad, line 39: zz doesn't mean much. how about renaming to something like 'indexesPerPolygon'?

packages/modeling/src/operations/booleans/scissionGeom3.js
```diff
@@ -0,0 +1,93 @@
...
+    const indexes = []
+    polygon.vertices.forEach((point) => {
+      indexes.push(...findMapping(polygonIndexesPerPoint, vec3.snap(temp, point, eps)))
```
mister-jscad, line 42: Maybe concat() is a better choice as there are limits on the size of arrays using '...'

packages/modeling/src/operations/booleans/scissionGeom3.js
```diff
@@ -0,0 +1,93 @@
...
+        })
+      } while (merges > 0)
+      mapi.indexes = indexes
+    }
+  }
+
+  // construit le tableau des geometry à retourner
+  // build array of geometry to return
+  const newgeometries = []
+  const zzl = zz.length
```
mister-jscad, line 81: zzl could be reused above

mister-jscad (changes requested):
Suggested a few changes.

packages/modeling/src/operations/booleans/scission.js
z3dev: It's probably possible, but not straight forward.

Should it be implemented now? Not sure.

mister-jscad (approved):
The changes look fine.

PR #849, opened 2021-05-26
Update TypeScript type definitions

minodisk:
### All Submissions:

### Changes:

1. Copying JavaScript tests for TypeScript.
2. Run the TypeScript tests.
3. If you get a type error in TypeScript:
    - If the test code needs to specify a type, add the type to the test code.
    - If the type definitions are wrong or missing, fix the type definition.
    - If the type definitions are correct but the tests are wrong, fix the TS and JS tests.
4. Tighten type definitions if possible.
    - Treat the implementation as correct and do not modify it.


minodisk:
Finished following the changes in the master branch 😉

z3dev:
@minodisk Thanks for these changes.

I understand the changes, but what's the goal? What do these changes fix? What improvements do these changes provide?

z3dev:
Question... why commit copies of the tests?

the new ***.test.ts files are not required because there's a script to make copies.

minodisk:
Thanks for looking at the PR. I'll answer your questions below.

---

> but what's the goal? What do these changes fix? What improvements do these changes provide?

The first goal is to correct wrong type definitions. The second is to make the code easier to write and maintain by making the types more strict.

With the current ambiguous typing, it is impossible to use without type assertions. I feel that we are not taking advantage of the benefits of writing code using types.
The following code is actually run through the current type definition and the type definition modified by this PR.

```typescript
geom3.toPoints(translateX(10, cube()));
```

### With current typing:
#### Typing:
```typescript
export function translateX(offset: number, geometry: Geometry): Geometry
```
#### Result:
image
Type Error

### This PR typing:
#### Typing:
```typescript
export function translateX<T extends Geometry>(offset: number, geometry: T): T
```
#### Result:
image
Passed

This simple code does not pass with the current type definition.
This is because `translateX` returns an ambiguous type `type Geometry = Geom2 | Geom3 | Poly3 | Path2`. In this PR, the type definition is changed to use generics and return one of the types `Geom2`, `Geom3`, `Poly3`, or `Path2` received as argument `geometry`.
It's just one example, but I've made a number of these improvements in this PR.

---

> Question... why commit copies of the tests?

As for the TypeScript test files, I added them to ensure that the type changes are correct. Also, by copying the TypeScript test files and testing them through the TypeScript compiler, I was able to find the wrong type definitions *1 and also the wrong test codes *2. However, in maintaining the library, you have to fix two test codes for one implementation, which is indeed a pain.
So, I'll remove the TypeScript test files from this PR.

*1: Example of wrong type definitions
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-790774bc8a829f679fcfe90a447a4e9e02b1a1e9203f5c1512bf9e4cbb6e2f41L6
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-df29bfccd06c1f7258142ce635876eb3d5e070392cffe71d9f6455dd67008846L5
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-d1cd958b6e59bc4f24cab656a050394fa46e3e501b05929d09b2b280d35b3891L1
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-69b66937222bc02d6eeb2c9c5197f953c8798a4d79635ba722939c88268192c4L5
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-cec789fda5f6d2d30d1487542ed2d4ccfac7292c070c1ba19769d087301bf6a5L3
- etc...

*2: Example of wrong test codes
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-17de66e04b655aec60f49df19f9bdc723f15bfc7781b143f9a907901c2887731L14
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-5d6a5e6ca462ebd24efb56eb2a82ee0793f9c622b80a5f922ed6a9a8f6852aa6L24
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-6bde3acabb34ca3f73c406f768254dc683e472b00ef2df29dec07bab1c5bb327L9
- https://github.com/jscad/OpenJSCAD.org/pull/849/files#diff-f733383637c58445b68948b571486a1dd064fc7b9cb4a6c7242f4ef85023ca14L116

z3dev:
> > Question... why commit copies of the tests?
> 
> As for the TypeScript test files, I added them to ensure that the type changes are correct. Also, by copying the TypeScript test files and testing them through the TypeScript compiler, I was able to find the wrong type definitions *1 and also the wrong test codes *2. However, in maintaining the library, you have to fix two test codes for one implementation, which is indeed a pain.
> So, I'll remove the TypeScript test files from this PR.

Super. I'll do a full review once the extra test files have been removed.

I saw the changes to the declarations, as well as corrections to the tests. All good. 👍

minodisk:
@z3dev 

Thanks, I got rid of the TypeScript test files and the settings needed for testing. I also used *File changed* to make sure that only the intended files were included in this PR.

packages/modeling/src/operations/transforms/align.d.ts
```diff
@@ -1,14 +1,14 @@
...
-import Vec3 from '../../maths/vec3/type'
+import { Vec1, Vec2, Vec3 } from '../../maths/types'
 import RecursiveArray from '../../utils/recursiveArray'
 
 export default align
 
 export interface AlignOptions {
   modes?: Array<'center' | 'max' | 'min' | 'none'>
-  relativeTo?: Vec3
+  relativeTo?: Vec1 | Vec2 | Vec3
```
z3dev, line 9: Actually, this declaration of relativeTo is incorrect. It should just be Array. Using Vec2 or Vec3 implies that those arrays are numeric, which is not correct. This option (if provided) can have any combination of:
- [null] or [number]
- [null, null] or [null, number] or [number, number]
- [null, null, null] or [null, number, null] or [null, null, number] or [null, number, number] or [number, number, number]

Strange indeed...

z3dev: Which brings up my next comment... Vec1 is not necessary now.

packages/modeling/src/operations/transforms/mirror.d.ts
```diff
@@ -1,20 +1,20 @@
 import { Geometry } from '../../geometries/types'
-import Vec3 from '../../maths/vec3/type'
+import { Vec3 } from '../../maths/types'
 import RecursiveArray from '../../utils/recursiveArray'
 
 export interface MirrorOptions {
-  origin: Vec3
+  origin?: Vec3
   normal: Vec3
```
z3dev, line 7: normal is optional as well.

z3dev (changes requested):
Nice set of improvement, but there are some questions and requests.

z3dev:
I'm really hating TS now... it rips the heart out JavaScript. JavaScript is so cool because it has dynamic typing and call signatures, and here comes TS which locks down the whole language to strong types and strong signatures. Why doesn't MicroSoft just buck up and rename TS to VisualBasic?

packages/modeling/src/colors/colorize.d.ts
```diff
@@ -1,9 +1,12 @@
...
 
 import { RGB, RGBA } from './types'
 
 export default colorize
 
 declare function colorize<T extends Geometry>(color: RGB | RGBA, object: T): T
-declare function colorize(color: RGB | RGBA, ...objects: RecursiveArray<Geometry>): Array<Geometry>
+declare function colorize<T>(color: RGB | RGBA, object: T): T & Colored
+
+declare function colorize<T extends RecursiveArray<Geometry>>(color: RGB | RGBA, ...objects: T): T
```
z3dev, line 11: Why is this declaration necessary? I don't understand the need to have 'T extends RecursiveArray<Geometry>'

hrgdavor:
> I'm really hating TS now... it rips the heart out JavaScript. JavaScript is so cool because it has dynamic typing.. 

I can see benefits for some, but I use JavaScript for scripting and all of it's nice features. These typescript definitions may be nice for some of us here, but personally I prefer JS for my jscad scripts.

z3dev:
> > I'm really hating TS now... it rips the heart out JavaScript. JavaScript is so cool because it has dynamic typing..
> 
> I can see benefits for some, but I use JavaScript for scripting and all of it's nice features. These typescript definitions may be nice for some of us here, but personally I prefer JS for my jscad scripts.

Agreed. These declarations are assisting others, so let’s continue.

packages/modeling/src/geometries/types.d.ts
```diff
@@ -2,11 +2,16 @@ import Geom2 from './geom2/type'
 import Geom3 from './geom3/type'
 import Poly3 from './poly3/type'
 import Path2 from './path2/type'
+import { RGB, RGBA } from '../colors'
 
 // NOTE (@ahdinosaur): Poly2 is not included as not well-supported.
 // see https://github.com/jscad/OpenJSCAD.org/pull/726#issuecomment-724575265
 export type Geometry = Geom2 | Geom3 | Poly3 | Path2
 
+export type Colored = {
```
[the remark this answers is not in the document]
minodisk: TypeScript requires all properties to be defined, but those that are not required can be defined optionally.
Functions that use only certain properties can be typed in a duck-typed way using generics.
It might even be possible to create a base type for Geometry with the properties `color?` and `name?`.

packages/modeling/src/maths/mat4/fromVectorRotation.test.js
```diff
@@ -113,15 +113,15 @@ test('mat4: fromVectorRotation() called with out parameter should return a mat4
   ret = fromVectorRotation(out, [0.5, 0.5, 0.5], [-0.5, 0.5, 0.5])
   t.is(ret, out)
   t.true(compareVectors(out, [0.3333333333333334, 0.6666666666666669, 0.6666666666666669, 0, -0.6666666666666669, 0.666666666666667, -0.3333333333333335, 0, -0.6666666666666669, -0.3333333333333335, 0.666666666666667, 0, 0, 0, 0, 1]))
-  t.true(compareVectors([-0.5, 0.5, 0.5], vec3.transform(vec3.create(), ret, [0.5, 0.5, 0.5])))
+  t.true(compareVectors([-0.5, 0.5, 0.5], vec3.transform(vec3.create(), [0.5, 0.5, 0.5], ret)))
```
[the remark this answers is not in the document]
minodisk: I thought the same thing about this test. The person who wrote this test may need to rewrite it, as it may not be the essential test code.

packages/modeling/src/operations/transforms/align.d.ts
minodisk: Thanks. So it's not Vec1, Vec2, or Vec3, but some other type. I'll look at the implementation and try to define the type.

packages/modeling/src/operations/transforms/mirror.d.ts
minodisk: Thanks. I'll make it optional.

packages/modeling/src/colors/colorize.d.ts
minodisk: This definition is not necessary and is wrong, so I'll fix it. And the return value is a flat array.

minodisk:
> > > I'm really hating TS now... it rips the heart out JavaScript. JavaScript is so cool because it has dynamic typing..
> >
> > I can see benefits for some, but I use JavaScript for scripting and all of it's nice features. These typescript definitions may be nice for some of us here, but personally I prefer JS for my jscad scripts.
>
> Agreed. These declarations are assisting others, so let’s continue.

Knowing this approach to TypeScript, please listen to this as just a suggestion.
In the TypeScript ecosystem, there is a repository called [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) that collects only type definitions. I can manage jscad's typedefs there. That way, jscad members don't have to spend time reviewing *"annoying"* typedefs, and can focus on implementation 😉 
I think it's a better way to provide types for those who need them, and not affect those who don't.
If you want to remove the typedefs from this repository, I'll do the migration, what do you think?

packages/modeling/src/geometries/types.d.ts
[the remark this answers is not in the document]
z3dev: Ok. Let’s create a (cough) base type.

z3dev:
> > > > I'm really hating TS now... it rips the heart out JavaScript. JavaScript is so cool because it has dynamic typing..
> > > 
> > > 
> > > I can see benefits for some, but I use JavaScript for scripting and all of it's nice features. These typescript definitions may be nice for some of us here, but personally I prefer JS for my jscad scripts.
> > 
> > 
> > Agreed. These declarations are assisting others, so let’s continue.
> 
> Knowing this approach to TypeScript, please listen to this as just a suggestion.
> In the TypeScript ecosystem, there is a repository called [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) that collects only type definitions. I can manage jscad's typedefs there. That way, jscad members don't have to spend time reviewing _"annoying"_ typedefs, and can focus on implementation 😉
> I think it's a better way to provide types for those who need them, and not affect those who don't.
> If you want to remove the typedefs from this repository, I'll do the migration, what do you think?

That’s pretty cool. But let’s complete this pull request. Your changes are good, and reviews are improving the definitions as well.

Once completed, you can move the declarations over to the common repository.

z3dev (changes requested):

z3dev (dismissed):
I'm approving of these changes because these improve the definitions, and fix errors in test suites.

However, there's still a long long way to go to make these TS declarations useful.

I spent just a little time looking at using 'tsd' further to test the declarations. It was failing everywhere; types were wrong, declarations were wrong, etc. There's only one test file as well. As with JS, TS must follow the example.
- Tests for everything
- All tests must pass

minodisk:
@z3dev 

I think I've fixed the parts you pointed out and all the RecursiveArray issues.

z3dev (approved):
@minodisk Thanks for these very nice changes.

ahdinosaur:
hey @minodisk, thanks for fixing my types! :+1: :purple_heart:

PR #850, opened 2021-05-29
feat(modeling): add new measurements for BoundingSphere, Center, CenterOfMass, and Dimensions

z3dev:
These changes add new measurements for all types of geometries. The new measurements include:
- Bounding Sphere
- Center (based on Bounding Box)
- Center of Mass
- Dimensions (based on Bounding Box)

### All Submissions:

### DIFF
--- a/packages/modeling/src/measurements/measureBoundingSphere.js
+++ b/packages/modeling/src/measurements/measureBoundingSphere.js
@@ -1,0 +1,147 @@
+const flatten = require('../utils/flatten')
+
+const vec2 = require('../maths/vec2')
+const vec3 = require('../maths/vec3')
+
+const geom2 = require('../geometries/geom2')
+const geom3 = require('../geometries/geom3')
+const path2 = require('../geometries/path2')
+const poly3 = require('../geometries/poly3')
+
+const cacheOfBoundingSpheres = new WeakMap()
+
+/*
+ * Measure the bounding sphere of the given (path2) geometry.
+ * @return {[[x, y, z], radius]} the bounding sphere for the geometry
+ */
+const measureBoundingSphereOfPath2 = (geometry) => {
+  let boundingSphere = cacheOfBoundingSpheres.get(geometry)
+  if (boundingSphere !== undefined) return boundingSphere
+
+  const centroid = vec3.create()
+  let radius = 0
+
+  const points = path2.toPoints(geometry)
+
+  if (points.length > 0) {
+    // calculate the centriod of the geometry
+    let numPoints = 0
+    const temp = vec3.create()
+    points.forEach((point) => {
+      vec3.add(centroid, centroid, vec3.fromVec2(temp, point, 0))
+      numPoints++
+    })
+    vec3.scale(centroid, centroid, 1 / numPoints)
+
+    // find the farthest point from the centroid
+    points.forEach((point) => {
+      radius = Math.max(radius, vec2.squaredDistance(centroid, point))
+    })
+    radius = Math.sqrt(radius)
+  }
+
+  boundingSphere = [centroid, radius]
+  cacheOfBoundingSpheres.set(geometry, boundingSphere)
+
+  return boundingSphere
+}
+
+/*
+ * Measure the bounding sphere of the given (geom2) geometry.
+ * @return {[[x, y, z], radius]} the bounding sphere for the geometry
+ */
+const measureBoundingSphereOfGeom2 = (geometry) => {
+  let boundingSphere = cacheOfBoundingSpheres.get(geometry)
+  if (boundingSphere !== undefined) return boundingSphere
+
+  const centroid = vec3.create()
+  let radius = 0
+
+  const sides = geom2.toSides(geometry)
+
+  if (sides.length > 0) {
+    // calculate the centriod of the geometry
+    let numPoints = 0
+    const temp = vec3.create()
+    sides.forEach((side) => {
+      vec3.add(centroid, centroid, vec3.fromVec2(temp, side[0], 0))
+      numPoints++
+    })
+    vec3.scale(centroid, centroid, 1 / numPoints)
+
+    // find the farthest point from the centroid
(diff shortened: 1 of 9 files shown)


z3dev:
@SimonClark let me know what you think about these new measurements. i think that aggregate versions could be useful as well.

packages/modeling/src/measurements/measureCenterOfMass.js
```diff
@@ -0,0 +1,90 @@
...
+const geom3 = require('../geometries/geom3')
+const path2 = require('../geometries/path2')
+
+const measureBoundingBox = require('./measureBoundingBox')
+
+const cacheOfCenterOfMass = new WeakMap()
+
+/*
+ * Measure the center of mass for the given geometry.
+ * The given geometry is infinitely thin, so there's no real mass, just a center.
```
SimonClark, line 15: I'm a little concerned that this becomes misleading.  If there's no mass, as asserts, there's no center of mass, and it should return null, or throw an error.

However, if you consider the center of mass to be what the center would be for, say, a piece of paper, then this would return an incorrect result.  

Either way, this is surprising.

packages/modeling/src/measurements/measureCenterOfMass.js
```diff
@@ -0,0 +1,90 @@
...
+  centerOfMass[0] = (bounds[0][0] + ((bounds[1][0] - bounds[0][0]) / 2))
+  centerOfMass[1] = (bounds[0][1] + ((bounds[1][1] - bounds[0][1]) / 2))
+  centerOfMass[2] = (bounds[0][2] + ((bounds[1][2] - bounds[0][2]) / 2))
+
+  cacheOfCenterOfMass.set(geometry, centerOfMass)
+  return centerOfMass
+}
+
+/*
+ * Measure the center of mass for the given geometry.
```
SimonClark, line 33: Code looks good. I think the math checks out, but I'm less certain, there. :). I'll see if I can throw some tests at it, too.

packages/modeling/src/measurements/measureDimensions.js
```diff
@@ -0,0 +1,28 @@
+const flatten = require('../utils/flatten')
+
+const measureBoundingBox = require('./measureBoundingBox')
+
+/**
+ * Measure the dimensions of the given geometries.
```
SimonClark, line 6: Nice!  I keep doing this manually :p

SimonClark (commented):
One comment on the geom2 centerOfGravity result that I'm uncomfortable with.

packages/modeling/src/measurements/measureCenterOfMass.js
z3dev: Yeah. There's probably a center of mass calculation for 2D solids as well. I'll research. This might be mid-point (average of all points).

Having said that... I think that for a path, there is no 'mass' as there is truly no mass in any direction. The path is infinitely thin, no volume, no width, no height, no depth.

I guess the question is.. what's the default center of mass for non-geometries?

FYI, measure bonding box returns [0,0,0] for non-geometries.

z3dev: Found the formula for the 2D center of mass from Paul Bourke. It's a centroid based on the area of the 2D polygon.

[centroid.pdf](https://github.com/jscad/OpenJSCAD.org/files/6680165/centroid.pdf)

packages/modeling/src/measurements/measureCenterOfMass.js
z3dev: See the centroid.pdf for the math. ;)

packages/modeling/src/measurements/measureCenterOfMass.js
```diff
@@ -0,0 +1,90 @@
...
+
+const vec3 = require('../maths/vec3')
+
+const geom2 = require('../geometries/geom2')
+const geom3 = require('../geometries/geom3')
+const path2 = require('../geometries/path2')
+
+const measureBoundingBox = require('./measureBoundingBox')
+
+const cacheOfCenterOfMass = new WeakMap()
```
z3dev, line 11: This secondary cache is unnecessary.

z3dev: The cache is required for storing center of mass for 3D geometries.

packages/modeling/src/measurements/measureCenterOfMass.js
z3dev: @SimonClark please review again.

center of mass calculations for geom2 have changed. this should be a far better implementation.

packages/modeling/src/measurements/measureBoundingSphere.js
```diff
@@ -0,0 +1,146 @@
...
+const path2 = require('../geometries/path2')
+const poly3 = require('../geometries/poly3')
+
+const cacheOfBoundingSpheres = new WeakMap()
+
+/*
+ * Measure the bounding sphere of the given (path2) geometry.
+ * @return {[[x, y, z], radius]} the bounding sphere for the geometry
+ */
+const measureBoundingSphereOfPath2 = (geometry) => {
```
SimonClark, line 17: ok, so I am not convinced that this is doing what it implies.  I agree that it calculates a bounding sphere for the geometry, but I don't believe that it calculates the smallest bounding sphere for the geometry.

imagine a cloud of points: 
```
    x          x
      x   x      
 x                                                      x
     x        x
```
The centroid would be heavily weighted to the left, the radius would be big enough to encompass the right-most point, and so the bounding sphere would include a bunch of empty space on the left.

z3dev: It’s definitely NOT the most efficient sphere. That calculation takes huge amounts of parallel computations based on simulation. This is just a “rough” estimate, but still valid.

hrgdavor: I tend to trust choices from threejs :)

PR #853, opened 2021-06-02
proposal remove: deserializeSolids and serializeSolids

hrgdavor:
### All Submissions:

I just tried to remove JSON.stringify and JSON.parse from communication with worker.
It seems unnecessary, and all test passed. Also some sample models worked just fine.

Use this version of the PR is to quick change to test the proposal. And if agreed then next step would be to remove deserializeSolids and serializeSolids modules completely.


z3dev:
@hrgdavor Interesting. I'm sure objects are transferred via JSON automatically, i.e. sending messages with a general payload just does JSON. Can you explore this further?

Also, is there anyway to compare the performance?

z3dev:
FYI, V1 was converting the geometries to compact binary form, which is basically a typed array. This was for performance reasons only.

The down side of using typed arrays is that normal attributes like ‘id’ or ‘name’ were excluded. Users should be able to add attributes to geometries for use in other libraries, etc.

hrgdavor:
stringify and parse are not used behind the scenes, but some optimized  "Structured Clone algorithm" that should be a bit faster, and also handles well circular references unlike JSON.

Someone else did the digging, and it seems that there is no benefit in doing JSON.stringify + parse. Performance difference is hard to measure.
https://surma.dev/things/is-postmessage-slow/
conclusion is that stringify is not worth it

hrgdavor:
> FYI, V1 was converting the geometries to compact binary form, which is basically a typed array. This was for performance reasons only.
> 
> The down side of using typed arrays is that normal attributes like ‘color’ or ‘id’ or ‘name’ were excluded. Users should be able to add attributes to geometries for use in other libraries, etc.

This could have been avoided by only converting numeric array data to binary form, and shallow copy all other properties of each geometry. 
However this approach would, I think, only bring down performance (but one can never assume about performance :) but test instead ).

If on the the other hand geometries would use ArrayBuffers internally (assumming previously mentioned problems are solved),
postMessage has a second argument where you put references to all buffers. In that case those bufers are transfered superfast, ownership changes to the receiver of the postMessage, and code that sends data loses access to them unless later returned back.

I think tat even better approach would be to run both model script execution and rendering in the same worker, and that way there is no need for postMessage and any conversion, they could just share data. I am working on this for the new web viewer.

z3dev:
> I think tat even better approach would be to run both model script execution and rendering in the same worker, and that way there is no need for postMessage and any conversion, they could just share data. I am working on this for the new web viewer.

😎 but be aware that there will be others that want to see just the results for the execution. There’s already an option for style of serialization. 😃

hrgdavor:
> > I think tat even better approach would be to run both model script execution and rendering in the same worker, and that way there is no need for postMessage and any conversion, they could just share data. I am working on this for the new web viewer.
> 
> 😎 but be aware that there will be others that want to see just the results for the execution. There’s already an option for style of serialization. 😃

Yes, execution and render code are separate and would remain separated. So, that use case would not be affected.

Viewer can use postMessage or pass geometries directly to renderer depending if they are in the same thread or not. 

The option for style of serialization is ok, but the version that uses JSON.stringify is likely unnecessary.

serializeSolids would then be ok to keep as it reformats geometries, but deserializeSolids would be obsolete.

z3dev:
@hrgdavor I tested these changes, and did not find any negative effects. I tried every design that I have, plus all previous test cases for previous issues. No problem.

So, please make the final changes. These too functions are a tight pair, so let's remove the JSON and just pass the objects straight through.

P.S. I read the article on post message. Very interesting and in depth. The best solution is just pass the objects, and let the java engines do the serialization. Fun reading!

hrgdavor:
Yeah, the devs working on the javascript engines are doing some great stuff. And the article was fun to read indeed. 
Although I am here because of my 3d printing hobby, I have learned a few nice details about javascript that I will be able to use at work.

packages/web/src/ui/flow/design.js
```diff
@@ -478,21 +477,6 @@ const actions = ({ sources }) => {
...
-        try {
-          const start = new Date().getTime()
-
-          const solids = deserializeSolids(event.data.solids)
-
-          const end = new Date().getTime()
-          console.warn(`elapse for solid generation: ${end - start}`)
-
-          const { lookupCounts, lookup } = event.data
-          return { solids, lookup, lookupCounts }
```
z3dev, line 491: The return from this map((event) is different now. Does this work?

packages/core/src/code-evaluation/serializeSolids.js
```diff
@@ -14,7 +14,7 @@ const serializeSolids = (solids) => {
     if (geom2.isA(object)) geom2.toSides(object)
     if (geom3.isA(object)) geom3.toPolygons(object)
     if (path2.isA(object)) path2.toPoints(object)
-    return JSON.stringify(object)
+    return object
```
z3dev, line 17: Let's add another NOTE... something like...

NOTE: Serialization of objects is the fastest due to optimizations in the JavaScript engines.

z3dev (commented):
Some questions..

z3dev:
@hrgdavor Please see the comments. I like the changes, but let's document WHY this was changed.

packages/web/src/ui/flow/design.js
hrgdavor: you are right, it stopped working when this was completely removed. I have made a fix

packages/core/src/code-evaluation/serializeSolids.js
hrgdavor: I have adeed a note. Is this ok now  for merging ? :)

z3dev (approved):
OK. These changes look good.

We don't have automated testing for the WEB UI, so please take extra time to manually test related functionality.

PR #854, opened 2021-06-02
Fix render onresize

hrgdavor:
### All Submissions:

render needs to be called after window resize

### DIFF
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -58,6 +58,8 @@ const viewer = (state, i18n) => {
     render = prepareRender(viewerOptions)
     const gestures = pointerGestures(el)
 
+    // if(typeof window != 'undefined') window.addEventListener('resize', evt=>updateView = true)
+
     // rotate & pan
     gestures.drags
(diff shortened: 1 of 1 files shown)


packages/web/src/ui/views/viewer.js
```diff
@@ -58,6 +58,8 @@ const viewer = (state, i18n) => {
     render = prepareRender(viewerOptions)
     const gestures = pointerGestures(el)
 
+    if(typeof window != 'undefined') window.addEventListener('resize', evt=>updateView = true)
```
z3dev, line 61: Please adjust to follow syntax rules for spacing, functions, etc.

z3dev (changes requested):
Small change requested otherwise this is fine.

packages/web/src/ui/views/viewer.js
hrgdavor: ok, forgot about that :) .. I have applied standardx to the file, also removed `if(typeof window != 'undefined')` as window global is used in few more places.

z3dev (approved):
Super. Thanks.

Issue #856, opened 2021-06-06
reimagine transforms to be more open to more rich geometry behaviors


hrgdavor:
## Expected Behavior
simplify transform, and also let it keep all properties except `transforms`

## Actual Behavior

translate function has this piece of code
```js
  const results = objects.map((object) => {
    if (path2.isA(object)) return path2.transform(matrix, object)
    if (geom2.isA(object)) return geom2.transform(matrix, object)
    if (geom3.isA(object)) return geom3.transform(matrix, object)
    return object
  })
```

that then calls transform on each type, but implementation there is all the same
 - call create that copies one or two properties and adds empty transforms
 - combines the the transforms 

this is bunch of redundant code that could be avoided if just following a convention

## proposal
there should be a general transform function that just copies all properties of a geometry,
creates new matrix and applies the transforms

```js
  const results = objects.map((geometry) => {
    const newGeometry = {...geometry}
    newGeometry.transforms = mat4.multiply(mat4.create(), matrix, geometry.transforms)
    return newGeometry
  })
```

as a bonus, this would keep the color of the geom

**whoever imagined the current approach in transforms, should look back, and reconsider**

## background
The reason this came into focus is that while considering making connectors, a specific way to do them came to mind
 - geometry could keep reference to original parameters that created it 
 - method would be added that could generate different connectors as needed based on the for example cylinder radius, and height and aslo current transforms

if transforms would work as proposed, this method would be copied too along with original parameters.
The method would work properly after transformation if it considered transforms while generating connector

z3dev:
Simplification of code is always best for maintenance and innovation. This can probably happen after something like this (from the wiki)

Geometries
- add both clone (shallow) and clone (deep) functions
- cloneShallow() creates a new object using Object.assign()
- cloneDeep() creates a new object by making new allocations of everything
NOTE: applyTransforms() must reset transforms with a NEW mat4 allocation

hrgdavor:
Personally, I think cloneDeep should not be needed as geometries are immutable and should be immutable.
cloneDeep should be avoided aggressively.

Googling Object.assign performance, it looks like it is either faster or same ... I would image it should be same as compiler should know it is the same code logically. https://jsben.ch/7oaEs

z3dev:
[...]

PR #859, opened 2021-06-07
clone utility

hrgdavor:
related to #856 

I have added a check if the cloned object has it's own clone function, but I think it is not necessary yet, and can easily be added later on.

I have applied standardx, but no idea what to test if even necessary for such basic code.


z3dev:
This is a great start but there are some issues with this approach.

First, this doesn't guarantee that the resulting object (geometry) is **immutable**, which is required throughout the design of the library. For example...
```
let a = cube()
let b = clone(a) // b now has the guts of a
let c = geom3.toPolygons(b) // or geom3.applyTransforms(b)
```
After the toPolygons, b has new polygons, and b reset transforms to mat4.identity. However, a has the old polygons still, BUT the transforms have been changed as well to mat4.identity(). There's a small change required in the applyTransform function of the geometries.

NOTE: geom3.toPolygons(), geom2.toSides(), path2.toPoints() is being called often throughout the libraries.

Second, a utility function is fine, and it might be used by the geometries. However, adding the utility function to the utilities API means that users can call this without worry. That's not the case currently. So, this new functionality might be better as part of the geometries, such as cloneShallow().

My suggestion is that geom3.clone(), geom2.clone(), and path2.clone() all produce 'deep' copies of the guts. And a new function is added to all geometries to 'extend' or 'mutate' or 'tweek' attributes. For example...
```
let a = cube()
let b = colorize([1, 0, 1], a) // and this calls geom3.tweek({ color: [1, 0, 1] }, a)
```
The colorize function is a good example, where the guts of the geometry are retained, and a new attribute 'color' is added.

hrgdavor:
I will look into these issues in more detail. 

Here are my initial thoughts:

A better approach is to fix the offending methods so that shallow clone remains safe. 
I am willing to assist in that direction.

Using deep-cloning just to play defensive/safe will inflict big performance hits very easily.

Clone utility is meant for advanced users that must be aware to not break immutability.

On the other hand, from what I understand by looking at the code... 
This clone utility would not be any less safe than translate, as even after translate(geom3) one could break the geometry(by messing with polygons data) that would affect other copies.

This inital PR is part of a push for geometries to remain usable (with extended functionalities) at least after doing basic operations like transform and colorize.

It is a line of thinking that came out of the proposal for new connectors.

hrgdavor:
I have created a pull request to address the mutability of the applyTransforms #866

I hope this direction is acceptable.

hrgdavor:
Closed in favor of fresh attempet to bring immutability into the core

PR #860, opened 2021-06-07
mat4.inverse

hrgdavor:

code is MIT licensed from https://github.com/jlmakes/rematrix/blob/master/src/index.js

not sure how to properly attribute it to the author


z3dev (changes requested):
I like the implementation as throwing an error is far better then returning null (gl-matrix).

But the signature should be the same as gl-matrix, as this is our base for math.

https://github.com/toji/gl-matrix/blob/master/src/mat4.js

Also, a very simple test case is required. :)

hrgdavor:
wait, can I just copy code from gl-matrix ?
 - use name: `invert` instead of `inverse`

why is there jscad mat4 and gl-matrix ? I am confused now 😄 😄

z3dev:
> wait, can I just copy code from gl-matrix ?

Sure. We already give credit to gl-matrix.

> * use name: `invert` instead of `inverse`
> 
> why is there jscad mat4 and gl-matrix ? I am confused now 😄 😄

long story, but basically gl-matrix moves and changes based on the needs of REGL (WebGL library). In addition, all the structures are typed buffers, which has been proven slow and low precision.

JSCAD maths is a small subset as well, and slightly different. There are some nice additions in mat4.

hrgdavor:
ok, I will use version from gl-matrix and check if code still works :) and add a small test :)

hrgdavor:
@z3dev please check :) added tests and used code from gl-matrix

packages/modeling/src/maths/mat4/invert.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * Creates a inverse copy of the given matrix.
+ * @author Julian Lloyd
+ * code from https://github.com/jlmakes/rematrix/blob/master/src/index.js
+ *
+ * @param {mat4} m - matrix to invert
+ * @returns {mat4} out
+ * @alias module:modeling/maths/mat4.inverse
+ */
```
z3dev, line 9: It would be best if the function name matched the file name. :)

packages/modeling/src/maths/mat4/invert.test.js
```diff
@@ -0,0 +1,29 @@
+const test = require('ava')
+const { create, invert, fromTranslation, fromXRotation } = require('./index')
+
+const { compareVectors } = require('../../../test/helpers/index')
+const vec3 = require('../vec3/index')
+
+test('mat4: invert() translate ', (t) => {
```
z3dev, line 7: Excellent! Hanks for taking the extra time.

z3dev (changes requested):
Just one little change. Sorry.

packages/modeling/src/maths/mat4/invert.js
hrgdavor: no need to say sorry 😺 I intended to rename both. I just forgot to rename function inside.
decided to name it invert  to match gl-matrix

Nice catch, thanks.

z3dev (approved):
👍

PR #863, opened 2021-06-11
Feat : Enhance Renderer

z3dev:
These changes accomplish two things:
- simplifying the draw commands, splitting out drawLines from drawMesh
- adding simple event handing to demo.html, allowing the demo design to be rotated

All demo applications work as expected. Test cases for regl-renderer all pass.

P.S. demo.html now becomes the simplest JSCAD application. :)

This includes a small change to the WEB UI viewer, which adds the drawLines command for initialization.
 
### All Submissions:


z3dev:
@hrgdavor Can you review these changes?

FYI, demo.html will render the demo design properly AFTER 'npm run build'.

packages/utils/regl-renderer/demo.html
```diff
@@ -137,6 +147,34 @@
...
+
+const enterHandler = (ev) => {
+  lastX = ev.pageX
+  lastY = ev.pageY
+
+  ev.preventDefault()
+}
+
+const moveHandler = (ev) => {
+  rotateDelta[0] = ev.pageX - lastX
```
hrgdavor, line 166: use += for delta so it does not skip stesp, otherwise, looks nice, This also a nice sample how to run render without most ! 😸 
```js
  rotateDelta[0] += ev.pageX - lastX
  rotateDelta[1] += lastY - ev.pageY
```

hrgdavor (commented):
looks good, small change requested

hrgdavor:
This code is so nice and simple, it would be a good candidate for me to try-out OffscreenCanvas
 - with fallback
 - it is supported on most, but still not 100%

If I put render and model evaluation both in same worker then there is no need for postMessage 
 - postMessage used when OffscreenCanvas is not supported.

z3dev:
> This code is so nice and simple, it would be a good candidate for me to try-out OffscreenCanvas
> 
> * with fallback
> * it is supported on most, but still not 100%
> 
> If I put render and model evaluation both in same worker then there is no need for postMessage
> 
> * postMessage used when OffscreenCanvas is not supported.

That would be very cool.
Is there some way to detect the off screen canvas support?

hrgdavor:
> That would be very cool.
> Is there some way to detect the off screen canvas support?

simple check like this should be sufficient
```js
if (typeof OffscreenCanvas !== "undefined")
```

z3dev:
> This code is so nice and simple,

Definitely! I think this is the result of several months of hard work on the rendering. Thanks a lot!

mister-jscad (approved):
Good to go.

PR #864, opened 2021-06-11
use isIdentity

hrgdavor:

improve no-op check
 - avoid creating a new matrix just to check if it is a no-op matrix


z3dev:
@hrgdavor Obviously, we are headed in the same direction. :)

Can you review #865 pull request? I think that has a slightly name for the new comparison function. (I did some DOC changes as well, as the math documentation was kind of sick.)

After #865 is merged to master, you can sync and update the geometry code to use the new function.

z3dev:
Hmm... why did the build fail?

hrgdavor:
> Hmm... why did the Travis CI fail?

I forgot to add it to index.js 😄 

> After #865 is merged to master, you can sync and update the geometry code to use the new function.

yes, np :)

hrgdavor:
> After #865 is merged to master, you can sync and update the geometry code to use the new function.

updated to use `isIdentity`, should be ok to merge now

z3dev (approved):
Good! Thanks for following up with these changes.

PR #866, opened 2021-06-12
change applyTransforms and transform functions to clone geometry and keep attached properties other that just geometry data

hrgdavor:

I have looked in the source and applyTransforms 
 - it mutates original geometry (I think needlessly so) 
 - it is only used in 10 files
 - the geometry is not used, only the transformed part (sides, polygons)
 - there is no need to mutate the original geometry

I have taken just geom2 to demonstrate that it is an easy fix that would make some other things possible and the operation safer.

This is a very low impact change looking at what code is changed to do it. But is is a huge deal to me to keep immutability intact.

related to #859 

I hope this direction is acceptable, and if so I would clean the rest of applyTransforms


z3dev:
Just a little background, applyTransforms() was a 'hidden' function initially. And only geom2.toSides() / geom3.toPolygons() / path2.toPoints() would call applyTransforms. Basically, a 'private' function which really isn't private in JavaScript.

Later, there was a need to call applyTransforms(). See core/src/code-evaluation/serializeSolids.js 
The reason was to make sure that the transforms were applied BEFORE rendering / exporting solids. (Fixing a bug as rendering has issues with mirroring transforms, etc.)

So, instead of hiding applyTransforms(), the function was exposed as part of the API of each geometry.

I understand why you want to make the changes, and move back to a 'private' function.

z3dev:
Next comment... using 2D geometry as an example. A geometry can be used several times, in any order. So, all functions of geom2 return a new geometry (except applyTransforms()). 
- create() creates a new geometry
- clone() create a new geometry
- from*() call create() to create a new geometry
- invert() creates a new geometry
- transform() creates a new geometry

The only exceptions are applyTransform(), discussed above. And toSides() which is the 'proper' way to access the sides (after applying the transforms).

We have to be very careful to make sure all the functions adhere to the 'immutable' rule.

z3dev:
Last comment... using geom2 as the example again...

Internal storage (sides) and transforms must be maintained carefully. If the transforms change then the sides need to be adjusted to reflect the change in transforms. If the sides change then the transforms need to be adjusted to reflect the change (reset).

We are trying to allow applications to add 'attributes' easily, such as adding a 'color' attribute. That's easy for the users. But what adjustments to the geometry is not clear.

One thing that definitely needs to happen is the reset of the transforms in applyTransforms(). The current logic needs to change to this:

```
  // mat4.identity(geometry.transforms) // BAD as this modifies the transforms in place, which is not good if shared
  geometry.transforms = mat4.create()
```

hrgdavor:
@z3dev if you are ok with this direction, I would like to make the changes so new clone function works and attributes like color are maintained.

serializeSolids would be changed to generate new geometries 
 - new array of polygons/sides is created anyway in current implemention so it would not be a big overhead

at first glance it is not even a big fix, I can do it rather quickly and it would enable some nice future features like connectors

z3dev:
> @z3dev if you are ok with this direction, I would like to make the changes so new clone function works and attributes like color are maintained.

I'm not convinced by the small set of changes in this pull request. The APi of the geometry is changed by this pull request, so why keep applyTransforms()? And how do these changes work together with some other changes to enable the maintenance of user attributes?

Maybe you can put this together with some additional changes, and then create some test cases to show that these changes work as expected.

And remember, all test cases have to pass to merge into master.

hrgdavor:
Ok, since you are not against the direction, I will do the implementation that passes all the tests.

Then we can discuss it along with what I find during the implementation.

z3dev:
FYI, you can proff the changes with just one of the geometries. Maybe path2 is the most difficult. :)

hrgdavor:
first bump in the road was `measureBoundingBox` but I think I found a nice solution that could be used in other methods that want cached value to not be copied.


```js
// create a cached bounding box that will not be copied when object is transformed
const setBoundingBox = (geometry, value) => Object.defineProperty(geometry, 'boundingBox', { value, enumerable: false, writable: false })
```

and small change how it is set on geometry
```js
geometry.boundingBox = [minpoint, maxpoint]
// ---->
setBoundingBox(geometry, [minpoint, maxpoint])
```

hrgdavor:
geom3 converted and tests pass.

a tricky bug was found, that was not a problem when applyTransform modified original

in modifiers/generalize the polygons are converted (transform applied) but transforms was not reset in cloned object
```js
clone.transforms = mat4.create()
```

z3dev:
A little feed back on… 

_first bump in the road was measureBoundingBox but I think I found a nice solution that could be used in other methods that want cached value to not be copied._

I’m planning on using caches for the measurements, based on #aaa . Performance tests showed that the cache is just as fast as setting an attribute.

z3dev:
> in modifiers/generalize the polygons are converted (transform applied) but transforms was not reset in cloned object, so I added the reset there

That's not good... There could be more issues with this approach in the IO libraries. Some extra testing will be required.

hrgdavor:
> > in modifiers/generalize the polygons are converted (transform applied) but transforms was not reset in cloned object, so I added the reset there
> 
> That's not good... There could be more issues with this approach in the IO libraries. Some extra testing will be required.

which part is not good ? bug or the fix for the bug :) ?


ofc there could be more issues, but those issues are from badly behaved code.
Tests pass so far, and after I change also poly3 I will test the  examples too.

I am fine with doing some extra testing, as it is a significant internal change (but for the better 😃  )

packages/modeling/src/geometries/geom3/transform.js
```diff
@@ -15,10 +13,9 @@ const create = require('./create')
  * let newgeometry = transform(fromXRotation(degToRad(90)), geometry)
  */
 const transform = (matrix, geometry) => {
-  const newgeometry = create(geometry.polygons) // reuse the polygons
-  newgeometry.isRetesselated = geometry.isRetesselated
```
z3dev, line 19: FYI, isRetesselated needs to stay for now. There are some additional plans to remove this, but some changes to the booleans are required.

hrgdavor: well, that one is copied along with everything else... the problematic one was bounding box that needed to not be copied :)

packages/modeling/src/geometries/geom2/toSides.js
```diff
@@ -1,8 +1,7 @@
 const applyTransforms = require('./applyTransforms')
 
 /**
- * Produces an array of sides from the given geometry.
- * The returned array should not be modified as the data is shared with the geometry.
+ * Produces a new array of sides from the given geometry.
```
z3dev, line 4: This is not true if the transforms are 'identity'. So the initial comment is correct, sometimes.

hrgdavor: changed the comment a bit to be more precise :)

z3dev: And sides should never be modified....

packages/modeling/src/operations/modifiers/generalize.js
```diff
@@ -67,6 +62,9 @@ const generalizeGeom3 = (options, geometry) => {
   // FIXME replace with geom3.cloneShallow() when available
   const clone = Object.assign({}, geometry)
   clone.polygons = polygons
+  // toPolygons called above will convert old polygons (apply transform and return the new polygons array)
+  // we must reset transforms to identity, otherwise the transform will be applied again in some other operation
+  clone.transforms = mat4.create()
```
z3dev, line 67: Why not use create(polygons)?

packages/modeling/src/geometries/geom3/transform.js
z3dev: I have the PR for the measurements completed, but something happened to Travis CI. So, hold on.

If you cannot wait, then you can make the same changes to your branch. There's only 4 files changed.

packages/modeling/src/operations/modifiers/generalize.js
hrgdavor: I did not want to change the code too much, just fix it so the transform matrix matches the new situation.
The original code uses cloning, and createPolygons would only keep the polygons.
Additionally, this code is the is in the direction of cloning that I am also aiming for.

packages/modeling/src/geometries/geom3/transform.js
hrgdavor: It is ok, still work to do here, and more testing, I will wait for that PR to test with it.

hrgdavor:
I have added few tests to confirm color is retained after transforms.
The changes are now in place.
I welcome suggestions for cleaning the code maybe and further tests. It is not important to push this quickly, it is better to take time and make it a good solid change.

will test and compare examples

this PR is now beyond just applyTransforms, it also changes how transform works, an Id am happy how it turned out.

packages/modeling/src/operations/modifiers/generalize.js
z3dev: Actually, a more cool javascript style would be…

`const clone = Object.assign({}, geometry, {polygons, transforms: mat4.create()})`

:)

hrgdavor: ok :)

packages/modeling/src/geometries/geom2/applyTransforms.js
```diff
@@ -2,25 +2,26 @@ const mat4 = require('../../maths/mat4')
 const vec2 = require('../../maths/vec2')
 
 /*
- * Apply the transforms of the given geometry.
- * NOTE: This function must be called BEFORE exposing any data. See toSides().
+ * Apply the transforms of the given geometry and return the sides.
```
z3dev, line 5: I think the old comments are correct now.

z3dev:
I think there will be performance degradation due to these changes. For example...

```
const geometry = translate([50, 50, 50], geometry)
const epsilon = measureBoundingBox(geometry)
const polygons = geom3.toPolygons(geometry)

```
This is a very common set of function calls. Even the boolean functions execute these functions for each geometry.

The first call to measureBoundingBox(geometry) calls geom3.toPolygons(geometry) internally. And uses the polygons to perform the calculations.

Note that the original geometry is not changed. It still has the same transforms, and the same polygons.

The second call to geom3.toPolygons(geometry) is called to obtain the polygons for boolean operation.

Each of the calls to geom3.toPolygons(geometry) is going to run the very costly process of applying transforms to each polygon. In the case of measureBoundingBox(), those 'new' polygons are thrown away.

UG

hrgdavor:
> I think there will be performance degradation due to these changes. For example...

Yes that is true for the current implementation of measureBoundingBox, but it could be changed in a small way not only to bring  back performance to the same level but to improve it further.

I could implement measureBoundingBox performance improvements after you merge current cache code for measureBoundingBox.

## how would it be done ?
When you have a geometry with thousands of polygons that has a transform it is a good example to compare current approach with the new approach I would propose for measureBoundingBox

1 - Current approach
 - apply transform to convert 1000 polygons
 - go through 1000 polygons and calculate boundingBox
 - return boundingBox

2 - Proposed approach (that would bring performace gains even in current master branch)

- go through 1000 polygons and calculate identity-boundingBox
- apply transform to identity-boundingBox
- return boundingBox

both those approaches cache the final result, but the second approach can be improved even further

2a - For example: if we have 100 copies of this same geometry that is translated all over the scene a further performance improvement could be made by caching the idenitity-boundingBox
- check if there is cached identity-boundingBox for the polygons array
- if not cached, go through 1000 polygons and calculate identity-boundingBox, then cache it
- create new boundingBox from idenitity-boundingBox clone + apply transform
- return boundingBox

## further steps in this direction
calling toPolygons should be avoided more agressively as it is a costly operation, I have in mind  future proposals where it could be avoided even more
-  in serializeSolids (renderer could be changed to apply transforms inside GPU)
 - avoid calling toPolygons in boolean operations (but only for the first geometry) ...

packages/modeling/src/geometries/geom2/applyTransforms.js
hrgdavor: I have commited the code that passes tests, to break-up the changes into smaller steps. 
If you agree, I would go even further and remove applyTransforms as they are only used in toSides, toPolygons, toPoints.

z3dev:
_When you have a geometry with thousands of polygons that has a transform it is a good example to compare current approach with the new approach I would propose for measureBoundingBox

1 - Current approach

    apply transform to convert 1000 polygons
    go through 1000 polygons and calculate boundingBox
    return boundingBox

2 - Proposed approach (that would bring performace gains even in current master branch)

    go through 1000 polygons and calculate identity-boundingBox
    apply transform to identity-boundingBox
    return boundingBox

both those approaches cache the final result, but the second approach can be improved even further_

@hrgdavor this alone is worth changing. this would be a huge performance improvement for many of the heavy functions, like the booleans. how about doing this first?

hrgdavor:
> how about doing this first?

@z3dev  no problem, I'll do it :)

z3dev:
> > how about doing this first?
> 
> @z3dev no problem, I'll do it :)

Sorry. As a separate PR, please. :)

hrgdavor:
Unfortunately, the method I have described only works for transform matrix that is combination of transform+scale. 
Any rotation or skewing will produce incorrect bounding box with this new method.

I will still introduce that PR that will benefit situations where geometry is only translated+scaled, regardless that it can not optimize cases with rotation.
(if possible I will include this optimization for 90,180 deg rotation too)


Also it is still possible to mitigate the issues you mentioned with performance regarding measurements and toPoints.
A simple cache for the toPoints using WeakMap will be suffiecient.
That caching of toPoints is only relevant for this PR, so I will implement it in this PR as performace-hit mitigation.
With this additional caching we still get to have immutability, and avoid the performance degradation.

packages/modeling/src/geometries/geom2/toSides.js
```diff
@@ -11,6 +15,13 @@ const applyTransforms = require('./applyTransforms')
  * @example
  * let sharedsides = toSides(geometry)
  */
-const toSides = (geometry) => applyTransforms(geometry).sides
+const toSides = (geometry) => {
+  let sides = cache.get(geometry)
+  if (sides) return sides
+
+  sides = applyTransforms(geometry).sides
+  cache.set(geometry, sides)
```
z3dev, line 23: applyTranforms() returns a NEW object every time, so why cache?

hrgdavor: without caching, if script calls toSides(geometry) multiple times those transformed polygons will be regenerated every time.
this could happen for example
 - during measureBoundingBox first time
 - for booleans or serializeSolids second,third... time

if we have a geometry (translated sphere)
  - toSides will call applyTransforms 
  - applyTransforms generates geometry polygons with transformations applied, and identity transform
  - toSides takes only the polygons and returns them
  - generating a geometry in applyTransforms is pointless


Actually with this PR, applyTransforms becomes obsolete in a way, and has overhead of creating a new geometry that will never be used, only the generated polygons will currently.

Transform from now on just changes transforms matrix and does not use applyTransforms

applyTransforms code could be merged into: toSides, toPoints, etc

hrgdavor:
Closing as code has changed since and this PR needs a fresh start, and maybe even to be split into multiple PRs

PR #868, opened 2021-06-16
reorg(modeling): changed to use caches (WeakMap) for measurements

z3dev:
These changes update the measurements to use caches (WeakMap) instead of setting attributes on geometries. Previous performance tests have shown that the caches are just as fast as setting attributes.

NOTE: This is part of the effort of support users that want to set attributes, such as name or color.

Changes based on a previous PR #738 by @ahdinosaur 

### All Submissions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "10"
   - "12"
   - "14"
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "10"
   - "12"
   - "14"
 sudo: false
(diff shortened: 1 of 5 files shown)


mister-jscad (dismissed):
Looks fine.

hrgdavor:
@z3dev I have left you a comment about travis-ci, just checking if you have seen it
....
looks like they may be forcing migration from .org to .com
https://travis-ci.org/github/jscad/OpenJSCAD.org/requests
https://travis-ci.community/t/how-to-migrate-from-org-to-com-in-the-light-of-org-shutting-down/10566

z3dev:
@hrgdavor I did! It was very helpful.

I’ve been discussing with Travis support, and making some progress.

The builds are running… but something else is still not right.

mister-jscad (dismissed):
Looks fine, just like before.

mister-jscad (approved):
Yes. Yes. Yes. Do it!

PR #869, opened 2021-06-18
Performance improvement for measure bounding box

hrgdavor:

This should improve measureBoundingBox performance for cases when geometry is translated or scaled.
function is added to mat4 isOnlyTransformScale to help recognize such matrix.

The idea is to measure bounding box of the geometry without the transform, and then apply transform to the bounding box.

### DIFF
--- a/packages/io/svg-deserializer/tests/instantiate.test.js
+++ b/packages/io/svg-deserializer/tests/instantiate.test.js
@@ -249,6 +249,7 @@ test('deserialize : instantiate svg (path: arc) to objects', (t) => {
   t.is(shape.points.length, 15) // segments double on a 3/4 circle
   t.deepEqual(measurements.measureBoundingBox(shape), [[64.91110599999999, -77.611103, 0], [90.21850570104527, -52.30370029895471, 0]])
   shape = observed[1]
+  console.log('shape',shape)
   t.is(shape.points.length, 15) // segments double on a 3/4 circle
   t.deepEqual(measurements.measureBoundingBox(shape), [[50.79999599999999, -136.03302387090216, 0], [72.27222493929787, -110.6793647936299, 0]])
 })
--- a/packages/io/svg-deserializer/tests/instantiate.test.js
+++ b/packages/io/svg-deserializer/tests/instantiate.test.js
@@ -249,7 +249,6 @@ test('deserialize : instantiate svg (path: arc) to objects', (t) => {
   t.is(shape.points.length, 15) // segments double on a 3/4 circle
   t.deepEqual(measurements.measureBoundingBox(shape), [[64.91110599999999, -77.611103, 0], [90.21850570104527, -52.30370029895471, 0]])
   shape = observed[1]
-  console.log('shape',shape)
   t.is(shape.points.length, 15) // segments double on a 3/4 circle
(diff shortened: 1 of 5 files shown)


z3dev:
UG. This is disappointing. Having to test the transforms is definitely not ideal.

I was having the same question… especially about a transform that mirrors the geometry.

hrgdavor:
> UG. This is disappointing. Having to test the transforms is definitely not ideal.

well, identity check is everywhere as well :) there is small performance overhead ofc. that would be significant if geometry would have under 10 points and if there would be thousands of them. Also caching makes hitting the measurement code less often.

hrgdavor:
I have a possible solution that could be used for any type of transform but would make a little bit wider bounding box.

If we could loosen the requirement for bounding box (at least form some operations) we could trade precision for speed
 - it does not have  be as tight as now
 - is only required to catch the geometry bounds approximately
 - is only required that geometry does not leak outside of the bounding box

The idea is such that we use the initial bounding box (before transforms) as an approximation for our geometry.
Then we have a cuboid that approximates our underlying complex geometry and calculate any future boundingBox from that cuboid (constant time, regardless of geometry complexity).


here is an example of what I had in mind and how different is precision bounding box (red) and fast bounding box green
http://3d.hrg.hr/jscad/three/threejscad.html?uri=#model.measureBoundingBox.js

image

1) blue cylinder and red bounding box
2) blue cylinder and red bounding box, both rotated
3) precision bounding box based on the cylinder points
4) fast bounding box based on approximation
5) visualisation of different size calculated for precision bb and fast bb

In case a precise and tight box is also needed, we could have 2 versions (one for speed and one for precision)

z3dev:
I believe the requirement is for precision as designs often use the bounding box to calculate offsets. I often use the bounding box to calculate dimensions (H/W/D), and position pieces for boolean operations.

A non-precision version might be useful for internals, such as determining if two shapes overlap. This may be the only use… I’ll post back if there are additional uses cases.

hrgdavor:
@z3dev I have found an additional path to improve performance of measureBoundingBox.

I we allow internall for geometry creation functions to calculate bounding box and provide it to measureBoundingBox's cache we can allow fast calculation of bounding box for cases where it can be calculated form geometry parameters.

I have added an example implementation for ellipsoid (when default axes are used)
```js
  const geom = geom3.create(polygons)

  // if default axes are used, bounding box is trivial to calculate
  if (noAxes) {
    // we just use the radiuses to easily calculate the corners for bounding box
    // and we also account for when center points are != 0
    const boundingBox = [
      radius.map((p,i) => -p + center[i] ),
      radius.map((p,i) =>  p + center[i] )
    ]
    measureBoundingBox.setCache(geom, boundingBox)
  }

  return geom
```
and this could be done for number of geometries, also would allow such performance shortcut for external libraries too.

Even when there is no cool trick to calculate boundingBox, I think it could be beneficial to calculate it right away as geometry points are added in some cases, as it should be faster as it can avoid a second pass through all the points later.

hrgdavor:
@z3dev please review, or comment if further improvements are needed here. Although this PR did not deliver performance for all transforms, I think it is still worth merging.

I would like to finish this one, and then look to make this alternative less precise(but fast) bounding box in separate PR.
That one could still be used by some operations like booleans the check overlap, but is not suitable for align.
The bounding box with less precission still guarantees the geometry is not outside of it, and should bring some more performance gains.

hrgdavor:
reverted

hrgdavor:
my first attempt at proper merge :)

Issue #812, opened 2021-04-30
regl-renderer : improve performance of viewer


z3dev:
## Expected Behavior

Produce native buffers for use with WEB GL.

## Actual Behavior

Provided geometries (solids) are converted to entities for use in WEB GL, but the buffers are arrays of Number values. This results in an unnecessary conversion inside the regl library. (See line 122 of https://github.com/regl-project/regl/blob/master/lib/buffer.js that copies the array)

## Steps to Reproduce the Problem

  1. Load a large JSCAD design which take a long time to render

## Specifications

  - Version: V2+
  - Platform: WEB
  - Environment: browser

hrgdavor:
@z3dev I have draft branch for this that works ok, but some major things came up that are far greater killer of performance...
more details in a new discussion #923 .

z3dev:
First, I still think that rendering the geometry in the web worker is the the best direction to pursue. That allows everything to happen in the web worker, as well as rendering optimizations can be made.

Second, yeah. Geometry can be converted to buffers and posted back (maybe even transferable). That's why there are to/fromCompactBinary() functions. But that didn't turn out as planned, as general attributes cannot be passed as part of the buffers, so users are kind of 'out of luck' if they want to specify color, name, id, etc. as part of the design. It may not sound like a BIG issue but there are lots of use cases (many that we don't know) which will certainly use attributes.

Third, V2 has just been released within this year. And making the BIG changes to data structures means... yeah... V3. There's no way around this as there are some nice websites / applications being built with V2 modules, which was one of the major goals. V2 needs to continue on for a while longer.

So, what to do... i'm not 100% certain... but some playing around with a new web ui is certainly justified... and try some of these BIG improvements... maybe create a prototype

hrgdavor:
postMessage to be fast does not need to have all data as native buffers

if you have a geometry: `{ color:[..], transforms:[..], polygons:[..]}` 
  - convert it to make a copy that uses native buffers 
  - new object will have all user properties but will replace `polygons` with native buffers `vertices`, `indices`, ``
  - fill `trensferables` a separate array that holds reference to each transferable(native array)
  - at the end postMessage([goems],[transferables])

```js
const prepareGeom = (geometry, transferables)=>{
  let out = {...geometry}
  delete geometry.polygons
  let vertices = ... do conversion
  let indices = ...  do conversion
  let colors = ... do for whn color per vertex is used
  transferables.push(vertices)
  transferables.push(indices)
[...]

Issue #691, opened 2020-09-20
V2 - Issues when mirroring...


SimonClark:
## Expected Behavior
Mirroring a shape produces a mirror of the shape.

## Actual Behavior
Instead, I get some weird inverted shape from the dark dimension.  If I invert that, then I get the shape I want inside out.  If I invert that, then I get the shape I actually want.

## Steps to Reproduce the Problem

 ```
const jscad = require('@jscad/modeling')
const { maths, utils, geometries } = jscad

const { polygon, cuboid, cylinder } = jscad.primitives
const { extrudeLinear } = jscad.extrusions
const { center, translateZ, translate, rotateX, mirrorY } = jscad.transforms
const { union, subtract } = jscad.booleans
const { measureAggregateBoundingBox } = jscad.measurements

const options = {
  outerFrameRadius: 2.5,
  backThickness: 4,
  screwRadius: 2,
  screwInset: 8,
  supportArmLength: 40,
  supportArmThickness: 10,
  supportArmRidgeThickness: 3,
  supportArmAngle: utils.degToRad(4),
  cornerRadius: 5,
  clipThickness: 10,
  clipHeight: 10,
  clipOverhang: 5
}

const main = (params) => {
  options.totalHeight = options.clipThickness + options.supportArmThickness + options.cornerRadius

  let supportArm = SupportArm()
  let clip = Clip()
  let bounds = measureAggregateBoundingBox(clip, supportArm)
  console.log(bounds)
  let back = Base(bounds[0][0],bounds[1][0])
  let fullClip = union(supportArm, clip, back)
  return [
    fullClip,
    translate([0,-20,0],mirrorY(fullClip)),
    translate([0,-70,0],geometries.geom3.invert(geometries.geom3.invert(mirrorY(fullClip))))
  ]
}

function SupportArm() {
  let x = options.outerFrameRadius
  let y = - options.outerFrameRadius - options.backThickness
  const points = [[x, y]]
  points.push([ x, y = 0 ])
  points.push([
    x = x - Math.sin(options.supportArmAngle) * options.supportArmLength,
    y = y + Math.cos(options.supportArmAngle) * options.supportArmLength
  ])
  points.push([x + 3, y])
  points.push([x = x + y, y = 0])
  points.push([x, y = -options.backThickness])
  const poly = polygon({ points: points.reverse() })

  const mainSupportArm = translateZ(
    options.clipThickness + options.cornerRadius,
[...]

PR #871, opened 2021-06-24
fix renderer to render mirror transforms correctly

hrgdavor:
### All Submissions:

I was looking into code to see how to defer transforms to render instead of converting all geometry data with applied transforms that is done in serializeSolids.

The renderer already supports transform matrix, so the serializeSolids step is not necessary at all when sending to renderer. This will also open up more optimizations for rendering same geometry that is on multiple locations (examples/core/colors/colorCube.js)


z3dev:
Wrong!
The rendering cannot handle MIRROR transforms.

hrgdavor:
> The rendering cannot handle MIRROR transforms.
> see #691

np, I will check it out. I have seen mirror transform checks in some places. 
I will try to make it work with mirroring too :)

hrgdavor:
ok, I found a fix for first part. A mirror matrix that has determinant < 0 reverses order of vertices, and opengl culling and face direction is wrong. Best option is to revese vertices before sending to opengl. and transform matrix can be kept as is.

in my test sample, mirrored objects have flipped faces
and is obvious that shader does not take into account geometry transform for lightsource location
image

after adding the code that reverses vertices for mirror transforms it looks like this
image

and it should loke like openjscad.xyz , like this:
image


Now I will look into fixing the shader so the lighting is correct

z3dev:
This is slightly better, but a better fix would be correcting the camera. 😌 

Please test well. And be sure these fixes also correct the original issue.

hrgdavor:
> This is slightly better, but a better fix would be correcting the camera. 😌

this is just partial fix,

The rest is not camera issue I think :) 
 - geometries are placed correctly and the shape is correct in regard to camera
 - the lighting of the geoms looks like it does not take into account transform


> Please test well. And be sure these fixes also correct the original issue.

ofc :) the idea is that with this PR, transforming of geometry data is deferred to render, and render needs to be fixed to apply the transforms correctly in the shader ... shapes and positions and normals are ok ... I still have lighting to fix.

hrgdavor:
looks like someone already placed support for normals transform matrix (unormal) into the shader

I have managed to get it close to original (but not for mirrored transforms ... I will have to check why they mess-up the lighting ... still learning)

new code with transforms applied in render
image

original from openjscad.xyz
image

There is a slight difference as the shader is setup a bit unconventionally, so it is hard to pinpoint.
The shader for opensjcad has light following the camera in a way

I am learning about webgl as I go, and have some idea what can be fixed, this particular issue of lighting transformed objects is mentioned here:
https://learnopengl.com/Lighting/Basic-Lighting
you can look at the section titled `One last thing` with this explanation
image



Pls, bear with me. This is a path worth taking, for there is much performance to be gained as next example will show.

you can compare this test example and how fast it reacts to changes to parameters (and this is without caching geometries between model recalculations)
https://openjscad.xyz/?uri=http://3d.hrg.hr/jscad/three/testMirror.js
versus the same example with threejs as renderer.
http://3d.hrg.hr/jscad/three/threejscad.html?uri=testMirror.js

The speed difference is very noticable even with such small number of geometries on the screen, and all of the difference is in the renderer setup part .
The jscad part that generates the model is actually pretty fast and that is the core of the project that is most important form what I have heard sso far, the less important is the renderer, and it is a shame to let it be the cause of slower performance.

hrgdavor:
Looks like there is no need to calcualte normals differently for mirrored transform during the step that converts jscad model to webgl data. I reverted that part of the fix.

When unormal takes into account model transform, the model is not messed up inverted when mirrored transform is present. So next steps are focusing on providing unormal matrix that accounts for  geometry transformations.

To complete this I have to figure out combination of small changes to unormal calculation and probably small changes to shader to get the desired result.

Current implementation looks kinda ok. 
_These spheres are not just translated, each is rotated slightly to force out issues if lighting calculation is out of whack._

without mirror
image

with mirror 
image

mirorred spheres should not be rendered differently regarding the lighting, but they are right now. the viewing works ok, but is not quite there yet.

hrgdavor:
there was a whole host of discussion regarding this same issue of transforms like negative scale or mirror in threejs from 2014 
https://github.com/mrdoob/three.js/issues/4904

and it was at first met with resistance, but in the end it was implemented in 2017.

I have found that the sort of fix that was done there was already present in jscad but not activated. 
https://github.com/jscad/OpenJSCAD.org/pull/871/commits/257b086e01d23febb6135601a9a94bb20f7524a5

Ofc I could find it only after I did some digging and understood a bit better what needs to be done.
Good thing is that it is just a case of telling webgl which faces to cull(remove) and that makes it possible to render same geometry for all transforms (incl mirrors)

... I still have some work to do to reconstruct the bit strange dynamic lighting setup of the renderer

z3dev:
> there was a whole host of discussion regarding this same issue of transforms like negative scale or mirror in threejs from 2014
> [mrdoob/three.js#4904](https://github.com/mrdoob/three.js/issues/4904)
> 
> and it was at first met with resistance, but in the end it was implemented in 2017.
> 
> I have found that the sort of fix that was done there was already present in jscad but not activated.
> [257b086](https://github.com/jscad/OpenJSCAD.org/commit/257b086e01d23febb6135601a9a94bb20f7524a5)
> 
> Ofc I could find it only after I did some digging and understood a bit better what needs to be done.
> Good thing is that it is just a case of telling webgl which faces to cull(remove) and that makes it possible to render same geometry for all transforms (incl mirrors)
> 
> ... I still have some work to do to reconstruct the bit strange dynamic lighting setup of the renderer

@hrgdavor super investigation!

I took the easy route, and forced applyTransforms() before rendering. I think that was a good choice. 😃 

And now, you’re making the correct changes, and improving performance. This is very exciting.

hrgdavor:
> I took the easy route, and forced applyTransforms() before rendering. I think that was a good choice. 😃

Yes, definitely! wahtever easy fix can be done first is better than waiting forever for "prefect" solution :)
 
> And now, you’re making the correct changes, and improving performance. This is very exciting.

Well, I am trying. First I randomly change code and see what happens, and if that fails, I try to understand it 😄

hrgdavor:
### on openjscad.xyz
looking at openjscad.xyz current version
I see no pattern what do the different spheres reflect with their shiny spot, it is a bit all over the place
image

### currently in this PR
I have made a version that is consistent regarding any transformation. 
The base light is the same as current master, but shiny reflection is not

image

@z3dev please review when you have time
it is a good starting point if you would like to see some tweaks to the lighting 
If you can live with this version of lighting setup, please merge, I have some other improvements in mind that require this PR :)

### strange build fail 1/3 
image
```
lerna info Executing command in 20 packages: "npm run test"
@jscad/gcode-deserializer: > @jscad/gcode-deserializer@1.0.0 test /home/travis/build/jscad/OpenJSCAD.org/packages/io/gcode-deserializer
@jscad/gcode-deserializer: > ava './test.js' --verbose --timeout 2m
@jscad/gcode-deserializer: sh: 1: ava: not found
@jscad/gcode-deserializer: npm ERR! code ELIFECYCLE
@jscad/gcode-deserializer: npm ERR! syscall spawn
@jscad/gcode-deserializer: npm ERR! file sh
@jscad/gcode-deserializer: npm ERR! errno ENOENT
@jscad/gcode-deserializer: npm ERR! @jscad/gcode-deserializer@1.0.0 test: `ava './test.js' --verbose --timeout 2m`
@jscad/gcode-deserializer: npm ERR! spawn ENOENT
@jscad/gcode-deserializer: npm ERR! 
@jscad/gcode-deserializer: npm ERR! Failed at the @jscad/gcode-deserializer@1.0.0 test script.
@jscad/gcode-deserializer: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
@jscad/modeling: > @jscad/modeling@2.5.0 test /home/travis/build/jscad/OpenJSCAD.org/packages/modeling
@jscad/modeling: > ava 'src/**/*.test.js' --verbose --timeout 2m
@jscad/gcode-deserializer: npm WARN Local package.json exists, but node_modules missing, did you mean to install?
@jscad/gcode-deserializer: npm ERR! A complete log of this run can be found in:
@jscad/gcode-deserializer: npm ERR!     /home/travis/.npm/_logs/2021-06-26T22_39_53_000Z-debug.log
lerna ERR! npm run test exited 1 in '@jscad/gcode-deserializer'
lerna WARN complete Waiting for 2 child processes to exit. CTRL-C to exit immediately.
npm ERR! Test failed.  See above for more details.
The command "npm test" exited with 1.
cache.2
store build cache
```

z3dev:
Hmm… Why did the NODEJS 10 build fail? I suspect there are issues with Travis CI.
If this continues then let’s remove NODEJS 10 from the builds. It’s already obsolete.

hrgdavor:
> Hmm… Why did the NODEJS 10 build fail? I suspect there are issues with Travis CI.
> If this continues then let’s remove NODEJS 10 from the builds. It’s already obsolete.

I tried again, the build on node 10 complains about missing "ava"
https://travis-ci.com/github/jscad/OpenJSCAD.org/jobs/519517414
`sh: 1: ava: not found`
` npm WARN Local package.json exists, but node_modules missing, did you mean to install?`

12,14 passed ok

z3dev:
@hrgdavor please change .travis.yml, and remove NODEJS version 10.
I think this will allow the Travis CI to complete.

hrgdavor:
@z3dev it builds now. Strangely, some other branches work fine with nodejs 10. must be some travis issue

z3dev:
@SimonClark can you review this pull request as well. it's a little difficult at my current location (Oregon).

Thanks

z3dev:
The changes to regl-renderer and web should be split. That way, there will be two merges of the pull requests, each with a different topic. One merge for correcting the mirroring in the regl-renderer, and one for performance improvements in the web ui.

packages/utils/regl-renderer/src/geometry-utils-V2/geom3ToGeometries.js
```diff
@@ -77,6 +77,7 @@ const geom3ToGeometries = (options, solid) => {
       indices,
       colors,
       transforms,
+      flip: mat4.determinant(transforms) < 0,
```
z3dev, line 80: Instead of adding a new attribute, can this check be performed in drawMesh()?

z3dev (commented):
@hrgdavor please see the comment on the new ‘flip’ attribute. Let me know what you think?

hrgdavor:
@z3dev I have removed `serialize:false` and moved flip calculation to drawMesh. After this one is merged I will make a separate small PR for `serialize:false`.

z3dev (approved):
These changes look fine.
@SimonClark please review as I cannot test these changes. Thanks.

z3dev:
@hrgdavor This is so close! In general, I would say that the rendering is better and more consistent now.

However, it seems that the reflection of the light source kind of 'stops' when the object moves from in front to in back. For example, open the Primitives 3D Demonstration. Then rotate the scene about Z. The light source on the back objects (sphere and rounded cylinder) does not reflect properly.

hrgdavor:
I have played with the numbers. because I cant get the logic of the reflection for the old shader.
Now both this one and old one have the reflection unexplained but it is present from top and bottom

image

image

hrgdavor:
@z3dev is this something that would be acceptable ?

z3dev:
> @z3dev is this something that would be acceptable ?

Sure. I will do some more testing tomorrow.

z3dev:
@hrgdavor this is better now. the reflections are still a little strange at times, but improved over the current version. the rendering of mirrored shapes is far far better as well.

z3dev:
> @z3dev I have removed `serialize:false` and moved flip calculation to drawMesh. After this one is merged I will make a separate small PR for `serialize:false`.

@hrgdavor quick question... should i test with "serialize: false"?

hrgdavor:
Yes, now u have to try it with serialize:false to have the geometry with tranforms that are not just identity and polygons already changed

z3dev:
Retested with "serialize: false"

```
diff --git a/packages/core/src/code-evaluation/rebuildGeometry.js b/packages/core/src/code-evaluation/rebuildGeometry.js
index ac628d89..33367596 100644
--- a/packages/core/src/code-evaluation/rebuildGeometry.js
+++ b/packages/core/src/code-evaluation/rebuildGeometry.js
@@ -29,12 +29,13 @@ const rebuildSolids = (data, callback) => {
   const defaults = {
     mainPath: '',
     vtreeMode: false,
-    serialize: true,
+    serialize: false,
     lookup: null,
     lookupCounts: null,
     parameterValues: {}
   }

```

The rendering looks good for all the demonstration designs, as well as a set of designs that I created over the years.

z3dev (approved):
This is a big performance boost!

SUPER BIG THANKS to @hrgdavor

z3dev:
@hrgdavor Any thing left to test or change? If not then I'll merge these changes for regl-renderer

hrgdavor:
From my point of view, it is ready for merge. With this merged, the #812 is next on my list as a continuation of perf impr. 

812 will also benefit any setup where other renderer is used like threejs or babylon (they all need to get buffers for webgl somehow)

Issue #876, opened 2021-07-14
V2 expansions.expand of "simple" geom3 crashes


kbruneel:
## Expected Behavior
For the expand not to crash

## Actual Behavior
It slowly fills up the memory until the browser crashes.

## Steps to Reproduce the Problem
I have trouble reproducing it and needs some help with this. Currently I play with some user inputs until the browser freezes. Then I click the pause button in the javascript console and use the call stack to find the expand in which it happens. It's always the same one.

I have tried to reproduce the behavior in an isolated test by copying the data from the geometry that is the input to the failing expand. This results in the following test code:

```
        var polies = [
            jscad.geometries.poly3.fromPointsAndPlane([[-19.6100000, -0.8000000, 11.8550000], [-19.6100000, -0.8000000, -11.8550000], [-19.6100000, -2.7500000, -11.8550000], [-19.6100000, -2.7500000, 11.8550000], ], [-1, 0, 0, 19.61]),
            jscad.geometries.poly3.fromPointsAndPlane([[-19.2723800, -0.8000000, 11.8550000], [-19.2723800, -2.7500000, 11.8550000], [-17.3200000, -2.7500000, -10.0000000], [-17.3200000, -0.8000000, -10.0000000], ], [0.9960335028595003, 3.0983531671633697e-16, 0.08897899292211668, -18.141090198747712]),
            jscad.geometries.poly3.fromPointsAndPlane([[-17.3200000, -0.8000000, -10.0000000], [-17.3200000, -2.7500000, -10.0000000], [-14.1071400, -2.7500000, -11.8550000], [-14.1071400, -0.8000000, -11.8550000], ], [0.5000110003630168, -1.3211722663644432e-16, 0.8660190526287371, -17.320381052574824]),
            jscad.geometries.poly3.fromPointsAndPlane([[-17.3200000, -0.8000000, -11.8550000], [-17.3200000, -0.8000000, -10.0000000], [-14.1071400, -0.8000000, -11.8550000], ], [0, 1, -1.0483128474924838e-16, -0.8000000000000003]),
[...]

PR #877, opened 2021-07-16
Fix for #876 (expand)

hrgdavor:
### All Submissions:

fix for #876 
test now pass, need still to compare changed polygons visually and compare with jscad.xyz


packages/modeling/src/operations/booleans/trees/Node.js
```diff
@@ -99,28 +99,35 @@ class Node {
         current = stack.pop()
         continue
       }
+      let index = 0 // default
       if (!node.plane) {
-        let index = 0 // default
-        index = Math.floor(polygontreenodes.length / 2)
+        index = current.nextIndex || 0
```
z3dev, line 104: Please revert back as this will affect performance.

This will probably affect the tests again as well.

z3dev (changes requested):
See the comments on one change.

packages/modeling/src/operations/booleans/trees/Node.js
hrgdavor: this integral part of the fix. Previous code used middle point and hoped it would work ... and the same as with choosing any random point (I see in code sbdy also tried this) it will work only for some.

Th code must try other points or it will in some cases just be stuck in an infinite loop.

z3dev: If that’s the case then I really doubt the reliability of this fix. The starting point of the tree traversal shouldn’t matter.

These changes will have to be tested carefully.

hrgdavor: It does not matter, it would work the same if we start from the middle, but it matters because it complicates the stop condition (line 123 and line 130)

instead fo going from 0... length 
 - start from middle
 - go to end
 - wrap around to 0
 - go to middle-1 and then check if we need to stop splitting

hrgdavor:
there is more elegant version #898 . closing this one as obsolete.

PR #879, opened 2021-07-17
bump regl version to 2.1.0

hrgdavor:
### All Submissions:

seems to be working fine. If it is really ok, it will be smaller step if they implement webgl2

(@z3dev unfortunately, the latest version of regl is also just webgl1)

### DIFF
--- a/packages/utils/regl-renderer/package.json
+++ b/packages/utils/regl-renderer/package.json
@@ -31,7 +31,7 @@
     "camera-unproject": "1.0.1",
     "gl-mat4": "1.2.0",
     "gl-vec3": "1.1.3",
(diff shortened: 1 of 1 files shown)


z3dev:
@hrgdavor did you perform some manual tests of the demos? WEB UI?

if those work fine then this will be a nice change.

thanks again!

hrgdavor:
@z3dev I have tried demo page in regl renderer. And also npm run dev in packages/web

Loaded few models and did not see any errors ot afrtifacts. It was not extensive test, but looks ok.

z3dev (approved):
Great! It’s good to know that the code base is compatible with newer versions.

Issue #885, opened 2021-07-21
TypeError: Cannot read property 'isClosed' of undefined during svg deserialization


brucedjones:
## Expected Behavior

The SVG below is successfully deserialized

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="5.791015625" height="11.71875"><path d="M1.84-3.17L0.53-3.48L0.97-7.11L5.12-7.11L5.12-5.83L2.30-5.83L2.14-4.42Q2.32-4.52 2.60-4.60Q2.89-4.68 3.16-4.68L3.16-4.68Q4.22-4.68 4.79-4.05Q5.36-3.42 5.36-2.29L5.36-2.29Q5.36-1.61 5.06-1.05Q4.75-0.50 4.20-0.20Q3.65 0.10 2.90 0.10L2.90 0.10Q2.23 0.10 1.64-0.18Q1.05-0.45 0.72-0.94Q0.39-1.42 0.40-2.02L0.40-2.02L2.05-2.02Q2.07-1.63 2.29-1.40Q2.52-1.17 2.89-1.17L2.89-1.17Q3.72-1.17 3.72-2.40L3.72-2.40Q3.72-3.54 2.70-3.54L2.70-3.54Q2.12-3.54 1.84-3.17L1.84-3.17Z"/></svg>
```

## Actual Behavior

Deserializer throws `Cannot read property 'isClosed' of undefined`

## Steps to Reproduce the Problem

run
```
import { deserializers } from "@jscad/io";

const svg =
[...]

PR #886, opened 2021-07-21
fix 885

hrgdavor:
### All Submissions:

fix for #885

@z3dev I have added a test case and removed throw from appndPoints. The reason is that closed path only means as a marker for those using it to connect last and first point later on ... like when rendering. So throwing err if somebody appnds to a closed path seems unnecessary.

I might be to quick on this judgement, but at elast we have a test case here as starting point.


hrgdavor:
I will  revert the initial change and improve svg deserializer so it does not use append. Append is used that creates new geometry every time.
This is inefficient and could be avoided if deserializer collects points first and creates geometry in the end. and it would also prevent this error.

z3dev:
@hrgdavor thanks for quick response.

as you are probably understanding, there are lots of issues, I.e. lots of bugs. It’s important to understand and localize issues within a specific scope. Sometimes those will be issues in the modeling library and sometimes there are issues in the IO libraries.

I think you are on the correct track here. The SVG library needs improvements to create usable paths or geometry.

please continue to improve the SVG IO library. I think this PR can be closed.

hrgdavor:
@z3dev please review. This is a second attempt at fixing this.

The problem is that the test sample svg is malformed and now error is thrown
`Malformed svg path at 2:445. Path closed itself with command #29: Q2.12 -3.54 1.84 -3.17`

The problem is that the mentioned command adds last point that is the same as first point in the path, so the appendPoints closes it. So I have added a check for such case and throw an error with the location of the problematic svg.

hrgdavor:
added parameter as requested by user in #885 

pathSelfClosed: 'error' - default throw an error
pathSelfClosed: 'trim' - ignore it
pathSelfClosed: 'split' - add an additional path.

default is throwing error that also says this option is available:
`Malformed svg path at 2:445. Path closed itself with command #29: "Q2.12 -3.54 1.84 -3.17". to avoid this error use pathSelfClosed:'split' or pathSelfClosed:'trim' option`

packages/io/svg-deserializer/svgElementHelpers.js
```diff
@@ -108,7 +108,7 @@ const svgSvg = function (element, { customPxPmm }) {
   if ('HEIGHT' in element) { obj.height = element.HEIGHT }
   if ('VIEWBOX' in element) {
     const list = element.VIEWBOX.trim()
-    const exp = new RegExp('([\\d\\.\\-]+)[\\s,]+([\\d\\.\\-]+)[\\s,]+([\\d\\.\\-]+)[\\s,]+([\\d\\.\\-]+)', 'i')
+    const exp = /([\d.-]+)[\s,]+([\d.-]+)[\s,]+([\d.-]+)[\s,]+([\d.-]+)/i
```
z3dev, line 111: Why make these changes to the regular expressions? saving space?

Actually, these expressions should be created once at the file level, and reused. That would improve performance.

z3dev (changes requested):
A few comments and requests.

In general, these changes look good. Some extra manual testing should be done with the 'svg samples'.

packages/io/svg-deserializer/index.js
```diff
@@ -39,7 +39,7 @@ const shapesMapJscad = require('./shapesMapJscad')
  * @param {integer} [options.segments] - number of segments for rounded shapes
  * @param {string} [options.target] - target 2D geometry; geom2 or path2
  * @param {string} [options.version='0.0.0'] - version number to add to the metadata
- * @param {string} input - SVG data
+ * @param {string} [options.pathSelfClosed='error'] - [error||trim||split] if path self-closes with one of commands without stop command right after * @param {string} input - SVG data
```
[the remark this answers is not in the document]
hrgdavor: sry, not much experience with docstrings, could you please write in comment what it should look like

packages/io/svg-deserializer/svgElementHelpers.js
hrgdavor: standardx complained that new RegExp should not be used

packages/io/svg-deserializer/index.js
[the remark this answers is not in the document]
z3dev: just split the line, as the 'input' parameter is part of that long line.

packages/io/svg-deserializer/svgElementHelpers.js
hrgdavor: reusing regex with global flag can be dangerous if not careful
https://medium.com/codesnips/js-careful-when-reusing-regex-636b92c6bf07
https://stackoverflow.com/questions/1520800/why-does-a-regexp-with-global-flag-give-wrong-results

but it looks those 2 are not global, there is no /g flag so we should be ok

packages/io/svg-deserializer/shapesMapGeometry.js
```diff
@@ -215,182 +226,189 @@ const expandPath = (obj, svgUnitsPmm, svgUnitsX, svgUnitsY, svgUnitsV, svgGroups
...
         }
         break
       case 'T': // absolute quadratic Bézier shorthand
-        while (pts.length >= 2) {
-          cx = parseFloat(pts.shift())
-          cy = parseFloat(pts.shift())
+        while (pts.length >= i + 2) {
+          cx = parseFloat(pts[i++])
+          cy = parseFloat(pts[i++])
           paths[pathName] = geometries.path2.appendBezier({ segments, controlPoints: [[svg2cagX(qx, svgUnitsPmm), svg2cagY(qy, svgUnitsPmm)], [svg2cagX(qx, svgUnitsPmm), svg2cagY(qy, svgUnitsPmm)], [svg2cagX(cx, svgUnitsPmm), svg2cagY(cy, svgUnitsPmm)]] }, paths[pathName])
```
z3dev, line 340: should ensurePath() be called here?

z3dev (commented):
Left a comment as not sure if a change is required.

packages/io/svg-deserializer/shapesMapGeometry.js
hrgdavor: yes, nice catch. Thanks :)

hrgdavor:
Fixed the missing ensurePath() . I thnik this is ready now. @z3dev pls review

z3dev (approved):
OK. This looks pretty good. Thanks, @hrgdavor 

Please test well with some additional SVG files.

hrgdavor:
I have tested with few icons from iconmonstr
image
image
image

seems to be working well

PR #889, opened 2021-07-22
path2.isCCW

hrgdavor:
### All Submissions:

I have not found  function to check if path CW/CCW . Here is proposed function usually found on stackoverflow and similar.

if there is one already ... my bad :)


hrgdavor:
code in operations/hulls/hullpoints2.js should be also considered as it has a shortcut version for 3 points

```js
const ccw = (i1, i2, i3) => (points[i2][0] - points[i1][0]) * (points[i3][1] - points[i1][1]) - (points[i2][1] - points[i1][1]) * (points[i3][0] - points[i1][0])
```

and it check versus Number.EPSILON
```js
while (ccw(stack[M - 1], stack[M], stack[i]) < Number.EPSILON)
```

so the question is if implementation in this PR matches that one. It would be bad if they give different result in some cases, and it needs checking in that case which result is correct.

z3dev:
This is going to take a little bit of a discussion.

_Should the library even try to determine handedness for paths?_

In general, paths can have any shape, including paths that cross itself, open paths, closed paths, left flowing, right flowing, etc. There's no real 'direction'. And even closed paths are really nothing more.

However, users want to use paths to define shapes. These are easy to understand and construct. And sometimes, mistakes are made.

_What can make paths easier to use?_

Obviously, the handedness of a path becomes useful, as the proper order of points allows the path to be used in other functions, i.e. expansions, conversions, etc.

Even internally, some calculations are being performed. See expandPath2() and offsetFromPoints() in expansions.

A few utilities might be useful for both paths and 2D geometries.

_What to do about mistakes?_

User will make mistakes. There are some checks for parameters but nothing for point sets, etc. It's basically up to the user to supply properly ordered points, etc. And if there are errors then something deep down usually goes wrong. The error may be irrelevant or nothing at all.

hrgdavor:
My intention was not to use this all over the place to check every input. But when  writing a scrpt and i have unreliable source of paths i need this utility function. Also it was necessary for dealing with true type fonts to recognize holes.

> Should the library even try to determine handedness for paths?

No. It wuold be detrimental to performance.

> What can make paths easier to use?

I would welcome other useful utilities. 

> What to do about mistakes?

User gets unexpected result, then uses forums, user groups, docs, tutorials, stackoverflow. No automatic fixing as it is expensive in my opinion.

hrgdavor:
@z3dev just a reminder, to review or comment further 😃

z3dev:
Just some more information. This is not trivial, as makerjs is finding out.

https://github.com/microsoft/maker.js/issues/512

hrgdavor:
> Just some more information. This is not trivial, as makerjs is finding out.
> 
> [microsoft/maker.js#512](https://github.com/microsoft/maker.js/issues/512)

Not sure what you are referring to.

I am noticing an issue makerjs has that we do not "Fonts are imported as Bezier curves, which are then converted to arcs" and it does not preserve directions. We just generate points and I doubt we are destroying information.

Also I am interested if you think the function itself is difficult, and what issues are you seing with it ?

z3dev:
> > Just some more information. This is not trivial, as makerjs is finding out.
> > [microsoft/maker.js#512](https://github.com/microsoft/maker.js/issues/512)
> 
> Not sure what you are referring to.

The whole idea of handedness in paths is not easy to determine, and correcting the paths is even more difficult. That's what makerjs is finding out now.

The PR contains a very simple function to determine handedness in only one case; paths that define a closed trivial polygon.

> I am noticing an issue makerjs has that we do not "Fonts are imported as Bezier curves, which are then converted to arcs" and it does not preserve directions. We just generate points and I doubt we are destroying information.

Well... not really. Those paths (which include lines, arcs, bezier curves, etc) are coming from SVG files (font paths are just about the same). The issue is about making everything CCW for use with CNC machines.

> Also I am interested if you think the function itself is difficult, and what issues are you seeing with it ?

As noted above, I think there's a whole lot more to think about. Users want to use paths, but there are some rules before paths can be turned into 2D or 3D geometries.

I'm thinking that a 'cookbook' needs to be created which shows how to cook paths correctly. And maybe there are some additional functions required as well, such as breaking paths into sub-paths, etc.

Sorry, I'm turning this PR into a discussion.

hrgdavor:
I would argue that it is still useful. Users should know about it.

One option is to put the code in the discussion page for ppl to copy paste.

Another option is to add to merge it and in the documentation mention the limitations and link to the discussion.

If you still think it is better to not add it right now, I will create a discussion with the code example for ppl to use and play with it. And there more edge cases could be discussed and limitations of the function and when it is reliable and when it is not.

hrgdavor:
Closed in favor of a discussion page #929

PR #891, opened 2021-07-24
developer friendly syntax for getParameterDefinitions #875

hrgdavor:
### All Submissions:

implementation of #875

as a bonus, webrequire also detects if the script exports `getParameterDefinitions`. In that case parsed definitions are used first, and definitions from `getParameterDefinitions` are added/replaced (if the latter has definition with same name)

```js
const { cuboid } = jscad.primitives


function main({// @jscad-params
   width=400,
   height=28,
}) {
   return cuboid({size: [width,800,28]});
}

function getParameterDefinitions(){
   return [{name:"height", type:'int', caption: 'Height', inital:30}]
}

module.exports = { main,  getParameterDefinitions}
```

since both are used and `getParameterDefinitions` function has priority height parameter has caption `Height`.
image

also UI was tweaked to show proper line number if parameter definition has syntax error

```js
5  function main({// @jscad-params
6     width=400,
7     height=28,
8     hole=true,// {type:xxx}   
9  }) {
```

image


z3dev:
@hrgdavor These changes look fine. Please change the code to meet standards, and remove ‘strict’.

hrgdavor:
@z3dev I have removed strinct and a leftover console log.

I have run standardx, but one thing that ti complains `The Function constructor is eval.` is necessary, same as it is needed in code-evaluation.

If you have some other things that need changing in code style let me know :)

z3dev (changes requested):
A few quick comments and requests.

The BIG issue is that this will not work for NPM projects. One of the goals of the core is that any design can be used via NPM or WEB. I hope there’s any easy solution.

packages/core/src/parameters/getParameterDefinitionsFromSource.test.js
```diff
@@ -0,0 +1,165 @@
+'use strict'
```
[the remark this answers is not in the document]
hrgdavor: removed

packages/web/src/ui/views/status.js
```diff
@@ -4,9 +4,13 @@ const html = require('nanohtml')
 const status = (state, paramsCallbacktoStream) => {
   const status = state.status
   const errorMessage = status.error !== undefined && status.error.message ? `${status.error.message}` : ''
-  const errorLine = status.error !== undefined && status.error.lineno ? `Line: ${status.error.lineno}` : ''
+  let errorLine = status.error !== undefined && status.error.lineno ? `Line: ${status.error.lineno}` : ''
   const errorStack = status.error !== undefined && status.error.stack ? `Stack: ${status.error.stack}` : ''
 
+  // for now could not find a way to apss right line number from custom error that knows
+  // the correct line number the error originates. So added support for line number inside the message
+  if(errorMessage.indexOf('line:') !== -1) errorLine = ''
```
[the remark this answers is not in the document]
hrgdavor: sure, I can change to `if(errorMessage.toLowerCase().indexOf('line:') !== -1)`

hrgdavor:
> The BIG issue is that this will not work for NPM projects. One of the goals of the core is that any design can be used via NPM or WEB. I hope there’s any easy solution.

ok, ... not sure what you mean here... what npm projects use `getParameterDefinitions` except gui evaluating the script to generate UI ...(then it has the source and can use this as it is integrated to code-evaluation)

could you explain your concern, so I can see what could be done

hrgdavor:
I have written a few babel plugins, with jsx ind other tricks. I know my way around javascript AST. So even if it becomes a big issue, a plugin could be made that transforms code and injects `getParameterDefinitions`

https://github.com/hrgdavor/babel-plugin-jsx-simple

z3dev:
> > The BIG issue is that this will not work for NPM projects. One of the goals of the core is that any design can be used via NPM or WEB. I hope there’s any easy solution.
> 
> ok, ... not sure what you mean here... what npm projects use `getParameterDefinitions` except gui evaluating the script to generate UI ...(then it has the source and can use this as it is integrated to code-evaluation)
> 
> could you explain your concern, so I can see what could be done

My mistake. Parameters can be passed via the CLI as options.

packages/core/src/code-loading/webRequire.js
```diff
@@ -110,6 +111,11 @@ const makeWebRequire = (filesAndFolders, options) => {
           _compile: (content, fileName) => {
             const moduleMakerFunction = new Function('require', 'module', content)
             moduleMakerFunction(_require.bind(null, entry.fullPath), matchingModule)
+
+            const paramDefFromSource = isMain ? getParameterDefinitionsFromSource(content, fileName) : []
```
z3dev, line 115: It's more important to parse definitions from the source WHEN there are definitions to be parsed, i.e. @jscad-params is present. That check only required a cheap scan of the content.

z3dev (changes requested):
First review with some suggested changes, and a few questions to discuss.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,176 @@
...
+    else if (l[i] === '\t') count += 2
+    else break
+  }
+  return count
+}
+
+const getParameterDefinitionsFromSource = (script) => {
+  const lines = []
+  let line
+  script.split('\n').forEach((l, i) => {
```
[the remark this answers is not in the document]
hrgdavor: \n should be enough, and trim will remove any lefotver \r

```js
'   dddd\r   '.trim()// -> dddd
```

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,176 @@
...
+
+const parseComment = (comment, line, paramName) => {
+  comment = extractTextFromComment(comment)
+
+  const ret = {}
+  const idx = comment.indexOf('{')
+  if (idx !== -1) {
+    try {
+      // eslint-disable-next-line
+      ret.options = Function('return ' + comment.substring(idx)).call()
```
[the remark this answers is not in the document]
hrgdavor: Ii is not about executing code, just about less strict syntax. If native JSON.parse is used then object keys also need quotes.

```js
// How many? {type:"slider", min:2, max:10}'
// must be changed to
// How many? {"type":"slider", "min":2, "max":10}'
```

Another option is using json5 format https://json5.org/

we could add dependency to https://www.npmjs.com/package/json5

packages/web/src/ui/views/status.js
```diff
@@ -4,9 +4,13 @@ const html = require('nanohtml')
 const status = (state, paramsCallbacktoStream) => {
   const status = state.status
   const errorMessage = status.error !== undefined && status.error.message ? `${status.error.message}` : ''
-  const errorLine = status.error !== undefined && status.error.lineno ? `Line: ${status.error.lineno}` : ''
+  let errorLine = status.error !== undefined && status.error.lineno ? `Line: ${status.error.lineno}` : ''
```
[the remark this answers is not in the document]
hrgdavor: That is ok with me, I will add in the message too

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,176 @@
...
+  script.split('\n').forEach((l, i) => {
+    const trim = l.trim()
+    if (trim) {
+      lines.push({ code: trim, line: l, lineNum: i + 1, indent: countSpaces(l) })
+    }
+  })
+
+  let i = 0; let lineNum; let code; let prev; let prevIndent
+  while (i < lines.length) {
+    line = lines[i].code.trim()
```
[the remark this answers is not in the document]
hrgdavor: 👍

packages/core/src/code-loading/webRequire.js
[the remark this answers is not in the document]
hrgdavor: Actually I am not a big fan of that, but If you prefer I can revert this to look for parameters in all files.

Personally for multifile project a file for component would contain parameter definitions when I want to test the component standalone by loading it alone. I would prefer parameters to be controlled only by top-level script.

packages/core/src/parameters/getParameterDefinitionsFromSource.test.js
```diff
@@ -0,0 +1,201 @@
...
+    'password',
+    { name: 'password', type: 'password', caption: 'Secret?' }
+  )
+})
+
+test('param slider', (t) => {
+  testBothDir(t,
+    '// How many? {type:"slider", min:2, max:10}',
+    'count',
+    { name: 'count', type: 'slider', min: 2, max: 10, caption: 'How many?' }
```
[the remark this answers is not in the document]
hrgdavor: ok

packages/core/src/code-loading/webRequire.js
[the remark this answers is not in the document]
z3dev: That's correct. Those parameters definitions can be used to interactively debug the component.

But can also be added to a larger list of definitions as well. 

These changes should be flexible, and allow definitions to be used any number of ways.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,175 @@
...
+
+const parseOne = (comment, code, line1, line2) => {
+  let def = parseDef(code, line2)
+  const { caption, options } = parseComment(comment, line1, def.name)
+  def.caption = caption || def.name
+  if (options) {
+    def = { ...def, ...options }
+    if (def.type === 'checkbox' && 'initial' in def) def.checked = true
+    if (def.type === 'slider') {
+      if (def.min === undefined) {
```
z3dev, line 88: I just saw this now. Actually, there are defaults for 'range' input types, which are min 0, and max 100.

https://www.w3schools.com/html/html_form_input_types.asp

packages/core/src/parameters/getParameterDefinitionsFromSource.js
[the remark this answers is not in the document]
z3dev: OK. I understand this one fine but I don't have to like it. ;)

packages/web/src/ui/views/status.js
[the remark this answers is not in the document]
z3dev: I like the changes to the error messages above.

Are these changes necessary now?

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,175 @@
...
+    } else if (/^[0-9]+$/.test(initial)) {
+      ret.type = 'int'
+      ret.initial = parseFloat(initial)
+    } else if (/^[0-9]+\.[0-9]+$/.test(initial)) {
+      ret.type = 'number'
+      ret.initial = parseFloat(initial)
+    } else {
+      try {
+        // eslint-disable-next-line
+        ret.initial = Function('return ' + initial).bind({}).call()
```
z3dev, line 153: So... this is evaluating the variable definition in order to obtain the initial value. Why support anything but 'text' values?

I don't see why this should call 'return X'. That is really out of scope for these definitions. Can you provide some examples which make sense?

z3dev (changes requested):
Sorry. Another set of questions / requests.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: I had a vague idea that in the future a more advanced input field in customizer could return and array of numbers, or an object. The parameters are not performance critical, it would be ok to only support strings, and any fancy input could always use JSON in the value.

packages/core/src/code-loading/webRequire.js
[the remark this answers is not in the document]
hrgdavor: Ok, I will revert the check, and allow this to run on all compiled scripts.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
z3dev: That sounds reasonable as custom parameter definitions should be possible in the future, hopefully.

Obviously, some nice documentation or example will be required later.

packages/web/src/ui/views/status.js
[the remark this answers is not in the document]
hrgdavor: ok, reverted

packages/core/src/parameters/getParameterDefinitionsFromSource.js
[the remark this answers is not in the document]
hrgdavor: I moved both evals to JSON5 ... it still allows for JS like syntax 
 - using single quotes `{caption;'Height'}`
 - not required to quote keys `{initial:11}`
 - ...
 -

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: moved to JSON5, so eval should not be a concern any more, and arrays or objects are supported as well

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,174 @@
...
+      ret.checked = initial === 'true'
+    } else if (/^[0-9]+$/.test(initial)) {
+      ret.type = 'int'
+      ret.initial = parseFloat(initial)
+    } else if (/^[0-9]+\.[0-9]+$/.test(initial)) {
+      ret.type = 'number'
+      ret.initial = parseFloat(initial)
+    } else {
+      try {
+        ret.initial = JSON5.parse(initial)
```
z3dev, line 152: Hmm... I doubt JSON5 will every be adopted, but this does simply the code. And this is only for the WEB based UI as well. So, OK.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,174 @@
+const JSON5 = require('json5')
+
+const countSpaces = (l) => {
```
z3dev, line 3: In general, there should be a few comments in the code to help developers understand what's happening. Of course, the originator understands NOW, but how about in 3-5 years?

Note: These are not comments for JSDOC

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  return def
+}
+
+const extractTextFromComment = (c) => {
+  const prefix = c.substring(0, 2)
+  if (prefix === '//') c = c.substring(2)
+  if (prefix === '/*') c = c.substring(2, c.length - 2)
```
z3dev, line 104: This is assuming that the comment is like `/* ... */`, so why not check for that?

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,174 @@
...
+
+  return c.trim()
```
z3dev, line 106: This trim is not necessary.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
```diff
@@ -0,0 +1,174 @@
...
+    if (code.length > 12 && code.indexOf('@jscad-params') !== -1) break
+  }
+
+  let groupIndex = 1
+  const defs = []
+
+  while (i < lines.length) {
+    code = lines[i].code
+    lineNum = lines[i].lineNum
+    if (code[0] === '}') break
```
z3dev, line 35: So... if I understand correctly, this is the condition that says 'END OF PARAM DEFINITIONS'. Doesn't this assuming a certain coding style?

z3dev (changes requested):
A few comments and requests.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: In the contrary, my perception is that it is very well adopted https://www.npmjs.com/package/json5 37,466,988 weekly downloads. I found about it while writing babel plugins, babel uses it internally, and many libraries that want to for example use json for config files but want cleaner syntax that is closer to JS code. :)

packages/core/src/code-loading/webRequire.js
```diff
@@ -110,6 +111,11 @@ const makeWebRequire = (filesAndFolders, options) => {
           _compile: (content, fileName) => {
             const moduleMakerFunction = new Function('require', 'module', content)
             moduleMakerFunction(_require.bind(null, entry.fullPath), matchingModule)
+
+            const paramDefFromSource = getParameterDefinitionsFromSource(content, fileName)
```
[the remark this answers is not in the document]
hrgdavor: I am against it. `hasParameterDefinitionsInComments` would have to scan the code the same as `getParameterDefinitionsFromSource`, but then if it finds something, the scanning is repeated to get the parameters

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: yes, it expects closing bracket to be in separate line. and this whole thing is definitely for when there is one parameter per line.

```js
function main({//@jscad-params
  radius = 21
}){
  return sphere({radius})
}
// or
let defaults = {
//@jscad-params
  radius = 21
}

```

this would break current parser.

```js
function main({//@jscad-params
  radius = 21}){
  return sphere({radius})
}
```
I will add a check if the closing bracket is not in separate line, but in the same line as last parameter,

This whole concept must rely on certain style of defining the parameters, and `getParameterDefinitions` should be used when not wanting this style.

packages/core/src/code-loading/webRequire.js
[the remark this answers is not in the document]
z3dev: `String.includes('@jscad-params')` is hundreds of times faster then splitting the code into substrings, etc, etc.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
z3dev: > yes, it expects closing bracket to be in separate line. and this whole thing is definitely for when there is one parameter per line.

ok. just checking. please keep this information for the documentation.

packages/core/src/code-loading/webRequire.js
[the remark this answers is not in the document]
hrgdavor: ok, thnx for the hint, yes the check does not have to split the string like parsing does :) I will update it.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: I think I need a more proper handling of multiline comments, maybe leave `/* ...*/` to be handled same as `//` with aditional check and thow an error saying multiline comments not supported by parameters parser ?

or some  other ide what to do with `/* ... */` when single line or multiline ?

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: moved it to only if substring is used

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: added check to handle closing bracket in same line with last  parameter. It is not an exhaustive check, but will help in most cases.

packages/core/src/parameters/getParameterDefinitionsFromSource.js
hrgdavor: added better handling, and disallowed multiline comments. singliline `/* ... */` are allowed and work.

hrgdavor:
@z3dev please review :) I have covered I think, all of your comments, hopefully in satisfactory way. :)

z3dev:
> @z3dev please review :) I have covered I think, all of your comments, hopefully in satisfactory way. :)

Super! I will review tomorrow. If no more concerns then I will approve and merge! 😃

z3dev (approved):
OK. These changes are looking fine now.

Thanks @hrgdavor for the long discussions.

z3dev:
@hrgdavor there are important rules to follow when doing merges to master. the project uses LERNA which depends on exact formats for the merge comments.

the comment on this merge was not correct, so there will be no information in the release notes.

hrgdavor:
could you please give me an example, what would be a correct title for this merge? 

I read the doc on the link, but it is not clear to me still :)

z3dev:
Sure! Probably something like this… SQUASH MERGE… comment like…

`feat(web): enhanced to support parameter definitions via comments`

The ‘feat’ will create a new minor version.

LERNA will detect that comment and automatically increment the version for the WEB package, something like 2.4.0

It’s cool but very easy to mess up.

hrgdavor:
Thanks, I will ask next times, until I get used to it :)

PR #897, opened 2021-08-03
fix #114

### LINKED ISSUE #114: BUG: hull() fails with TypeError
function main() {
// BUG: If resolution is changed to 18 then there's an error
// "TypeError: this.points[p1] is undefined"
    var r = 18;
    var a = CAG.circle({center: [-5,0], radius: 3, resolution: r});
    var b = CAG.circle({center: [+5,0], radius: 3, resolution: r});
    b = hull(a,b);
    return b;
}

hrgdavor:
### All Submissions:

fix for #114 

M would go to 0 and `ccw(stack[M - 1], stack[M], stack[i])` would fail due to array index `-1`


z3dev:
> M would go to 0 and `ccw(stack[M - 1], stack[M], stack[i])` would fail due to array index `-1`

@hrgdavor why did M become negative? was there a special ordering of points that caused this to happen?

hrgdavor:
M went to zero as there was no stop condition aside from ccw. When M is zero - 1 index is for the first point.

I did not analyze it in detail. It is probably an edge case.

ahdinosaur:
hi @hrgdavor, thanks i appreciate the fix :smiley_cat: 

however i'm not sure this fixes the bug. it now renders, but i don't think this is correct.

the 2d profiles i originally wanted to hull (created using `circle` with `segments = 6`):

Screenshot from 2021-08-04 16-10-04

the resulting hull:

Screenshot from 2021-08-04 16-10-38

the 2d profiles that i am using instead, which work (created using `circle` with default `segments`):

Screenshot from 2021-08-04 16-11-16

the resulting hull:

Screenshot from 2021-08-04 16-11-43

hrgdavor:
It was a quick attempt. Pls send the code so I could investigate further

hrgdavor:
@ahdinosaur pls check now, I have reverted initial fix. But changet treshold to Number.EPSILON and it looks like it is working.

I did not go deep into understanding the code, just tried out few things. It looks like it is not breaking stuff, and the example with segments:6,10 give expected results.

z3dev:
> I did not go deep into understanding the code, just tried out few things. It looks like it is not breaking stuff, and the example with segments:6,10 give expected results.

I don’t understand the calculation of ‘ccw’ which seems to be calculating a difference between distances. If I understand correctly that’s not going to determine the ‘handedness’ of points.

@ahdinosaur do you have any idea what CCW is based on?

z3dev:
I went back and looked at the original JAVA code, which this was based on. Here’s the original code for CCW.

 `
  /**
     * Returns the GrahamScan#Turn formed by traversing through the
     * ordered points <code>a</code>, <code>b</code> and <code>c</code>.
     * More specifically, the cross product <tt>C</tt> between the
     * 3 points (vectors) is calculated:
     *
     * <tt>(b.x-a.x * c.y-a.y) - (b.y-a.y * c.x-a.x)</tt>
     *
     * and if <tt>C</tt> is less than 0, the turn is CLOCKWISE, if
     * <tt>C</tt> is more than 0, the turn is COUNTER_CLOCKWISE, else
     * the three points are COLLINEAR.
     *
     * @param a the starting point.
     * @param b the second point.
     * @param c the end point.
     * @return the GrahamScan#Turn formed by traversing through the
     *         ordered points <code>a</code>, <code>b</code> and
     *         <code>c</code>.
     */
    protected static Turn getTurn(Point a, Point b, Point c) {

        // use longs to guard against int-over/underflow
        long crossProduct = (((long)b.x - a.x) * ((long)c.y - a.y)) -
                            (((long)b.y - a.y) * ((long)c.x - a.x));

        if(crossProduct > 0) {
            return Turn.COUNTER_CLOCKWISE;
        }
        else if(crossProduct < 0) {
            return Turn.CLOCKWISE;
        }
        else {
            return Turn.COLLINEAR;
        }
    }
}
`

hrgdavor:
There is a cool anim in wiki article: 
https://en.wikipedia.org/wiki/Graham_scan

z3dev:
A nice explanation on StackExchange…

In short: It's a shorthand notation for a mathematical hack.

Long explanation:

You can't do a cross product with vectors in 2D space. The operation is not defined there.

However, often it is interesting to evaluate the cross product of two vectors assuming that the 2D vectors are extended to 3D by setting their z-coordinate to zero. This is the same as working with 3D vectors on the xy-plane.

If you extend the vectors that way and calculate the cross product of such an extended vector pair you'll notice that only the z-component has a meaningful value: x and y will always be zero.

That's the reason why the z-component of the result is often simply returned as a scalar. This scalar can for example be used to find the winding of three points in 2D space.

From a pure mathematical point of view the cross product in 2D space does not exist, the scalar version is the hack and a 2D cross product that returns a 2D vector makes no sense at all.

z3dev:
So, after all that research and nice study into the ‘hack’, the fix from @hrgdavor seems sound. Anything greater than Number.MINIMUM should be considered counter clockwise orientation.

hrgdavor:
😄 I smashed the keyboard until it worked, @z3dev  checked if it makes sense :) time to merge

z3dev (approved):
Looks good! Thanks!

Issue #598, opened 2020-06-07
union of 5 simple objects crashes browser


SebiTimeWaster:
## Expected Behavior
that it works

## Actual Behavior
when i try to union the 5 elements the attached below script generated or click on "generate stl" (i presume it does a union in the background) the browser tab is not responding anymore until it crashes

## Steps to Reproduce the Problem

  1. generate 5 objects with attached below script
  2. click "generate stl"
  3. crash

## Specifications

  - Version: website
  - Platform:  newest ubuntu
  - Environment: newest chromium

```
function main() {
    const testObj = union(
        translate([5, 5, 5], sphere({ r: 5, fn: 60 })),
        translate([0, 0, 10], cube({ size: 10 })),
        translate([5, 5, 20], cylinder({ r1: 1, r2: 5, h: 10, fn: 60 }))
    );

    return bendObj(testObj, 20, 5);
}

const bendObj = (obj, objRadius, segments = 1) => {
    const slices = [];

    if (segments === 1) {
        slices.push(bend(obj, objRadius));
    } else {
        const objSize = obj.getBounds();
        const objW = Math.abs(objSize[0].x - objSize[1].x);
        const objD = Math.abs(objSize[0].y - objSize[1].y);
        const objH = Math.abs(objSize[0].z - objSize[1].z);
        const objSliceHeight = objH / segments;

        for (let x = 0; x < segments; x++) {
            const bentObj = bend(
                intersection(
                    translate([0, 0, x * -objSliceHeight], obj),
                    translate([-5, -5, 0], cube({ size: [objW + 10, objD + 10, objSliceHeight] }))
                ),
                objRadius
            );
            const angleAlpha = (x * objSliceHeight) / objRadius;

            slices.push(translate([0, -objRadius, 0], rotate([(angleAlpha * 360) / (Math.PI * 2), 0, 0], translate([0, objRadius, 0], bentObj))));
        }
    }

    return slices;
};

const bend = (obj, objRadius) => {
    const csg = new CSG();

    obj.polygons.forEach((polygon) => {
        const vertices = [];

        polygon.vertices.forEach((vertice) => {
            const vertex = vertice.pos;
            const angleAlpha = vertex.z / objRadius;
            const newRadius = objRadius + vertex.y;
            const posY = Math.cos(angleAlpha) * newRadius - objRadius;
            const posZ = Math.sin(angleAlpha) * newRadius;

[...]

PR #898, opened 2021-08-06
Fix 876 expand v2

hrgdavor:
### All Submissions:

this is a second attempt at fixing #876 and also affects #598
 - bonus fixed a bug that would crash retesellate

#877 was too complicated. instead of stopping right away that version when noticed the stop condition tried then to use other polygons from the list to do the splitting. This was wrong as it looks like it does not matter which one is chosen is stop condition is met, it will be met for all of them for sure. Hence it is enough to check just one of them.

@z3dev you noticed correctly that #877 one was fishy. This one does not affect tests, it just fixes the issue so I am sure it will be much easier to review :)


z3dev:
@hrgdavor Thanks. This looks more interesting.

Is this ready for review? There are still a few debugging statements.

hrgdavor:
@z3dev it is now ready for review. Thnx for noticing those. You may notice I like to rush things :).

When u have time, check my other PR's if you have suggestions. (I know you will not be able to properly test some of them until you get back)

Since I can't nag you on discord, here is small update on some of things in progress

I am working on few things in paralel, so there will be more PR's ... latest effort is 3mf serializer (it is exporting nicely and also embeds preview png )
image
Those marked with red numbers are the 3mf export.

One nice thing aabout 3mf is that it supports instancing (declare geometry once and reference in the final output). 

Also instancing greatly helps performance ... I was able to get 32x32x32 initial render of the model down to 20ms (30 thousand logos)
image
(regl renderer starts to struggle with grid larger than 5x5x5)


I also went through old issues and found metaballs. So I converted them to V2 code 
image

image

z3dev:
I tested a little with these changes. The change to reTesselateCoplanarPolygons() corrects issues, and allows the expansion to complete. But the results of the expansion are still very poor.

It seems that there's something very wrong with the expandShell() function, which does the 3D geometry expansion.

issue876

I think the example given in #876 will help to debug and fix this issue.

hrgdavor:
The fix to reTesselateCoplanarPolygons is just a sidefix so it does not produce invalid results.

The actuall fix aimed by this PR is about the Node.js and adding a stop condition to prevent infinite loop.

The question here is if that stop condition is enough, for this PR., and if expandShell() need another issue and another fix.

If this PR needs further improvement, pls give some direction :)

z3dev:
> The actual fix aimed by this PR is about the Node.js and adding a stop condition to prevent infinite loop.

I found that the fix to reTesselateCoplanarPolygons() was enough, and corrected the infinite loop.

Either way, there are not definite test cases for either of the fixes. So, it needs some more research on the real issue.

hrgdavor:
I have just noticed I named the PR "Fix 876 expand v2" so yes, direction may be ok, but needs more research.
Could you give me an example that is messed put like that one in the picture ?

hrgdavor:
The more I am looking how the expand is implemented, the less I like it. I need your help to differentieate CSG issue here from expand issues.

I would actually like to rename this PR or make another one with current fixes `Fix infinite loop in CSG subdivision and reTesselateCoplanarPolygons empty polygons`. 

I am not liking the approach made in this expand function, the performance implications and also the end result is messy, so I am not sure it is worth fixing.

z3dev:
> I am not liking the approach made in this expand function, the performance implications and also the end result is messy, so I am not sure it is worth fixing.

It's definitely one UGLY function. Although, I must give credit because this was written long long ago when the math was pretty iffy.

I'll try to knock together a nice little test case for the expand(), and try to make it break big time. I'll post back once complete.

There may be some related issue to the boolean functions, as there have been cases of infinite loops in the past. If you can find a test cases for these fixes then that would awesome.

hrgdavor:
I produce ugly code all the time, so I do not have an issue calling other ppl's code ugly :D and I am fine with ugly code that works.

The ugly code that sort of works usually ok, and if it does something that it is really needed, it then leads to more ideas as issues are exposed after actual use/abuse. So I am grateful someone implemented this version of expand, even though it is not perfect, it is mostly usable and allows us to see if it is something worth exploring further and improving.

I will look through issues if there are other examples of booleans hanging,, and ofc, if u find some, let me know.

hrgdavor:
There is another example from #598 that also crashes openjscad.xyz but demostrates that this fix is not complete as it leaves artifacts isntead of crashing.
The stop condition is good to prevent crash, but needs to be investigated to prevent from leaving holes

Here is v2 version of the code
```
const { measureBoundingBox } = jscad.measurements
const { cube, cuboid, sphere, cylinder } = jscad.primitives
const { translate, rotate } = jscad.transforms
const { colorize } = jscad.colors
const { union, subtract, intersect } = jscad.booleans

function main() {
    let out = [
        translate([5, 5, 5], sphere({ radius: 5, segments: 32 })),
        translate([0, 0, 10], cube({ size: 10 })),
        translate([5, 5, 20], cylinder({ startRadius: 1, endRadius: 5, height: 10, segments: 16 }))
    ]
    
    out = union(out);
    out = bendObj(out, 20, 5);
    out = union(out)
    return out
}

const bendObj = (obj, objRadius, segments = 1) => {
    const slices = [];

    if (segments === 1) { 
        slices.push(bend(obj, objRadius));
    } else {
        const objSize = measureBoundingBox(obj);
        const objW = Math.abs(objSize[0][0] - objSize[1][0]);
        const objD = Math.abs(objSize[0][1] - objSize[1][1]);
        const objH = Math.abs(objSize[0][2] - objSize[1][2]);
        const objSliceHeight = objH / segments;

        for (let x = 0; x < segments; x++) {
            const bentObj = bend(
                intersect(
                    translate([0, 0, x * -objSliceHeight], obj),
                    translate([-5, -5, 0], cuboid({ size: [objW + 10, objD + 10, objSliceHeight] }))
                ),
                objRadius
            );
            const angleAlpha = (x * objSliceHeight) / objRadius;

            slices.push(translate([0, -objRadius, 0], rotate([angleAlpha, 0, 0], translate([0, objRadius, 0], bentObj))));
        }
    }

    return slices;
};

const bend = (obj, objRadius) => { 
    const csg = jscad.geometries.geom3.create();

    obj.polygons.forEach((polygon) => {
        const vertices = [];

        polygon.vertices.forEach((vertex) => {
            const angleAlpha = vertex[2] / objRadius;
            const newRadius = objRadius + vertex[1];
            const posY = Math.cos(angleAlpha) * newRadius - objRadius;
            const posZ = Math.sin(angleAlpha) * newRadius;

            vertices.push([vertex[0], posY, posZ]);
        });

        csg.polygons.push({vertices});
    });

    csg.isCanonicalized = false;
    csg.isRetesselated = false;
    return csg;
};

module.exports = {main}
```

hrgdavor:
Original object after union
image

The object after bend function (has holes from the function, that are not fault of booleans):
image

The object after union, with 2 extra artefacts
image

hrgdavor:
After debugging the code and dumping the polygons that are part of the stop condition that was added, it is clear at least for this exact case that they are not the part of the issue, as all 3 `1,2,3` instances are somewhere else where object seems to be ok. The artefacts from union are `4,5`

image

hrgdavor:
disabling retesselate fixes one artefact.
image

hrgdavor:
@z3dev could you please review again. it looks like without changing this PR issues are solved by using generalize after the user function that does bending and before union.

```js
out = scad.modifiers.generalize({snap:true,simplify:false}, out)
```

z3dev:
@hrgdavor these changes are really good. i especially like the changes to the Node.js code as there have been lots of issues with infinite loops.

however, a simple test case would be good to have. let's continue to search for one.

hrgdavor:
Not sure what kind of test case is needed here, as it is bit hard to make test to force infinite loop.
The fix seems to improve the booleans to prevent at least this case that causes infinite loop.

I does not fix #907 but also does not make it worse.

It would be good to merge this and continue working out the cause of those artifacts in #907 in a separate PR

hrgdavor:
I have added the test that makes expand go into infinite loop (Node.js actually)

it failed on master branch 
```
@jscad/modeling:   × Timed out while running tests
@jscad/modeling:   1 tests were pending in src\operations\expansions\expandGeom3.test.js
@jscad/modeling:   ( ) operations » expansions » expandGeom3 » expand survives bug 876
```

but works on this PR

@z3dev pls review

packages/modeling/src/operations/expansions/expandGeom3.test.js
```diff
@@ -0,0 +1,32 @@
+const test = require('ava')
+
+const { geom3, poly3 } = require('../../geometries')
+const { expand } = require('./index')
```
z3dev, line 4: instead of using 'expand', let's test with expandGeom3

`const expandGeom3 = require('./expandGeom3')`

packages/modeling/src/operations/expansions/expandGeom3.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+test('expand survives bug 876', async (t) => {
```
z3dev, line 6: title change... 'expandGeom3: expand completes properly, issue 876'

z3dev (changes requested):
OK. I'm fine with this test.

Just a few small changes.

hrgdavor:
Fixed latest comments, @z3dev pls review :)

z3dev (approved):
Super. Thanks for the last set of changes!

PR #902, opened 2021-09-02
fix(web): enhanced createContext() for older browsers

z3dev:
Upon testing the latest changes to master, an issue was found during testing Safari 12.0.2 (old version). The call to canvas.getContext() was returning 'null' rather then throwing an exception. This created an unusable canvas element.

https://en.wikipedia.org/wiki/Safari_(software)#Safari_15

These changes check the result of getContext(), which should allow older versions of browsers to detect the context correctly.

These changes have been tested with:
- Safari 12.0.2 Mac OS 10.14.2 and Safari 14.1.2 on Mac OS 10.14.6 (latest)
- Firefox 91.0.2
- Chrome 92.0.4515
- Explorer 92.0.902

### All Submissions:

### DIFF
--- a/packages/web/src/ui/views/viewer.js
+++ b/packages/web/src/ui/views/viewer.js
@@ -195,7 +195,9 @@ const viewer = (state, i18n) => {
 const createContext = (canvas, contextAttributes) => {
   const get = (type) => {
     try {
-      return { gl: canvas.getContext(type, contextAttributes), type }
+      // NOTE: older browsers may return null from getContext()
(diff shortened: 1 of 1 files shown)


z3dev:
@hrgdavor can you review these changes as well? thanks ahead of time!

hrgdavor (approved):
changes look good. Unfortunately, I only can tell it works on my browsers. I do not have safari :)

mister-jscad (approved):
Looks fine.

PR #913, opened 2021-09-19
Fix 912 concat and appendArc

### LINKED ISSUE #912: `path2.concat()` of mirrored pair of paths followed by `extrudeRectangular` results in mesh with gaps
## Expected Behavior
Mirroring a path and concatenating it would result in a clean, deduplicated path ready to be extruded, with the same output as creating the path without mirroring (the output of which is shown on the right in the image below).

left: broken mesh version of chip clip model, right: resolved version

## Actual Behavior
The resulting mesh has gaps in it when attempting to mirror a path and then concatenate the mirrored and unmirrored versions together.

The source code of both versions shown in the image can be found in [this CadHub project](https://cadhub.xyz/u/franknoirot/chip-clip-jscad/ide).
  1. Create a path
  2. Concatenate the path with a mirrored version of itself
  3. Extrude the path with `extrudeRectangular()`

## Specifications
  - Version: latest
  - Platform: CadHub
  - Environment: browser

hrgdavor:
fixes #912 
there is duplicate point at the joint, but not 100% equal `[1, -3]` vs `[0.9999999999999978, -3.0000000000000018]`

First issue is that path2.concat does not remove duplicate point when adding paths

Second issue is that appendArc does not produce the exact same endpoint as user asked for.

On route is to fix equality check using epsilon, and other is to force appendArc to respect the desired endpoint.

This PR  forces appendArc to respect the desired endpoint.

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/path2/appendArc.js
+++ b/packages/modeling/src/geometries/path2/appendArc.js
@@ -121,7 +121,7 @@ const appendArc = (options, geometry) => {
     // Ok, we have the center point and angle range (from theta1, deltatheta radians) so we can create the ellipse
     let numsteps = Math.ceil(Math.abs(deltatheta) / (2 * Math.PI) * segments) + 1
     if (numsteps < 1) numsteps = 1
-    for (let step = 1; step <= numsteps; step++) {
+    for (let step = 1; step < numsteps; step++) {
       const theta = theta1 + step / numsteps * deltatheta
       const costheta = Math.cos(theta)
       const sintheta = Math.sin(theta)
@@ -130,6 +130,8 @@ const appendArc = (options, geometry) => {
       vec2.add(point, point, center)
       newpoints.push(point)
     }
+    // ensure end point is precisely what user gave as parameter
+    if (numsteps) newpoints.push(options.endpoint)
   }
   newpoints = points.concat(newpoints)
   const result = fromPoints({}, newpoints)
--- a/packages/modeling/src/geometries/path2/concat.js
+++ b/packages/modeling/src/geometries/path2/concat.js
@@ -26,7 +26,7 @@ const concat = (...paths) => {
   let newpoints = []
   paths.forEach((path) => {
     const tmp = toPoints(path)
-    if (newpoints.length > 0 && equals(tmp[0], newpoints[newpoints.length - 1])) tmp.shift()
+    if (newpoints.length > 0 && tmp.length > 0 && equals(tmp[0], newpoints[newpoints.length - 1])) tmp.shift()
     newpoints = newpoints.concat(tmp)
   })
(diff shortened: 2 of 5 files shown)


hrgdavor:
image

packages/modeling/src/geometries/path2/appendArc.js
```diff
@@ -130,6 +130,8 @@ const appendArc = (options, geometry) => {
       vec2.add(point, point, center)
       newpoints.push(point)
     }
+    // ensure end point is precisely what user gave as parameter
+    if (numsteps) newpoints.push(options.endpoint)
```
z3dev, line 134: Nice!

packages/modeling/src/geometries/path2/concat.js
```diff
@@ -1,8 +1,9 @@
 const fromPoints = require('./fromPoints')
```
z3dev, line 1: For consistent style, let's add this require at the top.

`const vec2 = require('../../maths/vec2')`

packages/modeling/src/geometries/path2/concat.js
```diff
@@ -1,8 +1,9 @@
...
 const toPoints = require('./toPoints')
-
+const equals = require('../../maths/vec2/equals')
 /**
  * Concatenate the given paths.
+ * If both contain the same point at the junction, merge it into one.
```
z3dev, line 6: I think this is a comment for the code change, not for the API documentation.

packages/modeling/src/geometries/path2/concat.js
```diff
@@ -24,7 +25,9 @@ const concat = (...paths) => {
   }
   let newpoints = []
   paths.forEach((path) => {
-    newpoints = newpoints.concat(toPoints(path))
+    const tmp = toPoints(path)
+    if (newpoints.length > 0 && equals(tmp[0], newpoints[newpoints.length - 1])) tmp.shift()
```
z3dev, line 29: It's possible to concat paths that are empty, i.e. no points. So an addition check is required. Maybe the new test case should include an empty path.

z3dev (changes requested):
Small changes requested, but looks good. A test case for the concat() change would be nice as well.

packages/modeling/src/geometries/path2/concat.js
hrgdavor: well, it is both :) ... may be good to leave it so end user knows the sideeffects of the concat.

packages/modeling/src/geometries/path2/concat.js
hrgdavor:
```js
const { equals } = require('../../maths/vec2')
```

z3dev (approved):
Looks good! Thanks.

PR #918, opened 2021-09-28
fix(modeling): corrected vec2.rotate, and enhanced test cases

z3dev:
I found this little bug while looking at some examples.

In addition, clamp and quantitzeForSpace were dropped from math/utils as these functions are not used / supported.

### All Submissions:


hrgdavor (approved):
looks ok :)

PR #925, opened 2021-10-02
Fix DXF serialization for near-black colors

platypii:
I was exporting my jscad designs with color, and everything was working except the color black (also near-black colors since dxf is paletted)

For example:

```
const { colorize, cssColors } = require('@jscad/modeling').colors
const { cuboid } = require('@jscad/modeling').primitives

function main() {
  return colorize([0, 0, 0], cuboid({}))
}

module.exports = { main }
```

Load the js directly into jscad and the color works:

black

But if you export it to a DXF file and then open that DXF file in jscad you see:

blue

Investigation showed that `getColorNumber` in the [serializer](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/io/dxf-serializer/index.js#L465) was searching from 0 to 255 through the color index. However the [deserializer](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/io/dxf-deserializer/helpers.js#L54) requires that the index be at least 1.

It appears that colorindex 0 was just added as a convenience but was not intended to be used. colorindex 250 is also black [0,0,0] and works when deserialized too. So I just changed the serializer here.

### All Submissions:

### DIFF
--- a/packages/io/dxf-serializer/index.js
+++ b/packages/io/dxf-serializer/index.js
@@ -462,7 +462,7 @@ const getColorNumber = (object, options) => {
     // find the closest Autocad color number
     const index = options.colorIndex
     let closest = 255 + 255 + 255
-    for (let i = 0; i < index.length; i++) {
+    for (let i = 1; i < index.length; i++) {
       const rgb = index[i]
       const diff = Math.abs(r - rgb[0]) + Math.abs(g - rgb[1]) + Math.abs(b - rgb[2])
       if (diff < closest) {


z3dev:
@platypii Cool! Thanks for the changes.

Quick question... Given this change, is a more appropriate color index chosen? Did the resulting DXF file have the correct (near) color?

platypii:
Yes this change results in black colors rendering correctly in the DXF file. It choose palette color 250 which is also black.

platypii:
It would also be possible to make the deserializer more tolerant by changing the 1 to a zero on this line:
https://github.com/jscad/OpenJSCAD.org/blob/master/packages/io/dxf-deserializer/helpers.js#L54

But I don't know if that's actually dxf format compatible when color index = 0

z3dev:
> It would also be possible to make the deserializer more tolerant by changing the 1 to a zero on this line: https://github.com/jscad/OpenJSCAD.org/blob/master/packages/io/dxf-deserializer/helpers.js#L54
> 
> But I don't know if that's actually dxf format compatible when color index = 0

If I remember correctly, the zero index is reserved for the block color, or something like that.

So, wouldn’t the ‘default’ color of the viewer be more appropriate? (current case)

platypii:
> If I remember correctly, the zero index is reserved for the block color, or something like that.
> So, wouldn’t the ‘default’ color of the viewer be more appropriate? (current case)

Then yes I think that letting zero map to the default color in the deserializer makes sense, and the serializer change in this PR is the right move to fix the color black.

z3dev (approved):
Thanks for finding this little bug, and fixing it!

PR #927, opened 2021-10-04
fix(modeling): change geometry transforms to preserve user attributes

z3dev:
This is another try at retaining user attributes after transforming geometries. The changes are based on several attempts (See other PRs) and analysis below.

The changes are minimal, and all tests pass without changes. Except the tests for colorize() which was changed.

### All Submissions:

### DIFF
--- a/packages/modeling/src/colors/colorize.js
+++ b/packages/modeling/src/colors/colorize.js
@@ -52,10 +52,10 @@ const colorize = (color, ...objects) => {
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
   const results = objects.map((object) => {
-    if (geom2.isA(object)) return colorGeom2(color, object)
-    if (geom3.isA(object)) return colorGeom3(color, object)
-    if (path2.isA(object)) return colorPath2(color, object)
-    if (poly3.isA(object)) return colorPoly3(color, object)
+    // if (geom2.isA(object)) return colorGeom2(color, object)
+    // if (geom3.isA(object)) return colorGeom3(color, object)
+    // if (path2.isA(object)) return colorPath2(color, object)
+    // if (poly3.isA(object)) return colorPoly3(color, object)
 
     object.color = color
     return object
--- a/packages/modeling/src/colors/colorize.test.js
+++ b/packages/modeling/src/colors/colorize.test.js
@@ -43,23 +43,15 @@ test('color (rgba on geometry)', (t) => {
   const obs = colorize([1, 1, 0.5, 0.8], obj0, obj1, obj2, obj3)
   t.is(obs.length, 4)
 
-  let exp = geom2.clone(obj0)
-  exp.color = [1, 1, 0.5, 0.8]
-  t.not(obj0, obs[0])
-  t.deepEqual(obs[0], exp)
+  t.is(obj0, obs[0])
+  t.deepEqual(obs[0].color, [1, 1, 0.5, 0.8])
 
-  exp = geom3.clone(obj1)
-  exp.color = [1, 1, 0.5, 0.8]
-  t.not(obj1, obs[1])
-  t.deepEqual(obs[1], exp)
+  t.is(obj1, obs[1])
+  t.deepEqual(obs[1].color, [1, 1, 0.5, 0.8])
 
-  exp = path2.clone(obj2)
-  exp.color = [1, 1, 0.5, 0.8]
-  t.not(obj2, obs[2])
-  t.deepEqual(obs[2], exp)
+  t.is(obj2, obs[2])
+  t.deepEqual(obs[2].color, [1, 1, 0.5, 0.8])
 
-  exp = poly3.clone(obj3)
-  exp.color = [1, 1, 0.5, 0.8]
-  t.not(obj3, obs[3])
-  t.deepEqual(obs[3], exp)
+  t.is(obj3, obs[3])
+  t.deepEqual(obs[3].color, [1, 1, 0.5, 0.8])
 })
--- a/packages/modeling/src/geometries/geom2/applyTransforms.js
+++ b/packages/modeling/src/geometries/geom2/applyTransforms.js
@@ -19,7 +19,7 @@ const applyTransforms = (geometry) => {
     const p1 = vec2.transform(vec2.create(), side[1], geometry.transforms)
     return [p0, p1]
   })
-  mat4.identity(geometry.transforms)
+  geometry.transforms = mat4.create()
   return geometry
 }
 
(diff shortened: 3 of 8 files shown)


z3dev:
### FIX TRANSFORMS

const B = clone(A)
- creates NEW sides by making deep copies
- creates NEW transforms by making copy
- NEW { sides: NEW, transforms: NEW }
- LOOSES extra attributes

const C = reverse(A)
- calls toSides()
- creates NEW sides by reversing
- creates NEW transforms from identity
- NEW { sides: NEW, transforms: NEW }
- LOOSES extra attributes

const sides = toSides(A)
- calls applyTransforms(A)

const A = applyTransforms(A)
- creates NEW sides by applying transforms
- updates OLD transforms
- OLD { sides: NEW, transforms: OLD } #1 CHANGE
- RETAINS extra attributes

const E = transform(matrix, A)
- calculates NEW transforms
- NEW { sides: A.sides, transforms: NEW } #2 CHANGE
- LOOSES extra attributes

CHANGES:
#1	create NEW matrix to prevent improper changes to shallow copies
#2	create NEW object using Object.assign(…) to RETAIN extra attributes

QUESTIONS:
- Should clone() also retain extra attributes? Probably
- Should reverse() also retain extra attributes? Hmm… it’s a different geometry… but maybe

hrgdavor:
I think that even applyTransforms should create a copy. The only thing reused there is original object with few attributes, but data is cloned-transformed. Immutability has a lots of benefits, and mutating here brings nothing useful and does not improve performance.

If I run script multiple times and cache some geometries for performance, this type of mutation could break the code. I know right now jscad web kills worker every time, but I hope to change this in near future, and by doing that allow more advanced usages with caching between parameter changes.

z3dev:
> If I run script multiple times and cache some geometries for performance, this type of mutation could break the code. I know right now jscad web kills worker every time, but I hope to change this in near future, and by doing that allow more advanced usages with caching between parameter changes.

You missed the point of toSides() and applyTransforms(). These are used to obtain the final sides of the geometry which is required when exporting to another format, or performing booleans, or extruding, etc.

The rendering of the geometry should not need the final sides, and just use the provided matrix. That’s first change was completed, but the last change to disable the serialize is not.

And these changes will allow for shallow copies as well. But there’s no use case for that currently.

hrgdavor:
> The rendering of the geometry should not need the final sides, and just use the provided matrix. That’s first change was completed, but the last change to disable the serialize is not.

I understand where it is used :) and aside from serialize it is also used for booleans. There is not real benefit of this mutating :) .

If you prefer I can make a separate PR for applyTransform after this one. I actually use this in my prototype, I need this immutability.

z3dev:
> If you prefer I can make a separate PR for applyTransform after this one. I actually use this in my prototype, I need this immutability.

Well… you did make a PR for this, and the internals became a real mess.
Did you have some additional thoughts on how to keep the geometry?

Also, the goal of this PR is not caching geometry. The goal is allowing user attributes, like color.

hrgdavor:
> Well… you did make a PR for this, and the internals became a real mess.

it was before render change to allow for matrix, and I  caught isues regarding the toSides and other strange stuff and it got complicated.

After you merge this PR I am confident I can get a cleaner version (maybe split in few steps/PRs). 

This mutation stuff has to go away, it brings nothing useful. 

And it can be done cleanly, I just need your feedback while wo I do it.



But now that few important things have already changed the #866 can be closed and the redone in another PR/s.

z3dev:
> > Well… you did make a PR for this, and the internals became a real mess.
> 
> it was before render change to allow for matrix, and I caught isues regarding the toSides and other strange stuff and it got complicated.
> 
> After you merge this PR I am confident I can get a cleaner version (maybe split in few steps/PRs).

OK. The last change to regl-renderer should help move this forward even more.

@hrgdavor can you change the core again, and disable the serialze of geometry? Probably, just changing the default is enough.

that should allow the changes in this PR to have more impact as well.

then, some more thoughts / changes for improvements can be made.

hrgdavor:
Ok, I will place a PR to serialize:false for jacad web

hrgdavor:
@z3dev there is new PR #928 . Please review, it is a simple change. then we can move forward with both, and I will see about the further improvements towards immutability.

packages/modeling/src/colors/colorize.js
```diff
@@ -52,10 +52,10 @@ const colorize = (color, ...objects) => {
...
-    if (geom2.isA(object)) return colorGeom2(color, object)
-    if (geom3.isA(object)) return colorGeom3(color, object)
-    if (path2.isA(object)) return colorPath2(color, object)
-    if (poly3.isA(object)) return colorPoly3(color, object)
+    // if (geom2.isA(object)) return colorGeom2(color, object)
+    // if (geom3.isA(object)) return colorGeom3(color, object)
+    // if (path2.isA(object)) return colorPath2(color, object)
+    // if (poly3.isA(object)) return colorPoly3(color, object)
 
     object.color = color
```
hrgdavor, line 60: colorize should also return a shallow copy with new color applied. It is relatively ok to mutate for the color change as other operations do shallow copy, but then no other method should do mutation.

I would be ok this staying this way for this PR, but will likely push for change in a future PR to have consistency of immutability.

packages/modeling/src/colors/colorize.js
```diff
@@ -52,10 +52,10 @@ const colorize = (color, ...objects) => {
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
   const results = objects.map((object) => {
-    if (geom2.isA(object)) return colorGeom2(color, object)
-    if (geom3.isA(object)) return colorGeom3(color, object)
-    if (path2.isA(object)) return colorPath2(color, object)
-    if (poly3.isA(object)) return colorPoly3(color, object)
```
hrgdavor, line 58: colorGeom2 colorGeom3 ... are not needed any more

hrgdavor (changes requested):
there are few obsolete function leftover, and an optional change proposal.

platypii:
@z3dev I considered changes very similar to this, but didn't understand the internals well enough to know the consequences. That's why I made #924 the way that I did. Happy to close that in favor of this proper fix.

Is there a standard for what is immutable and what is not? Does every operation (transforms, booleans) produce a totally independent object? eg- any change to the original object will not be reflected in the output object? It seems like that is not quite the case today.

z3dev:
> Is there a standard for what is immutable and what is not? Does every operation (transforms, booleans) produce a totally independent object? eg- any change to the original object will not be reflected in the output object?

That's correct. The geometries are considered immutable, i.e. changing the contents is not permitted. In addition, actions performed on a geometry always produce a NEW geometry.

packages/modeling/src/colors/colorize.js
z3dev: Thanks for the review.

I'm hoping that geometry.clone() can become a shallow copy now, and will test this change. If a shallow clone() is workable then I'll update colorize() to use clone() again.

z3dev:
@hrgdavor please review the latest changes. Thanks.

PR #928, opened 2021-10-04
change default serialize to false

hrgdavor:
### All Submissions:

CLI already has serialize:false and now that web renderer has support for matrix transform same default can be used for web.

### DIFF
--- a/packages/core/src/code-evaluation/rebuildGeometry.js
+++ b/packages/core/src/code-evaluation/rebuildGeometry.js
@@ -29,7 +29,7 @@ const rebuildSolids = (data, callback) => {
   const defaults = {
     mainPath: '',
     vtreeMode: false,
-    serialize: true,
+    serialize: false,
     lookup: null,
     lookupCounts: null,
     parameterValues: {}


z3dev:
@hrgdavor this change if fine, and i verified that the serialize of geometries no longer takes place. 👍 

however, i found an issue with some rendering, which occurs when the geometry is scaled. here's a simple example.

```
const { cube, sphere, circle, square, arc } = require('@jscad/modeling').primitives
const { scale, translate } = require('@jscad/modeling').transforms
const { colorize } = require('@jscad/modeling').colors

const main = () => {
  let obs = cube()
  obs = scale([10, 10, 10], obs)
  return obs
}
module.exports = {main}
```
Maybe a small change to the renderer is required again.
Also see the example called "Spheres of all sorts" which also does some scaling of geometries.

hrgdavor:
I am not seing issues with the cube or spheres (firefox and Chrome)

image

pls screenshot you issues :) as I am unable to replicate

z3dev:
@hrgdavor i'm not sure what's happening. here are the screen shot from firefox, chrome, and safari.
render_3
render_2
render_1

Issue #940, opened 2021-10-24
Missing faces in exported files


platypii:
## Expected Behavior

Objects should be solid.

## Actual Behavior

There is a bug that results in missing faces on objects when exported into any format. I run into this pretty frequently when making real world designs. This causes problems in 3D printing software.

It's interesting that if you just drag the .js file onto jscad, it renders correctly. If you export as any file type (stl, json, dxf, etc) they all have missing faces.

## Steps to Reproduce the Problem

It's fairly easy to find examples of missing faces if you start doing boolean operations with cylinders. For example, modeling screw holes. Cylinders must have a fairly high number of segments, and helps to be off-center.

```
const { subtract } = require("@jscad/modeling").booleans
const { cylinder } = require("@jscad/modeling").primitives

function main() {
  return subtract(
    cylinder({radius: 20, center: [1, 1, 0], segments: 200}),
    cylinder({radius: 11, center: [1, 1, 0], segments: 200})
  )
}

module.exports = { main }
```

  1. Open .js file
![jscad-error1](https://user-images.githubusercontent.com/1766297/138584312-7f31bc78-e5dc-4bd8-bdf4-5cc7386a5c43.png)

  2. Export as STL
  3. Open .stl file
![jscad-error2](https://user-images.githubusercontent.com/1766297/138584318-6c4e479f-141b-448e-99c1-76da6dfa7666.png)

## Specifications

  - Versions: 2.0.0 ... 2.4.4
  - Platform: Web, CLI
  - Environment: chrome and node.js

platypii:
Fixed by #941

z3dev:
@platypii sorry but I’d like to reopen until the root cause is identified.

platypii:
I can tell you exactly the root cause. The snapping causes a polygon to become non-convex. Specifically there is a pentagon which is convex before the snapping. After snapping, the first three points in the vertices become "inverted". It's important that the _first_ three points become inverted, because only the first three points are used to compute a face's "plane" [[1]](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/modeling/src/geometries/poly3/plane.js#L6)

This results in the plane facing inwards instead of outwards.

Before snapping:
![face1](https://user-images.githubusercontent.com/1766297/140787143-9ce1c292-39d1-4288-a881-4ad11a6e3f08.png)

After snapping:
[...]

PR #941, opened 2021-10-25
fix(modeling): handle negative area polygons in snapPolygons

platypii:
This is a fix for: #940 missing faces in exported files

Polygons can be inside out. Whether by design, or as a result of snapping to grid. However, in the existing code they are filtered out because their size is "less than epsilon". This PR makes it so that polygons are only removed if the _magnitude_ of the area is less than epsilon, regardless of orientation.

### All Submissions:


z3dev (approved):
@platypii Super nice. Thanks

Issue #515, opened 2020-02-10
OBJ: create serializer


z3dev:
It would be great to have the symetric capbilities for OBJ file formats, both serializer and deserialize.

See http://en.wikipedia.org/wiki/Wavefront_.obj_file

z3dev:
@platypii you will never believe this but i stumbled upon a little bug in the new serializer. try this design.

```
const { cylinder, cuboid } = require('@jscad/modeling').primitives
const { union } = require('@jscad/modeling').booleans
const { colorize } = require('@jscad/modeling').colors
const { rotateX, translate } = require('@jscad/modeling').transforms

// Issue #594 - Colors
// - Components will create and return parts, often with color, e.g. gold bolt
// - Even after aligning the part, the color should be retained

const createBolt = () => {
  let bolt = union(
    cylinder({height: 2, center: [0, 0, 1], radius: 2, segments: 8}), // head
    cylinder({height: 2 + 4, center: [0, 0, 3]})   // shaft
  )
  bolt.name = "2 x 4 mm bolt"
  return bolt
}

const main = () => {
  let bolt = rotateX(Math.PI, createBolt())

  // create four bolts for each corner
  const bolt1 = translate([+3, +3, 4], bolt)
  const bolt2 = translate([+3, -3, 4], colorize([1, 0, 0], bolt))
  const bolt3 = translate([-3, +3, 4], bolt)
  const bolt4 = colorize([0, 0, 1], translate([-3, -3, 4], bolt))

  const base = cuboid({size: [12, 12, 2]})

  return [base, bolt1, bolt2, bolt3, bolt4]
}

module.exports = { main }
```

You will notice that there are two red bolts. But the design only has one red bolt.
The issue comes because the 'usemtl' does not get reset after serializing the red bolt. So the material is getting carried forward.

A simple fix is to set 'usemtl nothing' for those shapes without a color. But there may be other solutions.

z3dev:
@platypii thanks again for this nice new feature. :)

PR #947, opened 2021-11-01
fix(obj-serializer) handling of default colors on objects

platypii:
Fix handling of default material color in the new obj-serializer module. Bug found by @z3dev in #515 thanks.

I found some applications use `usemtl default` for the "default" material, so I went with that. Open to suggestions though.

Includes a test which fails without this change.

### All Submissions:


z3dev (approved):
@platypii Thanks again for these fixes!

Issue #951, opened 2021-11-14
Faces are lost during generalize({snap: true}, ...)


jue89:
## Expected Behavior

Faces stay in the model.

## Actual Behavior

1. `require('@jscad/stl-serialize').serialuze` makes use of `modifiers.generalize({snap: true}, ...)`.
2. `modifiers.generalize({snap: true}, ...)` makes use of `poly3.measureArea()`.
3. The following geom3 returns an area of `NaN`:
   ```
   {
     vertices: [
       [ 24.499800000000008, 47.50044000000001, 30.000360000000008 ],
       [ 24.500580000000006, 47.50044000000001, 30.000360000000008 ],
       [ 24.988080000000007, 47.50044000000001, 30.000360000000008 ],
       [ 24.988080000000007, 55.04772000000001, 30.000360000000008 ],
       [ 24.499800000000008, 55.000140000000016, 30.000360000000008 ]
     ]
   }
   ```

The problem is that the normal vector of the first 3 vertices is `[0, 0, 0]`.
https://github.com/jscad/OpenJSCAD.org/blob/fcaff870fcd63d3d6047af122f1e79d1d4774aea/packages/modeling/src/geometries/poly3/measureArea.js#L18-L23

I'm not sure if this is the root cause or if straight lines (in this example vertices[0..2] are a straight line) shouldn't occur in the first place.

Currently I'm working around this problem by calling `generalize({triangulate: true}, ...)` on my model before returning it.

## Steps to Reproduce the Problem

```
const {poly3} = require('@jscad/modeling').geometries;

console.log(poly3.measureArea(poly3.fromPoints([
  [ 24.499800000000008, 47.50044000000001, 30.000360000000008 ],
  [ 24.500580000000006, 47.50044000000001, 30.000360000000008 ],
  [ 24.988080000000007, 47.50044000000001, 30.000360000000008 ],
  [ 24.988080000000007, 55.04772000000001, 30.000360000000008 ],
  [ 24.499800000000008, 55.000140000000016, 30.000360000000008 ]
])));
```

## Specifications

  - Version: @jscad/modeling@2.6.0
  - Platform: MacOS, Linux
  - Environment: nodejs

platypii:
Worth mentioning that the serializers already triangulate before exporting [[1]](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/io/stl-serializer/index.js#L63). So I'm not sure why you need to explicitly call generalize with triangulate yourself?

[...]

PR #952, opened 2021-11-14
fix(modeling): Compute face normals from average of all vertex normals

platypii:
Fix for #951 "faces are lost during generalize({snap: true}, ...)"

There is a bug in `poly3.measureArea` which returns `NaN` in the case where the first three vertices are colinear. This can be demonstrated with:

```js
console.log("colinear area", poly3.measureArea(poly3.fromPoints([
  [ 0, 0, 0 ],
  [ 1, 0, 0 ],
  [ 2, 0, 0 ],
  [ 0, 1, 0 ],
])))
```

```js
colinear area NaN
```

The problem is that jscad takes the first three vertices and computes the vertex normal for only `vertices[1]`. This is not always defined though (for example when colinear). In general, a polygon may not have a defined normal vector. However it's fairly common to use the average of the vertex normals as the face normal.

With the changes in this PR, the above code returns the expected value:

```js
colinear area 1
```

Considerations: this is more computationally expensive. It computes the normal for all vertices instead of just one. It also has to normalize each normal vector for the math to work out. I made a special case for faster triangles, which are coplanar by definition.

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/poly3/faceNormal.js
+++ b/packages/modeling/src/geometries/poly3/faceNormal.js
@@ -1,0 +1,39 @@
+const vec3 = require('../../maths/vec3')
+
+/**
+ * Calculate normal vector for a single vertex.
+ */
+const vertexNormal = (vertices, index) => {
+  const a = vertices[index]
+  const b = vertices[(index + 1) % vertices.length]
+  const c = vertices[(index + 2) % vertices.length]
+  const ba = vec3.subtract(vec3.create(), b, a)
+  const ca = vec3.subtract(vec3.create(), c, a)
+  vec3.cross(ba, ba, ca)
+  vec3.normalize(ba, ba)
+  return ba
+}
+
+/**
+ * Calculate normal vector for a face.
+ * The normal vector is not always well defined.
+ * A polygon may be non-planar. Vertices may be colinear.
+ * To handle these cases we compute the average of the vertex normals.
+ */
+const faceNormal = (vertices) => {
+  if (vertices.length === 3) {
+    // optimization for triangles, which are always coplanar
+    return vertexNormal(vertices, 1)
+  } else {
+    // average of vertex normals
+    const normal = vec3.create()
+    vertices.forEach((v, i) => {
+      vec3.add(normal, normal, vertexNormal(vertices, i))
+    })
+    // renormalize normal vector
+    vec3.normalize(normal, normal)
+    return normal
+  }
+}
+
+module.exports = faceNormal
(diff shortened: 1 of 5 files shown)


packages/modeling/src/geometries/poly3/faceNormal.js
```diff
@@ -0,0 +1,39 @@
...
+    // optimization for triangles, which are always coplanar
+    return vertexNormal(vertices, 1)
+  } else {
+    // average of vertex normals
+    const normal = vec3.create()
+    vertices.forEach((v, i) => {
+      vec3.add(normal, normal, vertexNormal(vertices, i))
+    })
+    // renormalize normal vector
+    vec3.normalize(normal, normal)
```
z3dev, line 34: The comment says average of normals, but this is not calculating an average.

packages/modeling/src/geometries/poly3/faceNormal.js
```diff
@@ -0,0 +1,39 @@
+const vec3 = require('../../maths/vec3')
+
+/**
+ * Calculate normal vector for a single vertex.
+ */
+const vertexNormal = (vertices, index) => {
```
z3dev, line 6: The big performance hit will be the number of vec3.create() calls.

if this function was embedded in the main function then temporarily storage can be allocated for ba and ca, and reused for each call to vertexNormal().

packages/modeling/src/geometries/poly3/faceNormal.js
platypii: I would argue it is the average. It takes the sum and renormalizes. But I updated the comment to be clearer "sum of normals".

packages/modeling/src/geometries/poly3/faceNormal.js
platypii: Inlined to avoid allocations :heavy_check_mark:

packages/modeling/src/geometries/poly3/measureArea.js
```diff
@@ -14,19 +15,18 @@ const measureArea = (poly3) => {
   }
   const vertices = poly3.vertices
 
-  // calculate a real normal
-  const a = vertices[0]
```
z3dev, line 18: There was a reason for calculating the “real” normal, but I didn’t comment well.
Do all tests pass? Have you tried some exports to STL?

platypii: All tests pass. Export works. In most cases, the vertex normal and the face normal will be the same (+/- epsilon). But I don't see how to fix the colinear case without something like this change.

platypii:
@z3dev saw your comment on [#951](https://github.com/jscad/OpenJSCAD.org/issues/951#issuecomment-969438884). I agree the same problem exists in `plane` calculation. Maybe this `faceNormal` should actually be moved to `plane.fromPoints`? That might also fix #940.

platypii:
Closing in favor of #953. This is only a partial fix for measureArea. #953 addresses the problem with planes in general.

PR #953, opened 2021-11-16
fix(modeling): compute planes from average of all vertex normals

platypii:
I made a new PR so that the change to `plane.fromPoints` could be considered separately from #952. In my opinion #952 should be closed in favor of this PR, but I will leave both open for discussion.

This is a fix for #951 "faces are lost during generalize({snap: true}, ...)".

There is a bug in `plane.fromPoints` which returns an undefined plane in the case where the first three vertices are co-linear. This can be demonstrated with:

```js
const { plane } = require("@jscad/modeling").maths
console.log("plane", plane.fromPoints(plane.create(),
  [ 0, 0, 0 ],
  [ 1, 0, 0 ],
  [ 2, 0, 0 ],
  [ 0, 1, 0 ],
))
```

```js
plane [0, 0, 0, 0]
```

The problem is that jscad takes the first three vertices and computes the vertex normal for only `vertices[1]`. This is not always defined though (for example when co-linear). In general, a polygon may not have a defined normal vector. However it's fairly common to use the average of the vertex normals as the face normal.

With the changes in this PR, the above code returns the expected value:

```js
plane [0, 0, 1, 0]
```

Considerations: this is more "correct" but also computationally expensive. It computes the normal for all vertices instead of just one. It also has to normalize each normal vector for the math to work out. I made a special case for faster triangles, which are co-planar by definition.

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/poly3/measureArea.js
+++ b/packages/modeling/src/geometries/poly3/measureArea.js
@@ -1,5 +1,5 @@
 const vec3 = require('../../maths/vec3')
-const plane = require('../../maths/plane')
+const plane = require('./plane')
 
 /**
  * Measure the area of the given polygon.
@@ -16,7 +16,7 @@ const measureArea = (poly3) => {
   const vertices = poly3.vertices
 
   // calculate a normal vector
-  const normal = plane.fromPoints(plane.create(), ...vertices)
+  const normal = plane(poly3)
 
   // determine direction of projection
   const ax = Math.abs(normal[0])
--- a/packages/modeling/src/geometries/poly3/measureArea.js
+++ b/packages/modeling/src/geometries/poly3/measureArea.js
@@ -1,4 +1,5 @@
 const vec3 = require('../../maths/vec3')
+const plane = require('../../maths/plane')
 
 /**
  * Measure the area of the given polygon.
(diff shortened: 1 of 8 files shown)


packages/modeling/src/geometries/poly3/measureArea.js
```diff
@@ -14,19 +15,18 @@ const measureArea = (poly3) => {
...
 
-  // calculate a real normal
-  const a = vertices[0]
-  const b = vertices[1]
-  const c = vertices[2]
-  const ba = vec3.subtract(vec3.create(), b, a)
-  const ca = vec3.subtract(vec3.create(), c, a)
-  const normal = vec3.cross(ba, ba, ca)
+  // calculate a normal vector
+  const normal = plane.fromPoints(plane.create(), ...vertices)
```
z3dev, line 19: I think that calling poly3.plane() would be better as this function caches the plane, and will return the plane if already available. It might help performance.

packages/modeling/src/maths/plane/fromPoints.js
```diff
@@ -10,17 +10,39 @@ const vec3 = require('../vec3')
  * @returns {plane} out
  * @alias module:modeling/maths/plane.fromPoints
  */
-const fromPoints = (out, a, b, c) => {
-  const ba = vec3.subtract(vec3.create(), b, a)
-  const ca = vec3.subtract(vec3.create(), c, a)
-  vec3.cross(ba, ba, ca)
-  vec3.normalize(ba, ba) // normal part
-  const w = vec3.dot(ba, a)
+const fromPoints = (out, ...vertices) => {
```
z3dev, line 13: ``

z3dev, line 13: Use the spread operation is sexy but it seems kind of a performance hit.

`Newplane = plane.fromPoints(…polygon.vertices)`

z3dev:
I’ll run the performance suite using theses changes to see what the performance impact becomes.

An alternative would be to calculate the midpoint of the polygon, and use that in the calculation of the normal. But that could also be expensive.

hrgdavor:
Maybe check if normal is messed up after calculating via first 3 points, and then only use new calculation via all vertices if really necessary.

packages/modeling/src/geometries/poly3/measureArea.js
platypii: Done

packages/modeling/src/maths/plane/fromPoints.js
platypii: I didn't want to break a widely used API. Maybe a new `plane.fromPointArray` function? Or do a type check if the parameter is an array?

z3dev:
@platypii sorry for the delay.

I started a new job recently and have become a weekend warrior, so hang in there.

z3dev:
@platypii the performance tests were good, so I’ll merge this in a few days.

z3dev (approved):
Looks good. Performance tests didn't show any degradation.

PR #967, opened 2021-12-19
fix(modeling): revert measureBoundingBox to only cache per geometry

z3dev:
As the title says.. sigh

### All Submissions:


z3dev:
@hrgdavor please review.

i hate to revert but think that there's some conflicts in the cache look up; by geom VERSUS by points.

alternatively, if a test case can be created then a fix may be possible.

i submitted this because i have no time to dig deep into this.

hrgdavor:
@z3dev I will check if test case is possible and fix first.

z3dev:
> @z3dev I will check if test case is possible and fix first.

@hrgdavor please check that measureBoundingBox is correct before and after cloning / transforming shapes. Maybe a new test case is required.

 I believe that allocation of a new bounding box is required when reusing the cached bounding box. In other words, the original bounding box should not be changed if coached.

mister-jscad (approved):
Looks fine. :(

Issue #966, opened 2021-12-17
Missing breakouts in my model due to eliminated allocations of measureBoundingBox


jue89:
## Expected Behavior

This model creates two STL files for a 3D-printable enclosure:

https://github.com/jue89/homie/tree/main/devices/relay/housing

With the old version of @jscad/modeling everything looks fine:

<img width="1136" alt="Screenshot 2021-12-17 at 17 15 45" src="https://user-images.githubusercontent.com/6105784/146575141-a345b022-1838-483e-acf0-cd1338e7ca25.png">

## Actual Behavior

The `subtract()` on one screw hole doesn't seem to be applied. I've used `git bisect` to identify the commit breaking my model: 8abe361fc0249ec10e438382353b316aec4644a4

<img width="1136" alt="Screenshot 2021-12-17 at 17 14 23" src="https://user-images.githubusercontent.com/6105784/146575167-ba59fdc4-272f-4f80-9c95-2563b7ec5da6.png">

## Steps to Reproduce the Problem

  1. Clone the repo stated above, checkout 959d6b6e6c94df76b4af7d1000838112e0785d16 and cd into the dir `devices/relay/housing`
  2. `npm install && node mains-relay.js` and see everything working ... (package-lock.json picks the old version)
  3. `cd node_modules/quickbox && npm install && cd ../.. && node mains-relay.js` and observe the missing screw hole.

## Specifications

  - Version: @jscad/modeling@2.7.0
  - Platform: MacOS
  - Environment: nodejs

jue89:
Those are the lines creating the breakouts: https://github.com/jue89/node-quickbox/blob/d84e8b3d89ee814a8a9eeb87933413dcf657a587/lib/box.js#L151-L154

z3dev:
UG… I’ll look at the recent changes but those should not have changed any behavior. I suspect that the caching logic has issues.

hrgdavor:
@jue89 can you give an example of box constructor that produces model in screenshots ? I need a sample code to test

hrgdavor:
@z3dev 
> UG… I’ll look at the recent changes but those should not have changed any behavior. I suspect that the caching logic has issues.

The issue is that applyTransforms is not respecting immutability. I think it was the next step that we decided to do separately and forgot to do.
I will make i PR with fix applyTransforms and it should also fix this issue.

[...]

PR #970, opened 2021-12-26
fix(modeling): applyTransform ceate new geometry

hrgdavor:
The fix 
 - fix applyTransforms to produce new goemetry so it does not break other operations
 - fix geenralize to use mat4.intentity transforms in output geom (because toPolygons was used and transform applied)
 - this fix also depends on #973 to be  fix for #966

old tests fixed


z3dev:
General comments.

The behavior change to applyTransforms is going to affect performance. There are many cases where measurements are being called, and the toPoints is being called. 

Also, these changes don’t fix measureBoudingBox. At least, I can’t see the fix. So, I don’t think the test cases are robust.

hrgdavor:
the fix is in two parts:

First part is in the caching code
```js
    // create a new boundingBox. Mutating the passed boundingBox would break cached calculations
    boundingBox = [
      vec3.transform(vec3.create(), boundingBox[0], transforms),
      vec3.transform(vec3.create(), boundingBox[1], transforms)
    ]
```

the second is in aplyTransform to not mutate/destroy the geometry.

-----------------

> There are many cases where measurements are being called, followed by calling toPoints.

That is a poor design choice, and such code should be optimized to perform better. I would like to see the examples where this applies.

-----------------

unit test there in the code  confirms both when error happens (without the fix), and works after fix .
the fix also works for the 2 more complex examples from related github issue #966

z3dev:
I think this should be split into two pieces. The fix for measureBoundingBoz should not rely upon other changes.

hrgdavor:
ok, it is split now, but both are needed for the fix to be complete

t1u1:
Thanks for your work @hrgdavor 

This looks like an important fix.

I have been building a complex model using jscad, and it has been going fine until today. When I added a transform.rotate inside the model, things started looking weird. It took me a long while to realize that it was the rotation that was causing the weird geometry. 

Perhaps it is related to this issue. I will try to reduce the model to a smaller test case, but in the meanwhile, if this PR looks good it will be great to have this merged.

hrgdavor:
@t1u1 this PR is more pragmatic change than a bug-fix. But when you have a test example, we can take look what is causing you issues.

z3dev:
> Thanks for your work @hrgdavor
> 
> This looks like an important fix.
> 
> I have been building a complex model using jscad, and it has been going fine until today. When I added a transform.rotate inside the model, things started looking weird. It took me a long while to realize that it was the rotation that was causing the weird geometry.
> 
> Perhaps it is related to this issue. I will try to reduce the model to a smaller test case, but in the meanwhile, if this PR looks good it will be great to have this merged.

See issue #985 about some viewer issues.

You can check if it’s a viewer issue. Just call toPolygons() before returning the shape from main(). That will rest the transforms, and the viewer should display properly.

t1u1:
Thanks @hrgdavor @z3dev 

After some more experiments I think the issue I am seeing is something else. Boolean operations involving geometries that touch each other cause weirdness. It was just getting triggered during rotation at certain angles.

Will file a separate issue for it.

z3dev:
@hrgdavor i think these changes are no longer important. There were changes to improve copy() functions as well as transform() functions.

if you still think these are important then let's discuss further.

hrgdavor:
I need to consider it a bit more, and see if it could cause issues. I vageuely rememeber that it could actually be ok, in the sense it would not break other instances.

hrgdavor:
I will revisit this at later time, adn create new issue if needed

PR #973, opened 2021-12-29
fix(modeling): undo-revert measureBoundingBox

hrgdavor:
fix a bug in measureBoundingBox caused by mutating original boundingBox with transform

this fix also depends on #970 to be complete  fix for #966


z3dev:
@hrgdavor thanks for splitting this out.

does this fix the issue #966 ?
also, can you add the same tests for geom2 and path2? It would be good to verify those work as well.

z3dev:
@hrgdavor Can this be closed?

hrgdavor:
It has been a long time. I forgot about it. I will check it out and report back.

Issue #558, opened 2017-06-16
Naming Objects


thesourcerer8:
I would like to be able to assign (programmatically) names to objects and have those names be used when exporting to X3D, STL, ... where the export formats support names, so that I can more easily reference the objects in X3D.
Support for the animation capabilities of X3D would be great as well.

I have been asked to reassign this issue here:
https://github.com/EinsteinsWorkshop/BlocksCAD/issues/17#issuecomment-308742189

z3dev:
@thesourcerer8 can you help define what you would like the X3D file to contain? there's a lot of possibilities.

https://www.web3d.org/x3d/content/examples/X3dSceneAuthoringHints.html#NamingConventions

thesourcerer8:
Yes, a DEF attribute seems to be the way to do it.

z3dev:
Do you want names in another export format, like DXF?

thesourcerer8:
I think it would be a good idea to put the names in all the output formats that support them.

z3dev:
Support of 'name' attribute added to V2 DXF serializer. #684

z3dev:
Comment... each serializer should provide a list of supported attributes, such as 'name' or 'class' or 'id'. In addition, these attributes need to be preserved by generalize(), possible a new option.

z3dev:
From the forum...

'class' element for sure is interesting as it will ease styling. About 'id', yes it must be used carefully to ensure that each object has its unique id, generally I use names forged from data (line between triangle 23 and 111 has id 'L_23_111', etc.). 'name' is deprecated so it's maybe wise to avoid it, on very old js code I remember i used 'tag', as svg doesn't use it, as svg can use custom attributes, it can be anything. And it would be really great to have exported svg from jscad with such fine details, because it will allow rework those outputs with great ease.

z3dev:
Support of 'id' and ‘class’ attributes added to V2 SVG serializer. #974

z3dev:
Summary
- AMF TBD
- DXF support for 'name' attribute #684 
- OBJ TBD
- STL impossible as format does not support
- SVG support for 'id' and 'class' #974 
- 3MF

z3dev:
Closing this issue as the relevant serializers have been changed to support names, ids, etc.

PR #974, opened 2022-01-02
Improve svg rendering

z3dev:
These changes improve the 'rendering' of designs exported to SVG. 
- set defaults for fill, fill-rule, stroke-width at the SVG level
- remove above from path serialization
- corrected fill color for continuous paths (geom2)
- added support for special attributes; id and class

### All Submissions:

### DIFF
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -37,6 +37,7 @@
       "packages/io/amf-deserializer/src",
       "packages/io/amf-serializer",
       "packages/io/svg-deserializer/src",
+      "packages/io/svg-serializer",
       "packages/io/x3d-deserializer/src"
     ],
     "includePattern": ".+\\.js(doc|x)?$",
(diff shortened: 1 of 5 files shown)


z3dev:
@hrgdavor please review / try if you have time. I think these changes will help users when exporting to SVG.

z3dev:
See #558 for how to add a ‘name’ to objects

hrgdavor:
np. I will take a look a bit and test

hrgdavor:
It will be few days before I can do so, I am traveling today and tomorrow, so I can't test this right away

hrgdavor:
I have been playing with it, and it looks good in the svg source. I am not a fan of using fill.

While jscad renders lines
image

the svg export uses fill

image

personally I would prefer stroke
image

hrgdavor:
I have some complaint/proposals that are unrelated to this PR.

jscad is generating poly for each curve immediately, so an export to SVG is a poly instead of circle, arc, besier.

Without changing jscad core, one option would be to explore integration with other 2d libs like maker.js ... actually all that is needed is to improve require in the future because V2 now allows us to script with any valid JS lib.

hrgdavor (approved):
It is a good step forward.

PR #976, opened 2022-01-18
fix(web): handle parameter definition exceptions

platypii:
There is a UI bug when a user provides malformed parameter definitions. An exception is thrown internally, but the UI gets stuck with the message:
> processing, please wait

This PR catches exceptions thrown inside the `designParameters` function. If an exception is thrown, it gets added to the `status` object so it can be displayed to the user. After applying this PR, the user message becomes more helpful:
> ERROR: Definition of choice parameter (tooling) should include a 'values' parameter

Example script:
```js
const { cuboid } = require("@jscad/modeling").primitives

function getParameterDefinitions() {
  return [{name: "tooling", type: "choice", caption: "Tooling"}] // Oops forgot "values"
}

function main(params) {
  return cuboid({})
}
module.exports = { getParameterDefinitions, main }
```

### All Submissions:

### DIFF
--- a/packages/web/src/ui/views/designParameters.js
+++ b/packages/web/src/ui/views/designParameters.js
@@ -4,10 +4,11 @@ const { createParamControls } = require('./parameterControls')
 
 const designParameters = (state, paramsCallbacktoStream, i18n) => {
   const { parameterValues, parameterDefinitions, parameterDefaults } = state.design
-  const { controls } = createParamControls(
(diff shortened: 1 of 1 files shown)


z3dev:
@platypii This is great! So many users are still complaining about the stack traces, error messages, etc. This is going to help many many users. SUPER

z3dev (approved):
@platypii Thanks again. I reviewed the code, and tried the changes. This is a nice fix.

PR #996, opened 2022-02-08
refactor(modeling): Use hypot instead of sqrt

platypii:
This PR changes `Math.sqrt(x*x + y*y + z*z)` into `Math.hypot(x, y, z)`.

See [MDN: Math.hypot](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot) for details.

Math.hypot has benefits of: not overflowing on large numbers, more accurate floating point math for small numbers, and generally nicer to read. It seems wise to code defensively, to avoid as many floating point bugs as possible.

I added a test case to demonstrate the problem.

### All Submissions:

### DIFF
--- a/packages/io/test/helpers/nearlyEqual.js
+++ b/packages/io/test/helpers/nearlyEqual.js
@@ -8,6 +8,10 @@ const nearlyEqual = (t, a, b, epsilon, failMessage) => {
   const absA = Math.abs(a)
   const absB = Math.abs(b)
   const diff = Math.abs(a - b)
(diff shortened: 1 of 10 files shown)


hrgdavor:
looks good to me :) ... nice catch

t1u1:
@z3dev This looks ripe for merging ? :smiley:

z3dev (approved):
@platypii thanks
let's hope there's a little performance boost as well.

PR #997, opened 2022-02-17
refactor(modeling): linear runtime toOutlines

platypii:
The `toEdges` function inside `geom2.toOutlines` includes a quadratic runtime loop to find unique vertices. I re-wrote the function to use a map so that it can run in linear time.

For a circle with 100 segments this reduces runtime from 0.9ms to 0.4ms (2x speedup). 
For a circle with 1000 segments this reduces runtime from 12ms to 2ms (6x speedup). 
For a circle with 10000 segments this reduces runtime from 150ms to 16ms (9x speedup). 
For a circle with 50000 segments this reduces runtime from 4400ms to 90ms (50x speedup). 
For a circle with 100000 segments this reduces runtime from 16000ms to 160ms (100x speedup). 

This helps performance more as the complexity of the design increases. But it is also still faster on small geometry.

Performance test code:

```js

const { toOutlines } = jscad.geometries.geom2

const main = () => {
  const geometry2 = circle({ radius: 5, segments: 10000 })
  const start = performance.now()
  const outlines = toOutlines(geometry2)
  console.log(`Outline generated ${performance.now() - start} ms`)
  return geometry2
}

module.exports = { main }
```

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/geom2/toOutlines.js
+++ b/packages/modeling/src/geometries/geom2/toOutlines.js
@@ -7,21 +7,16 @@ const toSides = require('./toSides')
  * This allows the edges to be traversed in order.
  */
 const toEdges = (sides) => {
-  const uniquevertices = []
+  const vertices = {}
   const getUniqueVertex = (vertex) => {
-    const i = uniquevertices.findIndex((v) => vec2.equals(v, vertex))
-    if (i < 0) {
-      uniquevertices.push(vertex)
-      return vertex
+    const key = vertex.toString()
+    if (!vertices[key]) {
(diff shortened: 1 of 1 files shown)


hrgdavor:
wow, nice catch. and nice fix :)

z3dev:
someone went deep down inside geom2.toOutlines
:)

t1u1:
> it's interesting that the creation of a string plus hashmap lookups can be faster then tight loops. javascript is a strange beast.

O(N) will always be faster than O(N^2) for large enough N, and for small N, the difference doesn't matter much anyway.

Issue #999, opened 2022-02-18
Uncaught Type Error in viewer.js in package Web 


samuelwang48:
## Expected Behavior

A model related method should not break if provided with invalid values such as null values.

## Actual Behavior

Given any method that takes repeatable geom2, geom3 or etc, upon being flatten, it doesn't detect or filter the value of being a null type, resulting in an Uncaught Javascript error. 

For example, the null should either be ignored, filtered or captured but it is let through and results in an uncaught error in viewer.js
```
  return [Primitive.sphere({
    center: [0, 0, 0],
    radius: 12
  }), null];
```

## Steps to Reproduce the Problem

  1. paste this piece of code into codemirror editor
  ```
const jscad = require('@jscad/modeling');
const Primitive = jscad.primitives;

function main() {
  return [Primitive.sphere({
    center: [0, 0, 0],
    radius: 12
  }), null];
}

module.exports = { main }
```
  2. SHIFT+ENTER
  3. Uncaught error thrown and there isn't any way to rerun except for a hard reload of the page.

## Specifications

  - Version: production
  - Platform: all
  - Environment: browser

samuelwang48:
Line 266 is assuming the object is an object instead of being a null
Line 264 is the fix I made locally

![Screen Shot 2022-02-18 at 01 05 13](https://user-images.githubusercontent.com/23735294/154628405-51ddec57-6ab5-483f-b8a0-3d272d2971b7.png)
![Screen Shot 2022-02-18 at 01 06 06](https://user-images.githubusercontent.com/23735294/154628407-69762b1b-7933-4c4a-a793-8a13b0b8452d.png)

samuelwang48:
I am open for a discussion on this issue.
Given the complexity of the codebase there might be several ways to solve it.
Here are a few options I can think of.
1. Address this issue in flatten() but IMO flatten should just do flatten even there are null elements in the list. Introducing filtering logic seems a bit off.
2. In all the model methods, every single one of them, add a filtering logic to make sure null is either not accepted, ignored or filtered to protect the existing implementations anywhere else.
3. In `compareSolids`, return false if any null is detected. It is not wrong to assert that because any solid compares to null is not equal but should a solid be compared to a null?

Option 2 seems to be a right thing to do
Option 3 generates no error in all existing test cases but side effect is unknown to someone new

hrgdavor:
I am in favor of viewer handling null by ignoring it.

samuelwang48:
[...]

PR #1002, opened 2022-02-18
Uncaught Type Error in viewer.js in package Web #999

samuelwang48:
### All Submissions:


z3dev (changes requested):
@samuelwang48 please revert the change to the jscad-web.min.js library. this is rebuilt during the release process, and shouldn't be part of code changes. thanks

you can just copy the latest from 'master', and commit that version. that should remove these changes.

samuelwang48:
> @samuelwang48 please revert the change to the jscad-web.min.js library. this is rebuilt during the release process, and shouldn't be part of code changes. thanks
> 
> you can just copy the latest from 'master', and commit that version. that should remove these changes.

Hi @z3dev the specific change aforementioned has been reverted. Feel free to let me know if there is anything else I can help. Thanks

z3dev:
@samuelwang48 thanks for these changes, but i think another solution might be slightly better.

The viewer **only** supports the JSCAD geometries, so it's probably better just to filter the solids returned from the design, here... viewer.js, line 146

`    const solids = state.design.solids.filter((solid) => solid && (solid instanceof Object))`

Thoughts?

samuelwang48:
> @samuelwang48 thanks for these changes, but i think another solution might be slightly better.
> 
> The viewer **only** supports the JSCAD geometries, so it's probably better just to filter the solids returned from the design, here... viewer.js, line 146
> 
> ` const solids = state.design.solids.filter((solid) => solid && (solid instanceof Object))`
> 
> Thoughts?

Hi @z3dev, Thanks! I've made the changes accordingly. The said issue seems to have been solved. All tests passed.

z3dev (approved):
@samuelwang48 thanks. i hope this is the last time fixing this issue. ;)

PR #1006, opened 2022-02-20
fix(io): remove size limit from Blob

platypii:
When exporting output files via the command line, there is a 32mb limit on the size of the output file. If you try to export an STL file larger than 32mb, you get an "out of range" error on the CLI (but not in the browser).

The reason is a fixed buffer size in the Blob shim used for node environments. I fixed it by using a dynamic buffer size, so that it will never be too small or too large.

```js
const { sphere } = require('@jscad/modeling').primitives
module.exports = { main: () => {
  return sphere({ segments: 850 })
}}
```

### All Submissions:


z3dev:
Ooooo…. NODEJS added Blob as part of 15.5 going forward. It only took them 10 years! HA HA HA

So, JSCAD should try using that if possible, then fall back to the JSCAD version.

z3dev:
> Ooooo…. NODEJS added Blob as part of 15.5 going forward. It only took them 10 years! HA HA HA
> 
> So, JSCAD should try using that if possible, then fall back to the JSCAD version.

Yeah. We need to support the latest LTS versions of NODEJS as well.

platypii:
:facepalm: you're right, once I updated my node from 12 to 17, I no longer hit the 32mb limit.

I think if we want to support the Blob shim for node pre-15, then this PR is still a good fix. If the plan is to deprecate the whole thing in favor of the official implementation, then can close the PR.

z3dev:
> 🤦 you're right, once I updated my node from 12 to 17, I no longer hit the 32mb limit.
> 
> I think if we want to support the Blob shim for node pre-15, then this PR is still a good fix. If the plan is to deprecate the whole thing in favor of the official implementation, then can close the PR.
> 
> FYI: Ubuntu 20.04 LTS ships with Node 10.19.0, Ubuntu 21.10 ships with Node 12.22.5.

Yeah. It might be a little early to remove the JSCAD Blob. I'm sure most users are still using early versions of NodeJS.

Maybe we should just push users to use a newer version of NodeJS if they happen to hit this issue. Is there an open issue about this? We could add some comments to that issue.

platypii:
If we're going to keep the Blob shim, it seems like it makes more sense to just fix the bug by merging this PR? The alternative is making a new PR with a user warning, and keeping a github issue around to point users to. Seems easier to just fix the bug?

z3dev:
@platypii just one last question... Blob is being used when exporting designs as STL, etc. Did you test a few exports?

platypii:
@z3dev yes I tested export on a variety of models. I tested both smaller and larger models to make sure the fix was good. Note: all tests were run on Node 12, which does require the Blob shim. :+1:

z3dev (approved):
Ok. Let’s adjust the shim for now. Thanks again @platypii

PR #1007, opened 2022-02-20
perf(modeling): linear unique points in hullPath2

platypii:
The function `hullPath2` has a routine to find unique vertices which was implemented with a quadratic array loop. I re-wrote the function to use a map so that it can run in linear time.

With 100 segments this reduces runtime from 2.3ms to 0.9ms (2.5x speedup).
With 1000 segments this reduces runtime from 23ms to 4.5ms (5x speedup).
With 10000 segments this reduces runtime from 750ms to 34ms (22x speedup).
With 50000 segments this reduces runtime from 11500ms to 140ms (80x speedup).
With 100000 segments this reduces runtime from 45000ms to 290ms (150x speedup).

Performance test code:

```js
const { geom2, path2 } = jscad.geometries
const { hull } = jscad.hulls
const { translate } = jscad.transforms

const main = () => {
  // construct path2
  const circ = circle({ segments: 10000 })
  const outline = geom2.toOutlines(circ)[0]
  const path = path2.fromPoints({closed: true}, outline)

  // construct hull
  const start = performance.now()
  const obj2d = hull(
    translate([17, 21], path),
    translate([7, 0], path),
  )
  console.log(`Hull time ${performance.now() - start} ms`)
  return obj2d
}

module.exports = { main }
```

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/hulls/hullPath2.js
+++ b/packages/modeling/src/operations/hulls/hullPath2.js
@@ -16,11 +16,15 @@ const hullPath2 = (...geometries) => {
 
   // extract the unique points from the geometries
   const uniquepoints = []
+  const found = {}
   geometries.forEach((geometry) => {
     const points = path2.toPoints(geometry)
     points.forEach((point) => {
-      const index = uniquepoints.findIndex((unique) => vec2.equals(unique, point))
-      if (index < 0) uniquepoints.push(point)
+      const key = point.toString()
+      if (!found[key]) {
+        uniquepoints.push(point)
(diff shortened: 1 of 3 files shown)


hrgdavor:
Another one :) ... nice find and fix

z3dev (approved):
@platypii do you sleep?
Thanks for this performance boost again.

platypii:
@z3dev nice point about Map/Set. I changed to Set and the 100k circle example went from 290ms -> 210ms. Awesome! Who needs sleep?

PR #1008, opened 2022-02-21
perf(modeling): expand performance and bug fix

platypii:
There is a bug in [expandShell.js](https://github.com/jscad/OpenJSCAD.org/blob/b06288af2fe8f7b007e4b10eac413c74bd6c1bfa/packages/modeling/src/operations/expansions/expandShell.js#L31) where the edge check in findIndex is always false, because it is comparing an edge to a vertex.

The result of that bug is that two cylinders are added to the expansion for each edge, instead of just one. Two cylinders on each edge is unnecessary and slows the computation down. This new code produces cleaner geometry too.

I also changed the functions to avoid linear searches for unique vertices and edges. Actually that's where I started, I only noticed the bug after I accidentally fixed it.

The following example (which is a real test from [expand.test.js](https://github.com/jscad/OpenJSCAD.org/blob/c8a37ad6430aec8361a5e121730b34629c0d969c/packages/modeling/src/operations/expansions/expand.test.js#L120-L123)) reduced runtime from 4200ms on master, to 1600ms on this branch. This is a 10% reduction in total `npm test` time! (23s -> 20s)

```js
const { expand } = jscad.expansions

const main = () => {
  const obj = sphere({ radius: 5, segments: 8 })
  return expand({ delta: 5 }, obj)
}

module.exports = { main }
```

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/expansions/expand.test.js
+++ b/packages/modeling/src/operations/expansions/expand.test.js
@@ -120,7 +120,7 @@ test('expand: expanding of a geom3 produces expected changes to polygons', (t) =
   const geometry2 = sphere({ radius: 5, segments: 8 })
   const obs2 = expand({ delta: 5 }, geometry2)
   const pts2 = geom3.toPoints(obs2)
-  t.is(pts2.length, 2065)
+  t.is(pts2.length, 1612)
 })
 
 test('expand (options): offsetting of a complex geom2 produces expected offset geom2', (t) => {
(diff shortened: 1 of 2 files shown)


z3dev:
@platypii can you try this with the examples in #876 ? i'm curious to see if the bug fix helps.

z3dev (approved):
@platypii cool changes. and a fix to a very subtle bug as well.

PR #1009, opened 2022-02-22
feat(modeling): extrudeLinear support for path2

platypii:
`extrudeLinear` previously only supported geom2. But it's pretty easy to convert a path2 to a geom2, so I made an `extrudeLinearPath2` function to support linear extrusion on path2 geometries. Includes tests.

### All Submissions:


packages/modeling/src/operations/extrusions/extrudeLinearGeom2.js
```diff
@@ -7,7 +7,7 @@ const slice = require('./slice')
 
 const extrudeFromSlices = require('./extrudeFromSlices')
 
-/*
+/**
```
z3dev, line 10: This function is not part of the extrusions (index.js), so should probably be left out of user documentation.

Also, due to the object orientated slant of JSDOC, we have to provide an alias for each function in order for the function to have a proper name space.

`@alias module:modeling/extrusions.extrudeLinear`

z3dev (changes requested):
@platypii let's keep the internal functions unexposed in documentation.

there might be some alternatives but then the whole set of documentation needs to be reconsidered.

packages/modeling/src/operations/extrusions/extrudeLinearGeom2.js
platypii: @z3dev my intention was not to include this in user documentation. I thought I was just fixing typos. I will undo the comment changes in this PR.

Its a bummer if we can't use jsdoc `/**` comments for private functions though. Without the double-asterisk, many IDEs won't show autocomplete hints or syntax highlighting, and many linters won't pickup bugs.

With `/*` non-jsdoc comment (no autocomplete hint, no jsdoc coloring):
no-jsdoc

Wtih `/**` jsdoc comment (full autocomplete, jsdoc color hints):
jsdoc

z3dev: That is a bummer. IDE highlights are nice hints for developers.

z3dev: I found @ignore in the JSDOC manual. This might allow the IDE to use the docs, while removing the function from documentation.

https://jsdoc.app/tags-ignore.html

packages/modeling/src/operations/extrusions/extrudeLinear.js
```diff
@@ -1,8 +1,9 @@
 const flatten = require('../../utils/flatten')
 
-const geom2 = require('../../geometries/geom2')
+const { geom2, path2 } = require('../../geometries')
```
z3dev, line 3: Not wrong but we try not to use the indexes when requiring specific objects. In the past, there have been circular references which node couldn't handle. Can you split this into two (2) requires statements?

packages/modeling/src/operations/extrusions/extrudeLinearPath2.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param {Object} [options] - options for extrude
+ * @param {Array} [options.offset] - the direction of the extrusion as a 3D vector
+ * @param {Number} [options.twistAngle] - the final rotation (RADIANS) about the origin
+ * @param {Integer} [options.twistSteps] - the number of steps created to produce the twist (if any)
+ * @param {path2} geometry - the geometry to extrude
+ * @returns {geom3} the extruded 3D geometry
+*/
+const extrudePath2 = (options, geometry) => {
+  if (!geometry.isClosed) throw new Error('extruded path must be closed')
```
z3dev, line 16: I wonder which is better... throwing an error or closing the path if not already (clone)?

If throw then a comment about 'only closed paths are accepted' would be helpful in extrudeLinear() documentation.

z3dev (changes requested):
A few suggestions. Let me know what you think.

z3dev (approved):

PR #1013, opened 2022-02-25
perf(modeling): fast offsetFromPoints

platypii:
In the `corner: edge` section of `offsetFromPoints`, there was a linear search that was hurting performance. I replaced it with a Map lookup for better performance.

`offsetFromPoints` is used internally by `expandGeom2`, `expandPath2`, and `extrudeRectangular`.

However in order to make the indices stable, I change the way "edge" points are constructed. Previously it would take the beginning and end points of the corner, and add a third "intersection" point. But that adds unnecessary co-linear vertices to the expansion. So instead I now delete the beginning and end points and only add the intersection point. This results in geometry which is the same shape, but simpler with fewer vertices and edges. This helps performance down the line too, if you're doing any operations on the cleaner expanded geometry.

before:
```
--o--o
     |
     o
     |
```

after:
```
-----o
     |
     |
     |
```

Performance:

// 100 segments: 3ms -> 2ms
// 1000 segments: 25ms -> 7ms
// 10000 segments: 450ms -> 41ms
// 50000 segments: 14000ms -> 140ms
// 100000 segments: 82000ms -> 260ms

```js
expand({ delta: 2, corners: 'edge' }, circle({ segments: 10000 }))
```

### All Submissions:


z3dev (approved):
@platypii thanks for the nice performance improvement!

PR #1016, opened 2022-02-27
Github Actions

t1u1:
This runs `npm install` and `npm test` on 3 different versions of node: 12, 14, 16

Is there anything else required as part of CI?

We could add `npm publish` for snapshots in the future, but that will require setting up auth keys in Github.


t1u1:
As I am a first time contributor to this repo, Github actions will not be triggered automatically on this PR. An admin will have to manually trigger them.

z3dev:
> This runs `npm install` and `npm test` on 3 different versions of node: 12, 14, 16
> 
> Is there anything else required as part of CI?

Actually, there are three steps
- npm install
- npm run bootstrap
- npm test

Thanks again

platypii:
Why is `lerna bootstrap` needed? Tests seem to work without it.

z3dev:
> Why is `lerna bootstrap` needed? Tests seem to work without it.

Good point. So, this should be fine.

t1u1:
@z3dev Are you seeing the button in this PR to trigger the action?

> > Why is `lerna bootstrap` needed? Tests seem to work without it.
> 
> Good point. So, this should be fine.

This made me wonder, how do we ensure that the tests are running correctly. So I made a commit that deliberately breaks the code a bit to see if CI is able to catch it and then rolled back the bad commit. Confirmed to be working fine!

Here are the results on the forked repo: https://github.com/t1u1/OpenJSCAD.org/actions

z3dev:
We just received credits to resume Travis CI. I'll merge this into the main stream this weekend when I have time to watch, and fix issues if any.

z3dev (approved):
Let's give this a shot...

PR #1021, opened 2022-03-05
Earcut triangulation algorithm for extrudeFromSlices

### LINKED ISSUE #907: Birthday Balloons
Screen Shot 2021-09-10 at 10 35 07 AM

## Expected Behavior
great birthday celebration

## Actual Behavior
deflated balloons ruining the party

  1. visit https://www.openjscad.xyz/#
  2. load 'Birthday Balloons' example
  3. Age 7
  4. 77, 777 is okay

## Specifications

  - Version:
  - Platform:
  - Environment: (browser, local server, node.js etc)

platypii:
It's finally ready! This is an implementation of the Earcut triangulation algorithm. This directly fixes #907 but I believe it will also fix a whole class of geometry artifact bugs. (@t1u1 can confirm)

Credit to the mapbox and three.js earcut implementations. The majority of the earcut implementation comes from them. But I changed a fair amount of code to 1) accept jscad polygons 2) use jscad style 3) break apart a multi-hundred line file into discrete modules. This required careful attention to how holes are handled, and required writing new code to associate holes with their containing solids: `PolygonHierarchy`. I also included my `repairSlice` function to mend gaps in a slice before extruding, since earcut requires a closed geometry. I wish it was simpler, but I am very happy with the results.

## No more glitches!

balloons

## Air-tight

Air-tight geometry from extrusions! The output of `extrudeLinear` generates polygons that only use the existing vertices. It doesn't create new splits (no BSP tree). This ensures that the geometry is always manifold, air-tight geometry!

## No more floating point errors

Test from `extrudeLinear.test.js` show fixes to ugly floating point errors:
```diff
-    [[0, 0, 15], [11.999999999999998, 0, 15], [6, 10, 15]],
-    [[6.000000000000001, 10, 0], [12, 0, 0], [8.881784197001252e-16, 0, 0]]
+    [[12, 0, 15], [6, 10, 15], [0, 0, 15]],
+    [[0, 0, 0], [6, 10, 0], [12, 0, 0]]
```

## Performance

When extruding a 50k segment circle, `extrudeLinear` has improved from 20 seconds to 0.5 seconds!

```js
extrudeLinear({}, circle({ radius: 5, segments: 50000 }))
```

Faster AND more accurate :sunglasses: 

### All Submissions:


z3dev:
@platypii when ready, just select someone to review. maybe @hrgdavor and @z3dev would be a good starting point.

looks like you had some fun!

packages/modeling/src/operations/extrusions/earcut/polygonTrees.js
```diff
@@ -0,0 +1,43 @@
+const geom2 = require('../../../geometries/geom2')
+const OrthoNormalBasis = require('../../../maths/OrthoNormalBasis')
```
z3dev, line 2: UG... I'm trying to get rid of OrthoNormalBasis. It's really untested and full of junk.
I know that it's still being used but only by booleans, but not for long.

If useful, can you break out the useful pieces just for earcut? Can something else be used?

packages/modeling/src/operations/extrusions/earcut/assignHoles.js
```diff
@@ -0,0 +1,85 @@
...
+ * @returns {Array} an array of polygons with associated holes
+ * @alias module:modeling/geometries/geom2.toTree
+ *
+ * @example
+ * const geometry = subtract(rectangle({size: [5, 5]}), rectangle({size: [3, 3]}))
+ * console.log(assignHoles(geometry))
+ * [{
+ *   "solid": [[-2.5,-2.5],[2.5,-2.5],[2.5,2.5],[-2.5,2.5]],
+ *   "holes": [[[-1.5,1.5],[1.5,1.5],[1.5,-1.5],[-1.5,-1.5]]]
+ * }]
```
hrgdavor, line 20: I think we need official support for this type of geometry, that is structured and gives more paths for optimizations. 
the code looks fine, I am just commenting this a as discussion point for the future.

packages/modeling/src/operations/extrusions/extrudeFromSlices.js
```diff
@@ -54,6 +55,9 @@ const extrudeFromSlices = (options, base) => {
 
   if (numberOfSlices < 2) throw new Error('numberOfSlices must be 2 or more')
 
+  // Repair gaps in the base slice
```
z3dev, line 58: Why the repair? Were there some cases where the other extrusion functions were creating invalid slices?

packages/modeling/src/operations/extrusions/extrudeFromSlices.js
```diff
@@ -54,6 +55,9 @@ const extrudeFromSlices = (options, base) => {
...
+  repairSlice(base)
```
hrgdavor, line 59: I am personally against automatic repair, but if there is a parameter that controls if repair is called, I will be ok if repair is done by default. I really want to be able to turn off repair as a user.

packages/modeling/src/operations/extrusions/slice/calculatePlane.js
```diff
@@ -23,10 +23,13 @@ const calculatePlane = (slice) => {
   let farthestEdge
   let distance = 0
   edges.forEach((edge) => {
-    const d = vec3.squaredDistance(midpoint, edge[0])
-    if (d > distance) {
-      farthestEdge = edge
-      distance = d
+    // Make sure that the farthest edge is not a self-edge
```
z3dev, line 26: after repair (above), is this still possible?

packages/modeling/src/operations/extrusions/slice/repairSlice.js
```diff
@@ -0,0 +1,47 @@
...
+    let bestReplacement
+    missingOut.forEach((key2) => {
+      const v2 = vertexMap[key2]
+      const distance = Math.hypot(v1[0] - v2[0], v1[1] - v2[1])
+      if (distance < bestDistance) {
+        bestDistance = distance
+        bestReplacement = v2
+      }
+    })
+    console.log(`Repair gap ${v1} → ${bestReplacement} dist ${bestDistance}`)
```
z3dev, line 35: console

z3dev (commented):
Please add a few comments to give credit to some of the sources. Lots of cool stuff in here.

I have a lot of question.. of course. And a few suggestions. But generally, this looks solid.

So... we use internally for 'capping' extrusions. Any other use? Will users be interested in the earcut functionality?

hrgdavor (changes requested):
I am not one to review stuff in depth, @z3dev is much better in that than me. I have one small request: 
 -add a param so auto-repair of base slice can be turned off.

Also I want to repeat here, that I have seen a function that creates a geometry definition that has neatly separated solid parts from holes. I really would like to see this type of geometry officially supported in API as a top-level construct like other 2d geom types.

packages/modeling/src/geometries/poly3/invert.js
```diff
@@ -9,7 +9,12 @@ const create = require('./create')
  */
 const invert = (polygon) => {
   const vertices = polygon.vertices.slice().reverse()
-  return create(vertices)
+  const inverted = create(vertices)
+  if (polygon.plane) {
+    // Invert plane to save recompute
```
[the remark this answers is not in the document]
platypii: Turns out we already have `plane.flip()`. I switched to using that.

packages/modeling/src/operations/extrusions/earcut/assignHoles.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+  })
+  // Map indices back to points
+  return children.map((holes, i) => ({
+    solid: outlines[solids[i]],
+    holes: holes.map((h) => outlines[h])
+  }))
+}
+
+/**
```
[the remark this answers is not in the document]
platypii: I learned about our doc system after I started this branch :-) I have fixed these doc comments now, thanks.

packages/modeling/src/operations/extrusions/earcut/assignHoles.js
```diff
@@ -0,0 +1,85 @@
...
+ * Find the item in the list with smallest score(item)
+ */
+const minIndex = (list, score) => {
```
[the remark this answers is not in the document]
platypii: Yes there is a precondition check on line 53. I added comments describing both the undefined behavior and the precheck.

packages/modeling/src/operations/extrusions/earcut/index.js
```diff
@@ -0,0 +1,247 @@
...
+ * main ear slicing loop which triangulates a polygon (given as a linked list)
+ */
+const earcutLinked = (ear, triangles, dim, minX, minY, invSize, pass) => {
+  if (!ear) return
+
+  // interlink polygon nodes in z-order
+  if (!pass && invSize) indexCurve(ear, minX, minY, invSize)
+
+  let stop = ear
+  let prev; let next
```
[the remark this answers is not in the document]
platypii: Fixed good catch

packages/modeling/src/operations/extrusions/earcut/linkedList.js
```diff
@@ -0,0 +1,51 @@
...
+
+const removeNode = (p) => {
+  p.next.prev = p.prev
+  p.prev.next = p.next
+
+  if (p.prevZ) p.prevZ.nextZ = p.nextZ
+  if (p.nextZ) p.nextZ.prevZ = p.prevZ
+}
+
+function Node (i, x, y) {
```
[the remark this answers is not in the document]
platypii: done

packages/modeling/src/operations/extrusions/earcut/polygonTrees.js
platypii: done. I only needed a couple line from OrthoNormalBasis, and they are pretty simple vector operations. I moved them into polygonTrees.

packages/modeling/src/operations/extrusions/earcut/linkedPolygon.js
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    p = p.next
+  } while (p !== start)
+
+  return filterPoints(p)
+}
+
+/*
+ * check if a polygon diagonal intersects any polygon segments
```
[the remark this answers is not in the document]
platypii: not sure I understand the question? this implementation is slightly different than the other intersect functions in the codebase, so I wasn't quite sure how to unify them easily.

packages/modeling/src/operations/extrusions/earcut/linkedPolygon.js
```diff
@@ -0,0 +1,197 @@
...
+    (
+      locallyInside(a, b) && locallyInside(b, a) && middleInside(a, b) && // locally visible
+        (area(a.prev, a, b.prev) || area(a, b.prev, b)) || // does not create opposite-facing sectors
+        equals(a, b) && area(a.prev, a, a.next) > 0 && area(b.prev, b, b.next) > 0
+    )
+
+/*
+ * check if two segments intersect
+ */
+const intersects = (p1, q1, p2, q2) => {
```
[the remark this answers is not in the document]
platypii: the calculations are pretty fast (addition/multiplication). also this implementation is slightly different than other intersect function so I also left it as is.

packages/modeling/src/operations/extrusions/slice/calculatePlane.js
platypii: I fixed this case in two ways:

1. In `repairSlice`, since there is no need to have duplicate edges in a "repaired" slice. 
2. In `slice.calculatePlane` since in theory someone could call it from a different function, and then it would be nice if the plane calculation worked.

So to answer your question, no, this case can't be hit in `extrudeFromSlices` IF repair is enabled. If repair is not enabled, or `slice.calculatePlane` is called directly (it is a [publicly documented api](https://openjscad.xyz/docs/module-modeling_extrusions_slice.html)), then this is a nice fix, in my opinion.

packages/modeling/src/operations/extrusions/extrudeFromSlices.js
platypii: The repair is necessary to make common examples work. Many of the existing jscad functions, like the BSP trees, will often return non-closed geometries. The birthday balloons, for example, produces a non-closed slice. I felt like it was important to not break existing designs, and so repair seemed necessary.

packages/modeling/src/operations/extrusions/slice/repairSlice.js
platypii: I thought about this one a fair amount. I agree that I don't like having console output mixed into a library. However, my reasoning was that I felt a user might want to know that modification of their design had happened. Especially with "repair" being optional now (based on hrg's feedback), a user can use the presence of the console.log to know whether it is safe to disable repair checks. Open to suggestions here.

packages/modeling/src/operations/extrusions/extrudeFromSlices.js
platypii: I have updated the code to include an option `repair` which defaults to true, but can be disabled. I felt it was important to default to true, since otherwise many examples would fail to extrude (including the birthday balloons). I hope that over time we can make the BSP tree really solid so that repair is not necessary.

packages/modeling/src/operations/extrusions/slice/repairSlice.js
hrgdavor: maybe have a log function as part of jscad internall that can be called, so such king logging could be caught and used in different ways. The code inside taht utility would disable linter(for the console log inside) because it is expected. 

We would then have ability to to log stuff but also have linter warn of unintended lefotever console calls that happen during testing stuff while developing.

hrgdavor (dismissed):
my comment about console logging is non blocking, so it looks good to move fwd after @z3dev checks.

packages/modeling/src/operations/extrusions/earcut/index.js
```diff
@@ -0,0 +1,247 @@
...
+  const triangles = []
+
+  if (!outerNode || outerNode.next === outerNode.prev) return triangles
+
+  let minX, minY, maxX, maxY, invSize
+
+  if (hasHoles) outerNode = eliminateHoles(data, holeIndices, outerNode, dim)
+
+  // if the shape is not too simple, we'll use z-order curve hash later; calculate polygon bbox
+  if (data.length > 80 * dim) {
```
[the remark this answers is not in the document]
platypii: You would have to ask the original authors why exactly they choose that. But basically if the number of points is greater than 80, then calculate the bounding box, and use the more complicated `isEarHashed` function. For small polys, use the slower, but simpler `isEar` function. Also affects the calculation of z-order.

packages/modeling/src/operations/extrusions/earcut/eliminateHoles.js
```diff
@@ -0,0 +1,128 @@
...
+  const bridge = findHoleBridge(hole, outerNode)
+  if (!bridge) {
+    return outerNode
+  }
+
+  const bridgeReverse = splitPolygon(bridge, hole)
+
+  // filter colinear points around the cuts
+  const filteredBridge = filterPoints(bridge, bridge.next)
+  filterPoints(bridgeReverse, bridgeReverse.next)
```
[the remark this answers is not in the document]
platypii: Not sure. It's from the original implementation, but I don't see any mention of it in the source paper [[1]](https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf). It seems to predate the javascript implementation and comes from the prior C implementation.

platypii:
@z3dev:
> So... we use internally for 'capping' extrusions. Any other use? Will users be interested in the earcut functionality?

Today the only use of `slice.toPolygons` is the extrude family of functions. Not sure what else would use it, I could imagine it being useful in other contexts though. Anytime you need to convert a non-convex path2 or geom2 into a set of convex polys, to make it into a face for a geom3. It is available via `slice.toPolygons` if a user wants.

packages/modeling/src/operations/extrusions/earcut/linkedListSort.js
```diff
@@ -0,0 +1,54 @@
...
+      for (i = 0; i < inSize; i++) {
+        pSize++
+        q = q.nextZ
+        if (!q) break
+      }
+
+      let qSize = inSize
+
+      while (pSize > 0 || (qSize > 0 && q)) {
+        if (pSize !== 0 && (qSize === 0 || !q || p.z <= q.z)) {
```
z3dev, line 27: Is this the comparison of elements? Is there anyway to break the comparison out into a function?

packages/modeling/src/operations/extrusions/slice/repairSlice.js
z3dev: Ok. Then let’s give a little more context on what’s being repaired.

Should this use console.warning()?

@hrgdavor has a nice idea, but that functionality should be added via a separate PR, which might generate  a little discussion as well. ;)

packages/modeling/src/operations/extrusions/earcut/polygonTrees.js
```diff
@@ -0,0 +1,64 @@
...
+const vec3 = require('../../../maths/vec3')
+const calculatePlane = require('../slice/calculatePlane')
+const assignHoles = require('./assignHoles')
+
+/*
+ * This class maps a 3D polygon onto a 2D space using an orthonormal basis.
+ * It tracks the mapping so that points can be reversed back to 3D losslessly.
+ * Also constructs a polygon hierarchy which associates holes with their outer solids.
+ */
+class PolygonTrees {
```
z3dev, line 13: Still having trouble swallowing this class name… not sure there’s a tree here… more like what the comment says… hierarchy

z3dev: P.S. I hate name choosing

z3dev (changes requested):
@platypii looking good. Just a couple of small nits. Sorry.

Also, please review the generated API documentation from ‘npm run docs’. I suspect there will be a coupled of unwanted global functions.

packages/modeling/src/operations/extrusions/earcut/linkedListSort.js
platypii: done, add a sorting function parameter

packages/modeling/src/operations/extrusions/earcut/polygonTrees.js
platypii: Changed to PolygonHierarchy

platypii:
@z3dev updated based on your feedback

packages/modeling/src/operations/extrusions/earcut/linkedPolygon.js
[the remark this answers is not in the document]
z3dev: It’s a question about naming… but fine

z3dev (approved):
@platypii this looks fine now. If there are any last minute changes then now is the time.

I will probably release this on Sunday JST

t1u1:
@z3dev Can you please make a release based on this? If there is a release I will finally be able to test this in production.

t1u1:
Confirming that the latest release is working great in my webapp! Thanks and bravo @platypii

:100: :+1:

PR #1028, opened 2022-03-11
Geometry validation tests

platypii:
This is the first of a series of work to improve the jscad geometries. The first step is to determine where there are bugs in the geometry to fix. So I wrote a `geom3.validate` function to run various tests against a geometry and throw an exception if errors, gaps, duplicate vertices, or non-manifold edges are found. Then I added validation to all the unit tests.

However what's interesting is that if you validate geometries from all the tests, **you find 50 non-manifold geometries**! This is my target set for bug fixing. But we want the tests to still pass now. So in this PR, for every test that failed to validate, I skipped failing validation tests. This serves as a reminder that a test failed to validate and should be fixed in the future. You can find all such examples using `grep notThrows.skip -rn --include=*.test.js packages/modeling`

After merging this PR, the game begins to replace the remaining `notThrows.skip` calls with `notThrows` by fixing the modeling operations. For example, the earcut branch fixes 21 of these tests, bringing the number of non-manifold geometries down from 71 to 50. I have more fixes in the pipeline that were found thanks to these validation tests.

### All Submissions:


z3dev:
@platypii just an FYI, it has never been a goal to create water tight geometry internally, as working in floating point math has all kinds of pitfalls. Having said that, better algorithms to produce better geometry is always a goal.

platypii:
@z3dev I agree it will not be possible to be perfect in all situations when using floating point. But many times I've seen that a glitch happens exactly at the location where there is a discontinuity in the geometry (#907). If we can make the inputs to BSP trees better quality, then it will be less likely to create a glitch. Some of the most common primitives actually have gaps in their geometry. And that's something we can totally improve.

This PR does not change any existing API code paths. It adds `validate` function and many more tests. If you have other ideas how you'd like to see this done, please let me know. But I think this is a very exciting change, and ready for review.

z3dev:
First up… I like the validate() function, and having the tests call validate is also an improvement. But I’m not sure why some tests added isA() rather than validate().

Second… I really hate functions that say returns true/ false for the return, and then throw internally.  So, I’m wondering if the validate functions should throw.

If throw is the expectation then it should be documented up front and in bold. This is a very special case.

If boolean then let the caller handle the error. For example, AVA will report the test failure.

Also, to throw or not could be configurable.

platypii:
> But I’m not sure why some tests added isA() rather than validate().

Some of the tests failed to `validate` because of geometry errors. For those tests I used `isA` instead as a less strict test. See PR description.

> So, I’m wondering if the validate functions should throw.

The exceptions help to understand exactly what caused the validation error. If it just returned false it would be really hard to debug the underlying geometry. So I think it is important to throw not just return false. I just added doc comments about throwing, in **bold**. 

I could remove the `return true` statement at the end. But I left it there so that it could still be included as an ava `t.true(validate(geometry))` statement. If it doesn't return true then I'm not sure what the tests should look like?

platypii:
It might help to understand some of my design choices if I explain how I use this validation tool to fix geometry bugs.

# Step 1

Get a list of examples which did not validate:

```bash
grep geom3.isA -rn --include=*.test.js packages/modeling
```

```
operations/expansions/expand.test.js:122:  t.true(geom3.isA(obs))
operations/expansions/expand.test.js:130:  t.true(geom3.isA(obs2))
operations/hulls/hullChain.test.js:78:  t.true(geom3.isA(obs))
operations/hulls/hullChain.test.js:85:  t.true(geom3.isA(obs))
operations/hulls/hull.test.js:220:  t.true(geom3.isA(obs))
operations/booleans/intersect.test.js:137:  t.true(geom3.isA(result1))
operations/booleans/intersect.test.js:182:  t.true(geom3.isA(result4))
operations/booleans/subtract.test.js:140:  t.true(geom3.isA(result1))
operations/booleans/subtract.test.js:149:  t.true(geom3.isA(result2))
operations/booleans/subtract.test.js:171:  t.true(geom3.isA(result3))
operations/booleans/union.test.js:151:  t.true(geom3.isA(result1))
operations/booleans/union.test.js:159:  t.true(geom3.isA(result2))
operations/booleans/union.test.js:187:  t.true(geom3.isA(result3))
operations/booleans/scission.test.js:23:  t.true(geom3.isA(result2[1][0]))
operations/booleans/scission.test.js:25:  t.true(geom3.isA(result2[2][0]))
operations/booleans/scission.test.js:43:  t.true(geom3.isA(result1[0]))
operations/booleans/scission.test.js:44:  t.true(geom3.isA(result1[1]))
operations/extrusions/extrudeFromSlices.test.js:91:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeFromSlices.test.js:109:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRectangular.test.js:68:  t.true(geom3.isA(obs))
operations/extrusions/extrudeRotate.test.js:14:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:64:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:75:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:86:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:91:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:98:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:105:  t.true(geom3.isA(geometry3))
operations/extrusions/extrudeRotate.test.js:125:  t.true(geom3.isA(obs))
operations/extrusions/extrudeRotate.test.js:154:  t.true(geom3.isA(obs))
primitives/cylinderElliptic.test.js:13:  t.true(geom3.isA(obs))
primitives/cylinderElliptic.test.js:72:  t.true(geom3.isA(obs))
primitives/cylinderElliptic.test.js:130:  t.true(geom3.isA(obs))
primitives/cylinderElliptic.test.js:146:  t.true(geom3.isA(obs))
primitives/cylinder.test.js:13:  t.true(geom3.isA(obs))
primitives/cylinder.test.js:43:  t.true(geom3.isA(obs))
primitives/torus.test.js:13:  t.true(geom3.isA(obs))
primitives/torus.test.js:18:  t.true(geom3.isA(obs))
primitives/torus.test.js:25:  t.true(geom3.isA(obs))
primitives/torus.test.js:30:  t.true(geom3.isA(obs))
primitives/torus.test.js:51:  t.true(geom3.isA(obs))
primitives/sphere.test.js:13:  t.true(geom3.isA(obs))
primitives/sphere.test.js:22:  t.true(geom3.isA(obs))
primitives/sphere.test.js:130:  t.true(geom3.isA(obs))
primitives/geodesicSphere.test.js:51:  t.true(geom3.isA(obs))
primitives/roundedCylinder.test.js:13:  t.true(geom3.isA(obs))
primitives/roundedCylinder.test.js:24:  t.true(geom3.isA(obs))
primitives/roundedCylinder.test.js:83:  t.true(geom3.isA(obs))
primitives/roundedCylinder.test.js:113:  t.true(geom3.isA(obs))
primitives/roundedCylinder.test.js:143:  t.true(geom3.isA(obs))
primitives/ellipsoid.test.js:13:  t.true(geom3.isA(obs))
primitives/ellipsoid.test.js:143:  t.true(geom3.isA(obs))
primitives/ellipsoid.test.js:150:  t.true(geom3.isA(obs))
```

# Step 2

Pick an example to investigate. In this case, I want to look at `cylinder.test.js`, because cylinders are a very important primitive used extremely often in designs.

So I edit `cylinder.test.js` on line 13 to use `validate` instead of `isA`:

```diff
-  t.true(geom3.isA(obs))
+  t.true(geom3.validate(obs))
```

# Step 3

Run `npm test` to see if the cylinder validates, and if not, why not?

```
npm test
```

The validate function throws an exception which explains that there are non-manifold edges, and which ones are bad:

```
@jscad/modeling:   primitives › cylinder › cylinder (defaults)
@jscad/modeling:   src/primitives/cylinder.test.js:13
@jscad/modeling:    12:                              
@jscad/modeling:    13:   t.true(geom3.validate(obs))
@jscad/modeling:    14:   t.is(pts.length, 96)       
@jscad/modeling:   Error thrown in test:
@jscad/modeling:   Error {
@jscad/modeling:     message: `non-manifold edges 6␊
@jscad/modeling:     1,0,-1 -> 0,0,-1␊
@jscad/modeling:     1,0,1 -> 1,0,-1␊
@jscad/modeling:     0,0,1 -> 1,0,1␊
@jscad/modeling:     0,0,-1 -> 1,-2.4492935982947064e-16,-1␊
@jscad/modeling:     1,-2.4492935982947064e-16,-1 -> 1,-2.4492935982947064e-16,1␊
@jscad/modeling:     1,-2.4492935982947064e-16,1 -> 0,0,1`,
@jscad/modeling:   }
@jscad/modeling:   › 1,-2.4492935982947064e-16,1 -> 0,0,1
@jscad/modeling:   › validateManifold (src/geometries/geom3/validate.js:50:11)
@jscad/modeling:   › Object.validate (src/geometries/geom3/validate.js:21:3)
@jscad/modeling:   › src/primitives/cylinder.test.js:13:16
@jscad/modeling:   ─
@jscad/modeling:   1 test failed
```

# Step 4

Investigate why the edges don't match up. In this case if you look closely at the edges you'll see that there is a wedge missing from the cylinder:

cylinder

# Step 5

Trace the code to figure out why there is a gap. In this case I found that `Math.sin(Math.PI * 2) !== 0` and so the ends of the cylinder were not matching up. I was able to fix this by adding a check for this case in `cylinderElliptic.js`. I have a pending fix which can seen [here](https://github.com/platypii/OpenJSCAD.org/commit/33a3b4b48dc7c100c73867829128cdb89bd5bada#diff-5673d2f80457968ad4d4d2562f7c47f1bec18496fa39231c26a72e4f8248f023).

z3dev:
> > But I’m not sure why some tests added isA() rather than validate().
> 
> Some of the tests failed to `validate` because of 

> > But I’m not sure why some tests added isA() rather than validate().
> 
> Some of the tests failed to `validate` because of geometry errors. For those tests I used `isA` instead as a less strict test. See PR description.

Yup. In those cases, you can do t.fails()
And AVA will continue, and report that a test is still failing

platypii:
@z3dev that is interesting, I did not know about ava's `failing` modifier before. I think that would be the ideal solution, but here is the problem I have. The `failing` modifier exists on the ava `test` object, not a specific `t` instance of the test. So for the cylinder example, I would have to put failing on the whole test:

```js
test.failing('cylinder (defaults)', (t) => {
  const obs = cylinder()
  const pts = geom3.toPoints(obs)

  t.true(geom3.validate(obs))
  t.is(pts.length, 96)
})
```

Which means that it won't do any of the other checks, like the `pts.length` check. If I applied this same `failing` pattern to `intersect.test.js` the result would be ALL the intersect tests were marked as failing, and there would be effectively NO intersect tests. If someone changed code and caused a regression, we wouldn't know it because the test is "expected" to be failing.

It would be nice if it could be scoped to just one line, but I don't see a way to do that. Any ideas?

platypii:
Replying to myself :-) What do you think of this solution @z3dev

```js
  t.notThrows.skip(() => geom3.validate(obs))
```

And then later we can remove the `.skip`

z3dev:
> Replying to myself :-) What do you think of this solution @z3dev
> 
> ```js
>   t.notThrows.skip(() => geom3.validate(obs))
> ```
> 
> And then later we can remove the `.skip`

That's cool.
I was thinking that any function relying upon the booleans would have failing tests. So, this would flag those pretty clearly.

platypii:
@z3dev check out the latest changes I think you'll be happy. I got rid of the `return true` statement, and instead use ava `t.notThrows` everywhere. In the cases where validation fails, it uses `t.notThrows.skip`. I think this is the best option for now, thanks for your feedback.

z3dev:
@platypii these changes look fine now. however, i still wonder why some tests added isA() rather than validate(). wouldn't all tests be better if validate() is called?

platypii:
@z3dev the remaining `isA` calls are all `path2.isA` and `geom2.isA`. In other words, I left the `isA` tests for anything that wasn't a `geom3`. This is leftover from the refactors in this PR. But it is also due to the fact that I did not create `path2.validate` or `geom2.validate` functions. I think there a few options, I am happy to do any of them based on your feedback:

1. Leave the `path2.isA` and `geom2.isA` tests in this PR as it is now.
2. Remove the `path2.isA` and `geom2.isA` tests from this PR.
3. Implement `path2.validate` and `geom2.validate` functions to make it consistent with `geom3`. The only reason I didn't do this was because the PR was long enough already, and I didn't know what to validate for other geometry types. I am happy to add simple placeholder implementations if you think that's better.

z3dev:
@platypii yeah. this is fairly large, so focusing on geom3 is fine. but let's open a new issue for geom2 and path2 validation so we don't forget. i do know that one of the validations for geom2 is running geom2.getOutlines()

z3dev (approved):
@platypii thanks for the super set of test improvements.

PR #1031, opened 2022-03-17
Webeditor styling and example loading fix

### LINKED ISSUE #1030: openjscad.xyz / Browsereditor line numbers overlapping
## Expected Behavior
Screenshot 2022-03-16 at 21 27 53

## Actual Behavior
Screenshot 2022-03-16 at 21 19 58

  1. Goto https://openjscad.xyz/
  2. Open Editor
  3. Input code

## Specifications

  - Version: Current running website
  - Platform: MacOS
  - Environment: browser (Brave, Safari)

I am pretty sure that  this also happens in other browsers. I have a fix on my local setup and would submit a PR. Just wanted to be sure that this is not just an issue I have.

NilsKrause:
* Fixes loading of examples in the editor only after clicking into it.
* Fixes the style of the editor no being applied without loading an example.
* Updates the CodeMirror dependency.

I took the freedom to delete the unused file editor.js and renamed the existing and used editor2.js to editor.js.

fixes #1030 (maybe more, but quick look over issues did not bring anything to my attention)


z3dev:
@NilsKrause thanks for the fixes! i'll try these out this weekend.

platypii (changes requested):
Thank you @NilsKrause this is fantastic! The editor is super important for educating new users about how to use JSCAD, and this will certainly help.

I tested this fix locally and can confirm that it fixes the two bugs: line wrapping, and blank editor. :+1: 

The only comment I have: there's still an extra scrollbar on my editor. It is not the fault of this PR. But if we're fixing the editor anyway, maybe we could fix the double scrollbars too? Not trying to create additional work. My suggestion would be change `overflow: scroll` to `overflow: auto` in the `section` styles in [demo.css](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/web/css/demo.css#L256) line 256. 

This problem is less obvious on mac, because the scroll bars auto hide, but on linux chrome its pretty bad:
scrollbars

NilsKrause:
Don't worry about 'additional' work 😄  it really isn't much and you're absolutely right. I didn't notice them on my mac, but in your screenshot it's not very aesthetic.

I'll look into it on the weekend.

If there are any more webeditor related issues we could include in the PR let me know.

platypii:
This might be a question for @z3dev but we have almost-duplicate code in `packages/web/css/editor.css` and `packages/web/css/codemirror.css` I wonder what the difference is? Is one dead code? Both seem to be referenced in different places.

Also, does the CSS need to be updated at the same time as the codemirror js version?

z3dev:
> This might be a question for @z3dev but we have almost-duplicate code in `packages/web/css/editor.css` and `packages/web/css/codemirror.css` I wonder what the difference is? Is one dead code? Both seem to be referenced in different places.

The whole CSS needs a review, pruning, and improving. There’s probably a lot of junk left over from V1.

> Also, does the CSS need to be updated at the same time as the codemirror js version?

I would suspect so, as that was the original as provided with CodeMirror.

platypii (approved):

platypii:
Tested the new changes locally, and can confirm that it fixed the scrollbar issue on linux. Cleaning up the duplicated code looks good to me too.

Thanks for a killer first contribution @NilsKrause !

z3dev (approved):
@NilsKrause super changes. thanks.

fyi, i tested on MacOS and IOS, and did not find any issues with the changes. everything worked well. :)

PR #1034, opened 2022-03-20
fix(modeling): fix areAllShapesTheSameType when types are different

platypii:
The function `areAllShapesTheSameType` does not return false in the case of passing different types.

The reason is: `return false` is called inside a `forEach`, and so it just returns from the inner anonymous function, not the outer one. I fixed it by using a for loop instead. I also added a test for this case.

This helps improve error handling in a case such as:

```js
union(square(), cube())
```

Without this fix, you could get a not-very-helpful error message of:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'map')
```

With this fix, you get the more useful message:

```
Uncaught Error: only unions of the same type are supported
```

### All Submissions:


z3dev (approved):
UG... thanks for the fixes. In recent tests, I actually hit this check. I wonder why no one has noticed.

PR #1036, opened 2022-03-21
New 3MF serializer

z3dev:
I had this sitting around since early V2, so decided to bring the code up to date, and publish. I've been using PrusaSlicer for testing the 3MF files, which seem to work really well.

Also, the compression of the 3MF package is completed by a small package called fflate, which @hrgdavor introduced. It worked first time. :)

### All Submissions:


z3dev:
3mf-03
3mf-02

packages/io/3mf-serializer/src/index.js
```diff
@@ -0,0 +1,256 @@
...
+ * @returns {Array} serialized contents, 3MF contents (XML) or 3MF packaging (ZIP)
+ * @example
+ * const geometry = primitives.cube()
+ * const package = serializer({unit: 'meter'}, geometry) // 3MF package, ZIP format
+ */
+const serialize = (options, ...objects) => {
+  const defaults = {
+    unit: 'millimeter', // micron, millimeter, centimeter, inch, foot, meter
+    metadata: true,
+    defaultcolor: [0, 0, 0, 1], // BLACK
```
platypii, line 64:
```suggestion
    defaultcolor: [0, 0.6, 1, 1], // blue
```

Thoughts on using blue instead of black as the default? It would match the jscad style more closely. I'm sure it depends on the viewer, but some viewers don't show pure black very well:

def-black

Versus blue:

def-blue

hrgdavor (approved):
3mf is a nice format, but unfortunaetly underused, and programs like PrusaSlicer only implement have a basic implementation (I am still angry from failing to convince them to support instancing from that is part of 3mf format). 

code looks good go for it :)

packages/io/3mf-serializer/src/index.js
z3dev: Hmm... 3MF specification says that 'displaycolor' is required, so RGBA has to be set. I'm not a fan of black but it does follow SVG. Hmm...

z3dev: I changed to the default color to JSCAD orange

platypii (approved):

PR #1038, opened 2022-03-23
fix(modeling): fix assignHoles function when there are nested holes

platypii:
When your validation PR reveals a bug in your own previous earcut PR.
spidermen

I was investigating the "skipped" validation tests, specifically the test on `extrudeRectangular.test.js:68` and realized that it was producing invalid geometry because of a bug in my earcut code. In the case where there were polygons with a hole nested inside another polygon hole, there was a bug. The test should have rendered as:
holes2

But it actually looked like this:
holes1

I traced the bug to the `assignHoles` function I wrote. I wrote a test for this case, and fixed the bug. I think it should be good now. One less bug, and one more valid geometry. :-)

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/extrusions/earcut/assignHoles.js
+++ b/packages/modeling/src/operations/extrusions/earcut/assignHoles.js
@@ -31,6 +31,7 @@ const assignHoles = (geometry) => {
       solids.push(i)
     }
   })
+
   // for each hole, determine what solids it is inside of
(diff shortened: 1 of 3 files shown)


hrgdavor:
Have not even thought about that type of situation. Nice catch.

z3dev (approved):
@platypii thanks for fixing thyself

Issue #1005, opened 2022-02-19
Not possible to draw a perfect cone using cylinderElliptic


samuelwang48:
## Expected Behavior

Since cylinder doesn't take two sets of arguments to render a cone, cylinderElliptic becomes an option to render a cone.

## Actual Behavior

cylinderElliptic() doesn't accept [0, 0] as a valid value for radius. This seems acceptable in OpenSCAD.

```
const jscad = require('@jscad/modeling');
const CssColor = jscad.colors.cssColors;
const Colors = jscad.colors;
const Primitive = jscad.primitives;

function main() {
  return Colors.colorize(CssColor.red, Primitive.cylinderElliptic({
    center: [0, 0, 0],
    height: 3,
    startRadius: [1, 1],
    endRadius: [0, 0]
  }));
}

module.exports = { main }
```

## Steps to Reproduce the Problem

  1. use the code snippet provided in codemirror in the UI
  2. SHIFT+ENTER
  3. Uncaught Error: bad output from script: expected geom3/geom2/path2 objects

## Specifications

  - Version: production
  - Platform: all
  - Environment: browser

z3dev:
True, but the radius can of on cap can be any floating above zero, like 0.0001

samuelwang48:
@z3dev thanks yes that's what I am using for now. Not sure this is an issue or not. If not please close the issue.

platypii:
Is there a reason that constraint is in place? When I change [cylinderElliptic.js:43](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/modeling/src/primitives/cylinderElliptic.js#L43) to allow zero, things appear to still work.

```diff
-  if (!endRadius.every((n) => n > 0)) throw new Error('endRadius values must be greater than zero')
+  if (!endRadius.every((n) => n >= 0)) throw new Error('endRadius values must be greater than zero')
```

![Screenshot at 2022-02-19 18-25-50](https://user-images.githubusercontent.com/1766297/154825974-b0cf897b-4889-4d31-bf4e-763d80bbbc73.png)

paulftw:
> Is there a reason that constraint is in place?
Doesn't make sense to me either. If it passes the tests why don't we change to `>=`? both end and start radius would need this change.

platypii:
[...]

PR #1039, opened 2022-03-25
fix(modeling): cylinder enhancements

platypii:
This PR relaxes some of the constraints on cylinderElliptic to allow radius of 0. It also includes fixes to the geometry construction to ensure that all cylinders are valid, manifold geometries.

# Cones

This PR allows `startRadius` and `endRadius` to be zero. Previously this was prevented by precondition checks. Relaxing the checks revealed additional bugs. I refined the if-checks on when to add certain faces. If the area of the cap is zero, then we shouldn't create any triangles there. But we should still create the side triangles. See #1005.

154825974-b0cf897b-4889-4d31-bf4e-763d80bbbc73

# Validation

Some of the cylinder validation tests were previously skipped because of invalid geometry (#1028). I started digging into the `cylinder` code, and found that the start was not matching perfectly the end of the cylinder rotation.

158036974-1ca43fd2-a55a-4b7a-abfd-60fc437d04da

As it iterates around the cylinder creating faces, it finishes with an angle of 2 * Pi. The problem comes from the fact that, in floating point, `Math.sin(2 * Math.PI) !== 0`.

Math.sin(Math.PI) => 1.2246467991473532e-16
Math.sin(2 * Math.PI) => -2.4492935982947064e-16
Math.cos(0.5 * Math.PI) => 6.123233995736766e-17
Math.cos(1.5 * Math.PI) => -1.8369701987210297e-16

If you `grep` for these numbers you'll see them show up all over the place in tests. So the question is what to do about it?

We could leave it, and accept it as unavoidable when working in floating point. The problem is that many of the algorithms used in jscad assume that their input is a closed geometry. When there are gaps, even microscopic ones, errors tend to be magnified when run through algorithms like BSP. A tiny gap was responsible for the artifacts in #907.

So how to fix it? I tested many options. In the end, I found the fastest and most accurate to be simply rounding to zero if the output is within epsilon of zero. From a performance point of view, it is hard to even measure a difference between this and just the plain Math functions.

The result of using these alternate `sin` and `cos` functions is that a huge number of tests become valid geometries. Maybe not as important, but it also cleans up a lot of near-zero values from the expected test outputs.

So the result is solid, closed geometries. The output is cleaner. It's not any slower. And users can create perfect cones.

### All Submissions:


z3dev:
@platypii these changes look fine. however, i'd like a mathematical reason for the choice of the constant in the rezero() function.

math constant EPS is the same value, but probably doesn't make sense as the values returned are from -1 to 1, which means that a smaller choice would probably be more appropriate.

From previous testing of floating point numbers, as related to normals, this is what I found...

```
// Normals are directional vectors with component values from 0 to 1.0, requiring specialized comparison
// This EPS is derived from a series of tests to determine the optimal precision for comparing coplanar polygons,
// as provided by the sphere primitive at high segmentation
// This EPS is for 64 bit Number values
const NEPS = 1e-13
```

i did a little running around on the net, and found this article about floating point calculations. it goes deep into the issues of rounding, etc. it might be a good read as well.

https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

platypii:
Very interesting link. Lots to learn about floating point.

> i'd like a mathematical reason for the choice of the constant in the rezero() function.

Good point. I choose `1e-5` arbitrarily, since it was the same as the `EPS` in `constants.js`. However you are right I think it should be smaller. Here is my evaluation of the options:

- `1e-5`. Same as EPS. The problem here is that it is a big enough number that it can be hit by semi-realistic geometries. For example, a cylinder with 1,000,000 segments will create duplicate vertices because of this rounding. A million is a lot of segments, but still shows the problem.
- `1e-13`. Same as NEPS. This seems like a good number to be consistent with the rest of the codebase, while still being more limited than EPS. I like this option.
- `3e-16`. The smallest number which still fixes the manifolds. Makes sense when you compare this to the output of `Math.sin(2 * Math.PI)` => `-2.4492935982947064e-16`
- `2.22e-16`. Same as Number.EPSILON. Too small, fails to fix the geometry errors. 

I just pushed a commit with change to use NEPS of `1e-13`. Let me know if you think it should be something else.

packages/modeling/src/utils/trigonometry.js
```diff
@@ -0,0 +1,25 @@
+
```
z3dev, line 1: Since this was added to user level utilities, then let’s expose the functions sin/cos in documentation.

z3dev (changes requested):
@platypii sorry. One last little detail

packages/modeling/src/utils/trigonometry.js
platypii: Added docs, types, and tests for these functions :+1:

z3dev (approved):
@platypii thanks again. cool improvements.

PR #1041, opened 2022-03-27
test(modeling): validation of 2D geometries

platypii:
This is the promised followup to the previous validation PR #1028. In that PR I added a `geom3.validate` and instrumented all the 3D geometry tests with it.

In this PR I do the same for 2D geometries, introducing `geom2.validate` and `path2.validate`. It checks for things like closedness, duplicate points, self-edges, transforms, nans, and infinities.

It turns out all the existing 2D geometry tests were already valid. That's awesome! But that doesn't mean that jscad has no bugs in 2D. So I added an example test which fails validation and is therefore skipped in `union.test.js`.

Yay tests :balloon: 

### All Submissions:


hrgdavor:
Cool. Thanks for the effort of improving the reliability of the api.

hrgdavor (approved):
looks good :)

z3dev (approved):
@platypii thanks again.

Issue #1032, opened 2022-03-17
Incorrect order of arguments in enclosedEdge (modling/src/operations/modifiers/edges.js)


briansturgill:
## Expected Behavior
This template doesn't really cover this issue.
I'm in process of translating jscad to C#.... the type checking caused me to find a clear problem in enclosedEdge.

The line3 method closestPoint has a signature of closestPoint(line, point).
In enclosedEdge (edges.js, lines 116 and 117) the calls to closestPoint clearly have the point and line arguments reversed.

Here are the relevant incorrect lines:
        const closest0 = vec3.snap(vec3.create(), eps, line3.closestPoint(openedge[0], line3d))
        const closest1 = vec3.snap(vec3.create(), eps, line3.closestPoint(openedge[1], line3d))
        
Tracing through the usage of enclosedEdges, only the generalize..."repair" unit test would exercise this code, and in this case, didn't
catch the problem.

Amazingly, this is the only problem I've found during the translation... and I'm more than halfway through!

## Steps to Reproduce the Problem

  1.
  2.
  3.

## Specifications

  - Version: Current
  - Platform: N/A
  - Environment: (browser, local server, node.js etc)

z3dev:
@briansturgill C#!!! Wow! That's pretty cool. I know that someone translated this to C++ once, long long ago.

We will look at this little 'mistake' in more detail, test, and fix.

platypii:
Steps to reproduce the problem:
```js
generalize({ repair: true }, union(cube({ size: 8 }), cube({ center: [0, 0, 4] })))
```

```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at Object.subtract (blob:https://openjscad.xyz/3d865e0e-a752-47e3-945a-f1ae9718dce7:2438:32)
    at Object.fromPoints (blob:https://openjscad.xyz/3d865e0e-a752-47e3-945a-f1ae9718dce7:2090:122)
    at enclosedEdge (blob:https://openjscad.xyz/3d865e0e-a752-47e3-945a-f1ae9718dce7:2774:1492)
```

z3dev:
Probably anything "repair" will die. It was a ambitious idea...

platypii:
> Probably anything "repair" will die.

Should I change #1042 to just delete all the generalize repair code? Happy to delete lines of code :fire:

PR #1042, opened 2022-03-29
fix(modeling): fix enclosedEdge function

platypii:
Fix for #1032. This PR includes:

- A test which explodes without this fix
- Fix the argument order when calling `line3.closestPoint` as found by @briansturgill
- Fix the line above that, which was calling `line3.fromPoints` with 2 arguments instead of 3
- Remove commented out `console.log` calls
- Renamed variables to use camelCase
- Minor comment cleanup (eg- spelling)

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/expansions/offsetFromPoints.js
+++ b/packages/modeling/src/operations/expansions/offsetFromPoints.js
@@ -111,7 +111,7 @@ const offsetFromPoints = (options, points) => {
         newPoints[i] = ip
         newPoints[(i + 1) % newPoints.length] = undefined
       } else {
-        // paralell segments, drop one
+        // parallel segments, drop one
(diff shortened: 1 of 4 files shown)


z3dev:
@platypii these changes look fine but this functionality may never work. i was trying to fix the T-junctions but this is not trivial, and as you noticed needs a whole lot more testing. although nice to have, i think that slicers don't care about T-junctions.

so, the question remains... should this be kept, and fixed later. i'm fine keeping this but it will be a lot of work later to make this rock solid.

platypii:
> so, the question remains... should this be kept, and fixed later.

I think we should close this PR in favor of #1049 which removes the code entirely. Even with these fixes, the repair function does not really work as intended. I suggest deleting it, and if we need it later, make a new PR with new working repair code.

Issue #1044, opened 2022-04-01
Another incorrect arguments issue.


briansturgill:
## Expected Behavior
That unit test depending on nearlyEquals actually work! :-)

## Actual Behavior
The missing "t," argument causes one important argument to nearlyEquals to be undefined and the test will always return true.

## Steps to Reproduce the Problem

  1.Don't bother.
  2.Just change the nearlyEquals calls to have missing "t, " as the first argument. (8 in each file)
  3.3 The two files with problems are: poly3/measureBoundingSphere.test.js and poly3/measureArea.test.js

## Specifications

  - Version:
  - Platform:
  - Environment: (browser, local server, node.js etc)

PR #1045, opened 2022-04-01
Require a test object as first argument to nearlyEqual

platypii:
Quick fix for #1044 

- Add a check to nearlyEqual to make sure that a test object is passed
- Fix tests that incorrectly called nearlyEqual
- Needed to increase epsilon a bit on one of the tests

Thanks @briansturgill nice catch!

### All Submissions:


z3dev (approved):
UG… I’m still not sure why the old version was working. Anyway, moving on with improvements.

platypii:
> I’m still not sure why the old version was working

It was "working" because it would get to the line `if (relative > epsilon) {` but epsilon was `undefined` so evaluated to false. Basically it said "the error is not larger than epsilon, must be okay"

PR #1046, opened 2022-04-01
fix(modeling): torus manifolds

platypii:
Continuing the JSCAD geometry improvements. Today we're fixing the Torii. 🍩 

torus non-manifold edges
<details>
<summary>torus-manifold.js</summary>

```js
const { colorize } = jscad.colors
const { geom3 } = jscad.geometries
const { cylinder, torus } = jscad.primitives
const { translate, rotate } = jscad.transforms

const main = () => {
  const obs = torus({})
  return showNonManifold(obs)
}

const showNonManifold = (obj) => {
  try {
    geom3.validate(obj)
    return obj
  } catch (e) {
    const edges = e.message.substring(e.message.indexOf('\n', 19) + 1)
      .split('\n')
      .map((e) => e.split(" -> ").map((v) => v.split(",").map(Number)))
    console.log('non-manifold edges', edges)
    return [
      colorize([0,.6,1,.9], obj),
      colorize([1,0,0], edges.map(([start, end]) => cylinderV1({ start, end, radius: 0.05 })))
    ]
  }
}

const cylinderV1 = (opt) => {
  const rot = [0, 0, 0]
  const center = [0, 0, 0]
  if (opt.start && opt.end) {
    center[0] = (opt.start[0] + opt.end[0]) / 2
    center[1] = (opt.start[1] + opt.end[1]) / 2
    center[2] = (opt.start[2] + opt.end[2]) / 2
    const dx = opt.end[0] - opt.start[0]
    const dy = opt.end[1] - opt.start[1]
    const dz = opt.end[2] - opt.start[2]
    opt.height = Math.hypot(dx, dy, dz)
    rot[1] = Math.atan2(dx / opt.height, dz / opt.height)
    rot[0] = Math.asin(-dy / opt.height)
  }
  return translate(center, rotate(rot, cylinder(opt)))
}

module.exports = { main }
```

</details>

This fix is slightly different than the others, because it is not using the util trig functions. It was necessary to have a special case here, because rounding doesn't help when `startAngle` is non-zero.

Yay for more water tight primitive geometries! 🍩 

### All Submissions:


packages/modeling/src/operations/extrusions/project.test.js
```diff
@@ -74,9 +75,8 @@ test('project (X and Y axis)', (t) => {
     [1.0000200000000001, -3.999986666666667],
     [0.9238600000000001, -4.382700000000001],
     [0.9807933333333334, -4.1951],
```
z3dev, line 77: Any idea why the order of points is changing? I cannot imagine that the order of points in the torus is changing.

packages/modeling/src/operations/extrusions/extrudeRotate.js
```diff
@@ -114,7 +114,11 @@ const extrudeRotate = (options, geometry) => {
 
   const matrix = mat4.create()
   const createSlice = (progress, index, base) => {
-    const Zrotation = rotationPerSlice * index + startAngle
+    let Zrotation = rotationPerSlice * index + startAngle
+    // fix rounding error when rotating 2 * PI radians
+    if (totalRotation === Math.PI * 2 && index === segments) {
```
z3dev, line 119: Probably very little performance impact but wondering if we could improve this.
- Add PI2 constant to maths/constants.js
- Set isFullRotation boolean above, and use here

z3dev (commented):
Just a few comments and questions.

packages/modeling/src/operations/extrusions/extrudeRotate.js
platypii: Pi was a mistake, it always should have been Tau. :-)
https://tauday.com/tau-manifesto

z3dev: > Pi was a mistake, it always should have been Tau. :-)
> https://tauday.com/tau-manifesto

😎 A very interesting debate with a lot of merit, but changing to Tau is like asking the USA to use metric measurements. 😃

packages/modeling/src/operations/extrusions/project.test.js
platypii: The torus points do not change order. This just reflects a change in the order of the `sides` of the geom2, the actual shape is the same.

The order of the projected sides change because in `project.js` the geometry is transformed by first a `unionGeom3` and then `unionGeom2`. This converts the geometries to BSP trees, and the output order is affected by very small changes in the input geometry. The shape itself is the same though. (+/- epsilon)

z3dev (approved):
@platypii thanks again.

z3dev:
@platypii this branch did not automatically update to the latest. please do the need full.

z3dev (approved):
Looks good!

PR #1048, opened 2022-04-03
performance improvements to modeling modifiers

z3dev:
These changes contain some performance improvements to the modifiers; mergePolygons and triangulatePolygons. The changes are pretty trivial, and eliminate unnecessary allocations.

Also, I have moved the retessellate functions from booleans to modifiers, as that's the basic functionality of those functions. retessellate can be integrated into 'generalize' later if useful.

Finally, the move of retessellate was done before making some additional changes. Basically, I want to get rid of retessellate as part of the booleans, and use mergePolygons instead. Retessellate modifies the polygons and calculates new points in some cases, which is BAD. mergePolygons reuses points and polygon edges, which is GOOD. this will be hugely important if the BSP trees can produce clean, water tight solids.

### All Submissions:


hrgdavor (commented):
Looks good, I was not aware there was an alternative for retesellate. Ofc any move to make booleans more robust is welcome :)

platypii (approved):
Very nice.

I wonder if there are more places in the codebase we can replace `*.fromPoints()` with `*.create()` to save on allocations 🤔  As long as we are immutable, there should be no need to copy points, right?

PR #1049, opened 2022-04-03
remove generalize repair function

platypii:
This is an alternative fix for #1032 instead of #1042 

Instead of fixing the broken generalize repair function, it just removes it.

The code was broken. In any case where it actually tried to repair a vertex, it would explode and throw an exception.

I can see the value of a repair function, but this one was broken, and clearly no users were depending on it, so I think it should just be removed for now. Open to suggestions, but I would vote for merging this PR, and closing #1042.

### All Submissions:


z3dev (approved):
Agreed. Broken code is bad in many ways, so let’s remove for now. It’s very simple to bring back as well.

Issue #1053, opened 2022-04-06
rewrite proposal hullGeom2.js/hullPoints2.js


hrgdavor:
## Expected Behavior
inc performance

## Specifications

  - Version: all
  - Platform: all
  - Environment: all

there is a proposal in discussion #1050  for change in implementation of hullGeom2.js/hullPoints2.js 

 It gets rid of the index vector and the use of Atan2 and is generally easier to understand... translation back to JS should be straight forward.

```js
namespace CSharpCAD;

public static partial class CSCAD
{
    private class pt
    {
        internal readonly Vec2 point;
        internal readonly float_t angle;
        internal readonly float_t distSq;
        internal pt(Vec2 point, float_t angle, float_t distSq)
        {
            this.point = point;
            this.angle = angle;
            this.distSq = distSq;
        }
    };
    /*
     * Create a convex hull of the given geom2 geometries.
     * Uses https://en.wikipedia.org/wiki/Graham_scan
     */
    public static Geom2 HullGeom2(params Geometry[] geometries)
    {
        // Extract unique points from the geometries.
        // To avoid a second pass, also determine the minimum point.
        var uniquePoints = new HashSet<Vec2>();
        var min = new Vec2(float_t.PositiveInfinity, float_t.PositiveInfinity);
        foreach (var g in geometries)
        {
            var sides = ((Geom2)g).ToSides();
            foreach (var side in sides)
            {
                var point = side.v0;
                uniquePoints.Add(point);
                if (point.y < min.y || (point.y == min.y && point.x < min.x))
                {
                    min = point;
                }
            }
        }

        // Returned "angle" is really 1/tan (inverse of slope) made negative to increase with angle.
        // This function is strictly for sorting in this algorithm.
        float_t fakeAtan2(float_t y, float_t x)
        {
            // The "if" is a special case for when the minimum vector found in loop above is present.
            // We need to insure that it sorts as the minimum point. Otherwise this becomes NaN.
            if (y == 0 && x == 0) { return float_t.NegativeInfinity; }
            return -(x / y);
        }

        // Gather information for sorting by polar coordinates.
        var points = new List<pt>(uniquePoints.Count);
        foreach (var v in uniquePoints)
        {
            // Use of fakeAtan2 avoids use of Math.Atan2 which slows things down.
            // A simple timing test suggests it saves about 10% of total time.
            var angle = fakeAtan2((v.y - min.y), (v.x - min.x));
[...]

PR #1055, opened 2022-04-07
refactor(modeling): new hullPoints2 implementation

platypii:
Backport the cleaned up hullPoints2 from @briansturgill. See #1050 and #1053 

It was a pretty straightforward port back to javascript. Some notes:

- Factored out `uniquePoints` function into modeling/utils
- Use shared `uniquePoints` in `hullGeom2` and `hullPath2`
- Rewrite `hullPoints2` using new structure from @briansturgill
- Although @briansturgill's algorithm combined the uniquePoints and find-min in one pass, I choose not to because it was harder to re-use code that way. I tried it as one pass and saw no measurable performance difference.
- Performance: no measurable difference between master and this PR

### All Submissions:


hrgdavor (approved):
Too bad it did not bring performance gains, but at least we have a nice reusable function for unique points.

packages/modeling/src/utils/uniquePoints.js
```diff
@@ -0,0 +1,33 @@
+const geom2 = require("../geometries/geom2");
```
z3dev, line 1: @platypii Do you really think that this will be used anywhere else? If not then probably this should just be part of the hulls.

Also, I prefer function names with 'verbs' but not so picky. If you come up with a nice name then please change.

packages/modeling/src/utils/uniquePoints.js
```diff
@@ -0,0 +1,33 @@
...
+  geometries.forEach((geometry) => {
+    if (geom2.isA(geometry)) {
+      geom2.toPoints(geometry).forEach(addPoint)
+    } else if (geom3.isA(geometry)) {
+      geom3.toPoints(geometry).forEach(addPoint)
+    } else if (path2.isA(geometry)) {
+      path2.toPoints(geometry).forEach(addPoint)
+    }
+  })
+
```
z3dev, line 29: Does it make sense to call found.clear() to release those strings back to garbage cleaning? I haven't seen any documentation on the benefits of clear() so just wondering.

packages/modeling/src/utils/uniquePoints.js
platypii: I think it might be useful elsewhere, but for now the only places I saw were in the hulls.
So I just moved it into `hulls` directory.

I also gave it the verb name `toUniquePoints`. Open to suggestions if you have a better name.

packages/modeling/src/utils/uniquePoints.js
platypii: The Set goes out of scope as soon as the function returns, allowing the keys to be freed. There should be no need to call clear().

packages/modeling/src/utils/uniquePoints.js
z3dev: @platypii thanks. yeah. let's move toUniquePoints() out of hulls if there's another use case. i'm still searching...

packages/modeling/src/operations/hulls/hullPoints2.js
```diff
@@ -1,108 +1,66 @@
...
-    j++
-  }
-  stack[0] = stack[points.length]
-  stack[1] = min
+  // gather information for sorting by polar coordinates
+  const points = []
+  uniquePoints.forEach((point) => {
+    // use faster fakeAtan2 instead of Math.atan2
+    const angle = fakeAtan2(point[1] - min[1], point[0] - min[0])
+    // const angle = Math.atan2(point[1] - min[1], point[0] - min[0])
```
z3dev, line 29: remove commented out code

packages/modeling/src/operations/hulls/hullPoints2.js
```diff
@@ -1,108 +1,66 @@
 const vec2 = require('../../maths/vec2')
 
-const angleBetweenPoints = (p0, p1) => Math.atan2((p1[1] - p0[1]), (p1[0] - p0[0]))
+// interface PolarPoint {
```
z3dev, line 3: Not sure what this comment adds. It doesn't explain anything about polar points or the contents.

z3dev (changes requested):
@platypii thanks for the port. just a couple of nits.

packages/modeling/src/operations/hulls/hullPoints2.js
platypii: That was mostly a note-to-self about what was in the `points` objects. I removed the commented code now. Good catch.

packages/modeling/src/operations/hulls/hullPoints2.js
platypii: done

z3dev:
@briansturgill thanks for the new implementation!

PR #1058, opened 2022-04-09
fix(modeling): reworked project to correct union of small polygons

z3dev:
These changes fix the issues with the project function. Previously, projections producing small polygons were causing the final union to fail, and produce invalid geometry.

### All Submissions:


packages/modeling/src/operations/extrusions/project.js
```diff
@@ -27,32 +26,30 @@ const projectGeom3 = (options, geometry) => {
...
-    projection = geom3.transform(rotation, projection)
+    projpolys = projpolys.map((p) => poly3.transform(rotation, p))
   }
 
-  // convert the projection (polygons) into a series of 2D geometry
-  const projections2D = geom3.toPolygons(projection).map((p) => geom2.fromPoints(poly3.toPoints(p)))
-  // union the 2D geometries to obtain the outline of the projection
-  projection = unionGeom2(projections2D)
+  // sort the polygons to allow the union to ignore small pieces efficiently
+  projpolys = projpolys.sort((a, b) => poly3.measureArea(b) - poly3.measureArea(a))
```
platypii, line 48: Why does sorting help here?

If a polygon is less than epsilonArea, it won't even be added to the `projpolys` list.

Is this to improve speed? Or quality of the geometry output? Was that found by trial and error?

z3dev: I found this by trial and error, and wondering... Why can't the BSP trees handle those little polygons?

Once the large polygons are processed first, then the smaller polygons are easily added to the BSP trees. The quality of the geometries improved significantly.

PR #1061, opened 2022-04-11
update of example code

rozek:
Hello!

First of all: thank you very much for your wonderful work!

However, it seems, that the documentation is not always correct - today, I needed the "SVG Deserializer" and found that the provided code did not work.

Here is the correct one.

With greetings from Germany,

Andreas Rozek


packages/io/svg-deserializer/README.md
```diff
@@ -36,10 +36,10 @@ npm install @jscad/svg-deserializer
 ## Usage
 
 ```javascript
-const svgDeserializer = require('@jscad/svg-deserializer').deserialize
+const svgDeserializer = require('@jscad/svg-deserializer')
 
 const rawData = fs.readFileSync('PATH/TO/file.svg')
-const geometry = svgDeserializer.deserialize(rawData, 'file.svg', {output: 'geometry'})
+const geometry = svgDeserializer.deserialize({ filename:'file.svg', output: 'geometry' }, rawdata)
```
platypii, line 42: Typo: `rawData` variable should be camelCase
```suggestion
const geometry = svgDeserializer.deserialize({ filename:'file.svg', output: 'geometry' }, rawData)
```

platypii (changes requested):

rozek (commented):
Done. Thank you for your remarks!

PR #1064, opened 2022-04-13
perf(modeling): faster toOutlines

platypii:
Performance improvements for `toOutlines`. This is used internally by `extrudeLinear`, `extrudeRectangular`, and `offsetGeom2`. Changes include:

- Use `Map` instead of `{}`
- Split up large `toOutlines` function into smaller `toSharedVertices`, `toVertexMap` and `popNextEdge` functions
- Remove unnecessary allocations of `vec3.create()`
- Remove unnecessary allocations in `geom3ToGeometries`

This results in performance improvements when running `toOutlines` on a circle:

1k segments: 2.5ms -> 2.1ms
10k segments: 18ms -> 14ms
50k segments: 92ms -> 72ms
100k segments: 172ms -> 125ms

### All Submissions:


packages/modeling/src/geometries/geom2/toOutlines.js
```diff
@@ -102,4 +92,28 @@ const toOutlines = (geometry) => {
   return outlines
 }
 
+// find the first counter-clockwise edge from startEdge and pop from nextEdges
+const popNextEdge = (startEdge, nextEdges) => {
+  if (nextEdges.length === 1) {
+    return nextEdges.pop()
+  }
+  const v0 = vec2.create()
```
z3dev, line 100: Isn’t this function creating a vertex for each invocation?

packages/modeling/src/geometries/geom2/toOutlines.js
```diff
@@ -64,31 +74,11 @@ const toOutlines = (geometry) => {
       if (!nextpossiblesides) {
         throw new Error('the given geometry is not closed. verify proper construction')
       }
-      let nextsideindex = -1
```
z3dev, line 67: Small request… in the throw above, can we add the vertex? That will provide a little more information for users.

z3dev (changes requested):
One question and one small request.

packages/modeling/src/geometries/geom2/toOutlines.js
platypii: added vertex info into the error. much better this way!

packages/modeling/src/geometries/geom2/toOutlines.js
platypii: Yes it is. Previously it was created only once per call to toOutlines, now it creates v0 for each step in the loop. Ideally that would not be necessary, but its hard to avoid when factoring out the `popNextSide` function.

I think it's still important to split out the `popNextSide` function even if it means an additional allocation. First, the code is more readable. Second, this is not the expensive part of toOutlines, so the impact is minimal. Third, breaking up large functions actually helps performance. I spend a lot of time in the chrome performance debugger looking at charts like the following. You can see that the v8 javascript engine will optimize certain functions. Making functions small and with narrow scope variables helps the compiler to be fast and efficient.
trace2

Open to suggestions, but I felt this was worth it.

z3dev: The function is faster now so there's no worry. I just wanted to make sure that the allocation was noticed.

z3dev (changes requested):
@platypii the toOutlines() changes look fine.
The other refactoring changes are find as well.

However, the change to slice.calculatePlane() looks like a separate performance change, so I would like this moved into another PR. I have another change planned for slice as well, as this needs to be moved to geomeotries. A small breaking change to the API.

z3dev (approved):
Ready to go…thanks @platypii

PR #1065, opened 2022-04-15
test(modeling): validate poly3 are coplanar

### LINKED ISSUE #1059: Possible flaw in poly3.validate
## Expected Behavior
Perhaps, poly3.validate should check to see that the vertices are coplanar?

  1.
  2.
  3.

## Specifications

  - Version:
  - Platform:
  - Environment: (browser, local server, node.js etc)

platypii:
Resolves #1059. This adds additional validation to poly3 to check that vertices are coplanar. Thanks @briansturgill!

The minimum error NEPS = 1e-13 was chosen because the minimum value needed for tests to pass was 4e-15. NEPS of 1e-13 keeps a tight bound, but still allowing reasonable floating point error.

Edit: Also moved NEPS into maths constants.

### All Submissions:


packages/modeling/src/geometries/poly3/validate.js
```diff
@@ -1,7 +1,11 @@
+const signedDistanceToPoint = require('../../maths/plane/signedDistanceToPoint')
 const vec3 = require('../../maths/vec3')
 const isA = require('./isA')
 const isConvex = require('./isConvex')
 const measureArea = require('./measureArea')
+const plane = require('./plane')
+
+const NEPS = 1e-13
```
z3dev, line 8: Maybe it's time to move this to maths/constants.js... we have multiple copies everywhere. :(

```
primitives/triangle.js:const NEPS = 1e-13
utils/trigonometry.js:const NEPS = 1e-13
maths/utils/aboutEqualNormals.js:const NEPS = 1e-13
operations/modifiers/mergePolygons.js:const NEPS = 1e-13
operations/modifiers/retessellate.js:const NEPS = 1e-13
```

z3dev (commented):
@platypii what do you think about defining NEPS in maths/constants?

we can renovate the other code later in another PR.

packages/modeling/src/geometries/poly3/validate.js
platypii: Moved NEPS into constants, and updated references to it :+1:

z3dev (dismissed):
this will be a nice addition to the validate function. thanks.

platypii:
@z3dev just realized I missed NEPS in the typescript file, pushed a commit with a fix

z3dev (approved):
Oh… just in time fix!

PR #1068, opened 2022-04-17
quality of life change for demo.html

hrgdavor:
This is very small quality of life change for the demo.html

I have removed `preventDefault` from mouseup and mousedown as it is note required for proper function of orbit controls.

I had to do this in new prototype to be able to have a functioning popup-menu. Without this, opening the menu and then clicking on the 3d view (canvas) the popup menu would not hide.


platypii (approved):
Good change. It is best practice not to call preventDefault unless needed.

Tested locally, looks good.

PR #1069, opened 2022-04-17
remove path2.eachPoint

### LINKED ISSUE #991: path2 eachPoint() has an unused argument in its signature
## Expected Behavior
If the "options" not in use, should it be deleted? It is currently being documented in the official doc.
```
const eachPoint = (options, thunk, path) => {
  toPoints(path).forEach(thunk)
}
```
## Actual Behavior
Shown in the doc and the source but not being used. 

  1. go to https://www.openjscad.xyz/docs/modeling_src_geometries_path2_eachPoint.js.html
  2. inspect the doc and the source

## Specifications

  - Version: production
  - Platform: all
  - Environment: all

platypii:
Resolves #991.

This PR removes `path2.eachPoint`. Here are my reasons for removal:

- Nothing in the JSCAD codebase uses `eachPoint`
- The same operation is trivially done with `toPoints(path).forEach` which uses native language features and is more clear to developers
- None of the other geometries define an eachPoint function
- Simplify the JSCAD API
- The function has an unnecessary `options` parameter (issue 991). To remove the parameter would be a breaking API change.
- If "fixing" the api would be a breaking change, and it's unlikely users depend on it anyway, why not just remove it?

### All Submissions:


z3dev (approved):
Clean and mean is far easier to maintain. Thanks.

PR #1062, opened 2022-04-12
style(web): css tweaks

platypii:
Minor clean up and fixes to CSS

- Remove unused big.css, min.css, openjscad.css files
- ~Remove unimplemented "file handling" option~
- Fix padding on "Load a JSCAD project" button
- Fix #controls and #header elements overflowing the page width
- Add semi-transparent background to option menus
- Fix padding of fieldsets in options menu
- Use max-height instead of height so that menus aren't larger than necessary
- Full height on code editor

Hopefully not controversial changes, just little UI improvements. Tested on chrome and firefox, desktop and mobile.

### All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [x] Does your submission pass tests?

PR #1071, opened 2022-04-18
fix(web): remove file handling option

platypii:
Removes the "File Handling" from the options menu.

This option is currently not implemented and the interface is broken. It starts as checked, but if you uncheck it, and then re-open the menu, it will be checked again. Furthermore, there does not appear to be any code that references the `convertSupportedTypes` parameter, so it seems like this option is not wired up to the code at all.

In #1062 @z3dev said:

> I'm not sure about the elimination of the "File Handling" option. I know that it's not used but could be in the future. Also, there are other options that should be removed as well, e.g. Storage path. So, let's do these as part of another PR.

If it is implemented in the future, it is easy to restore this code. In the meantime, I strongly feel that we should NOT give users an option that is broken or unimplemented.

I am happy to also remove the "storage path" option, but wanted to keep this PR focused on the file handling.

### All Submissions:


z3dev:
@platypii ok. Your reasoning is pretty much what I tell others as well.

let’s remove the file storage as well. It’s lame.

FYI, the file handling is for driving the IO library options. This setting should allow users to convert to JSCAD script or just geometry. But there are a lot of other options for the IO libraries, such as unit, etc. Those should be par of the choices as well. This is a far larger task, which I don’t want to take on with the current UI.

z3dev (approved):
Kind of a shame that these were never implemented. 😞

hrgdavor:
> FYI, the file handling is for driving the IO library options.

@z3dev maybe we should open a ticket or a discussion to  define those options for future reference and also so I can take them in consideration during the development of the new prototype. It is better to be aware of such features in advance to better guide choices during deveompment.

PR #1072, opened 2022-04-21
perf(modeling): replace poly3.fromPoints with poly3.create

platypii:
Inspired by changes made in #1048 by @z3dev.

There are two methods for creating a new poly3 from points: `poly3.create` and `poly3.fromPoints`. Create is strictly faster than fromPoints, because fromPoints clones the vec3's and allocates a new vertices list [[1]](https://github.com/jscad/OpenJSCAD.org/blob/02c740ce56c04ffb7b83a5e48009abb111c64001/packages/modeling/src/geometries/poly3/fromPoints.js#L21). Whereas `create` just uses the passed vertices list as-is.

As long as everything is treated as immutable in jscad, there should be no reason to have to clone these points. This saves not only the initial copy and free, but it also saves memory by sharing vertices in cases like the primitives. This saves on allocations and garbage collection.

I split this up into 3 commits: 1) the primitives 2) the tests and 3) everything else.

_Side note: it would be cool if we could get rid of one of these two functions. Maybe something for V3?_

### All Submissions:


hrgdavor (approved):
I agree with the changes. looks good to me. thanks for the continuous improvements.

z3dev:
@platypii the changes look fine, and I went through each win detail to make sure that vertices are unique. Everything looks good.

however, there’s a lot of change. So, please through every design at these changes. It’s hard to verify everything so a little extra vigilance is due here.

PR #1077, opened 2022-04-30
test(modeling): split boolean tests into 2d and 3d

platypii:
This PR refactors the boolean tests to split one large `union.test.js` into `unionGeom2.test.js` and `unionGeom3.test.js`. Same for intersect and subtract.

The reason for this is that first of all, union.test.js was already very long at almost 300 lines of code.

But also, I plan to add, like, a lot more tests.

So refactoring this is an incremental step to keep the PRs smaller. No tests have actually been changed. Just moved around for better organization.

### All Submissions:


z3dev (approved):
Sure. More testings should be encouraged. :)

PR #1078, opened 2022-04-30
Improve error handling

z3dev:
A little background on these changes.

The current error handing is kind of broken (See the first set of screen shots.) There are many reasons for the really poor messages, but basically errors are propagating from web-worker thread to main thread via a DOM event called [ErrorEvent](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent). The information propagated is incomplete or irrelevant. And of course, important information about the real error is lost immediately.

These changes do several things.

First, the core library was enhanced (not really API breaking changes) to catch all errors (both evaluation and runtime errors), and correctly report the error back via the callback. This is NEW functionality, and may break applications that are catching the DOM ErrorEvent.

Second, the webRequire function was enhanced to catch evaluation errors (see the Function call), enhance the error with file name, and create a specific stack message with context. It actually works!

Third, the WEB UI was enhanced to handle the error message properly in the design state. As well as display the error message in a beautiful new format. :)

Forth, there were a few little bug fixes to the WEB UI to report parameter definition errors properly.

**NOTE: This still isn't perfect, but improved in many ways.**

### All Submissions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,0 @@
-language: node_js
-node_js:
-  - "16"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: node_js
+node_js:
+  - "16"
--- a/packages/core/src/code-evaluation/rebuildGeometry.js
+++ b/packages/core/src/code-evaluation/rebuildGeometry.js
@@ -72,8 +72,8 @@ const rebuildSolids = (data, callback) => {
   } catch (error) {
     callback({
       type: 'errors',
-      name: error.name,
-      message: error.message,
+      name: error.name ? error.name : 'Error',
+      message: error.message ? error.message : error.toString(),
       description: error.description ? error.description : '',
       number: error.number ? error.number : '',
       fileName: error.fileName ? error.fileName : '',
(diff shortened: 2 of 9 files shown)


z3dev:
Syntax Error (single file)

old-syntax

Runtime Error (single file)

old-runtime

Syntax Error (project)


Parameter Definition Error

z3dev:
NEW AND IMPROVED!!!

Syntax Error (single file)

Runtime Error (single file)

Syntax Error (project)

Parameter Definition Error

platypii:
Can you share some of the "bad" model code that generated those errors? That would be helpful for testing.

When I run this branch locally and give invalid jscad files, I very frequently see things like:

> undefined: Uncaught Error: radius must be greater than zero

I'm not sure why it's prefixed with "undefined" but I see that in both chrome and firefox for various models (eg- `cylinder({ radius: "bad" })`)

platypii (dismissed):
Overall this appears to be a significant improvement.

In my testing, I did find one scenario where the error messages got "worse" than master though. If you make a model that throws a non-Error error, the output is less useful than before.

```js
function main() {
  throw "boom"
}
module.exports = { main }
```

Previously this would give the error:
```
ERROR:
Uncaught boom
Line: 4
```

On the `improve-error-handling` branch it now shows:
```
undefined: undefined
```

It seems like the new code is assuming that only `Error` can be thrown but that's not strictly true. Maybe there is an easy fix? I still think these are overall positive changes though.

z3dev:
@platypii nice catch. the MDN examples actually throw all kinds of objects, so i adjusted the code to create a message from the objects.toString(). That should suffice.

platypii (approved):
Looks good now :+1:

PR #1081, opened 2022-05-04
Nuts and Bolts example

platypii:
One of the most important mechanical building blocks are: bolts, nuts, screws, etc. Some of the most popular generative models on thingiverse are screw generators.

Have you ever tried making one in JSCAD? It's actually kind of tricky. So I took a shot at writing my own.

Initially I tried to use `extrudeLinear` with a `twistAngle` but that results in ugly and invalid geometry. The reason is that the wall faces get stretched as it rotates around. Instead I use `extrudeFromSlices` directly, and I carefully generate each slice by hand, so that the wall faces all line up.

I added this as an example model, so that hopefully users might find it helpful.

nutsandbolts

### All Submissions:

### DIFF
--- a/packages/examples/core/extrusions/nutsAndBolts.js
+++ b/packages/examples/core/extrusions/nutsAndBolts.js
@@ -1,0 +1,68 @@
+/**
+ * Nuts and Bolts
+ * @category Creating Shapes
+ * @skillLevel 8
+ * @description Demonstrating the advanced extrusion using slices to generate screw threads.
+ * @tags extrude, slice, slices, extrudefromslices, callback
+ * @authors platypii
+ * @licence MIT License
+ */
+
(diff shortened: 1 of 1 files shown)


z3dev (commented):
Fun! I like functional examples!

How about different colors for nut and bolt?
Does this get picked up automatically as part of the examples?

platypii:
@z3dev added color
nutsandbolts

Also refactored it to make it easier for users to adjust options for generating bolts of different sizes:
boltoptions

And yes, it does get picked up automatically as part of the examples:
boltexample

z3dev:
I think it's time for another port to JSCAD V2... This is #1 configurable model on Thingiverse.

https://www.thingiverse.com/thing:193647

z3dev (approved):
I like this version even more. Thanks @platypii

PR #1083, opened 2022-05-05
perf(modeling): use Map instead of {} for repairSlice

platypii:
Faster performance in `repairSlice` function by using a `Map` instead of an object `{}`.

This gives a ~18% performance improvement on 50k segment geometries inside the `repairSlice` (105ms -> 85ms).

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/extrusions/slice/repairSlice.js
+++ b/packages/modeling/src/operations/extrusions/slice/repairSlice.js
@@ -5,27 +5,36 @@ const vec3 = require('../../../maths/vec3')
  */
 const repairSlice = (slice) => {
   if (!slice.edges) return slice
-  const vertexMap = {} // string key to vertex map
-  const edgeCount = {} // count of (in - out) edges
+  const vertexMap = new Map() // string key to vertex map
+  const edgeCount = new Map() // count of (in - out) edges
+
+  // build vertex and edge count maps
   slice.edges.forEach((edge) => {
     const inKey = edge[0].toString()
     const outKey = edge[1].toString()
-    vertexMap[inKey] = edge[0]
-    vertexMap[outKey] = edge[1]
-    edgeCount[inKey] = (edgeCount[inKey] || 0) + 1 // in
-    edgeCount[outKey] = (edgeCount[outKey] || 0) - 1 // out
+    vertexMap.set(inKey, edge[0])
+    vertexMap.set(outKey, edge[1])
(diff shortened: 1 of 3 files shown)


packages/modeling/src/operations/extrusions/slice/repairSlice.js
```diff
@@ -5,42 +5,54 @@ const vec3 = require('../../../maths/vec3')
...
     missingOut.forEach((key2) => {
-      const v2 = vertexMap[key2]
+      const v2 = vertexMap.get(key2)
       const distance = Math.hypot(v1[0] - v2[0], v1[1] - v2[1])
       if (distance < bestDistance) {
         bestDistance = distance
         bestReplacement = v2
       }
     })
     console.warn(`repairSlice: repairing vertex gap ${v1} to ${bestReplacement} distance ${bestDistance}`)
```
z3dev, line 44: Why leave the console message? This could be very redundant if called repair is called a lot.

packages/modeling/src/operations/extrusions/slice/repairSlice.js
```diff
@@ -5,42 +5,54 @@ const vec3 = require('../../../maths/vec3')
  */
 const repairSlice = (slice) => {
```
z3dev, line 6: Why not just repair? If used with full name spacing then this would become...

`
extrutions.slice.repairSlice(myslice)
`

Looks a bit strange.

Also, slices should be immutable like other geometry. Is repair() being used somewhere?

z3dev (changes requested):

packages/modeling/src/operations/extrusions/slice/repairSlice.js
platypii: The repair slice name existed before this PR. But, I just pushed changes:

- rename `repairSlice` to `repair`
- made repair immutable

Repair is being used by all the extrude functions (linear, fromSlices, etc).

packages/modeling/src/operations/extrusions/slice/repairSlice.js
platypii: The console warning was discussed [here](https://github.com/jscad/OpenJSCAD.org/pull/1021#discussion_r820205467).

I like having the warning so that users can know that there was a change made to their geometry. Open to discussion of removing it though.

z3dev (approved):
@platypii thanks. let's go with these changes.

PR #1084, opened 2022-05-07
fix (modeling): ensure color is preserved in boolean operations

frading:
Hello,

I am currently adding jscad to [polygonjs](https://github.com/polygonjs/polygonjs), so that it would be possible to create csg operations via a node-based interface.
One problem I've hit is that some operations, like the boolean ones, do not preserve the poly3 color of the input geometries. This PR should fix this.

See screenshots for an illustration of the problem:

the boolean inputs geom3:
jscad_boolean_objectA
jscad_boolean_objectB

the result (without this PR) looses the color:
jscad_boolean_result_no_color

or with the PR, the color is preserved:
jscad_boolean_result_with_color

Let me know if you'd like me to adjust anything in this PR, or anything that makes it easier on your end.

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/poly3/invert.js
+++ b/packages/modeling/src/geometries/poly3/invert.js
@@ -15,7 +15,9 @@ const invert = (polygon) => {
     // Flip existing plane to save recompute
     inverted.plane = plane.flip(plane.create(), polygon.plane)
   }
-  inverted.color = polygon.color;
+  if(polygon.color) {
+    inverted.color = polygon.color;
+  }
   return inverted
 }
 
--- a/packages/modeling/src/geometries/poly3/invert.js
+++ b/packages/modeling/src/geometries/poly3/invert.js
@@ -15,6 +15,7 @@ const invert = (polygon) => {
     // Flip existing plane to save recompute
     inverted.plane = plane.flip(plane.create(), polygon.plane)
   }
+  inverted.color = polygon.color;
   return inverted
 }
 
--- a/packages/modeling/src/operations/modifiers/reTesselateCoplanarPolygons.js
+++ b/packages/modeling/src/operations/modifiers/reTesselateCoplanarPolygons.js
@@ -20,6 +20,7 @@ const reTesselateCoplanarPolygons = (sourcepolygons) => {
   const destpolygons = []
   const numpolygons = sourcepolygons.length
   const plane = poly3.plane(sourcepolygons[0])
+  const color = sourcepolygons[0].color
   const orthobasis = new OrthoNormalBasis(plane)
   const polygonvertices2d = [] // array of array of Vector2D
   const polygontopvertexindexes = [] // array of indexes of topmost vertex per polygon
@@ -307,6 +308,7 @@ const reTesselateCoplanarPolygons = (sourcepolygons) => {
           const points2d = prevpolygon.outpolygon.rightpoints.concat(prevpolygon.outpolygon.leftpoints)
           const vertices3d = points2d.map((point2d) => orthobasis.to3D(point2d))
           const polygon = poly3.fromPointsAndPlane(vertices3d, plane) // TODO support shared
+          polygon.color = color
 
           // if we let empty polygon out, next retesselate will crash
           if (polygon.vertices.length) destpolygons.push(polygon)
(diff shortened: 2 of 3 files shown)


z3dev:
@frading thanks for the PR.

I'm kind of split about these changes.

It's kind of cute but the question would be what happens next? Take that cube after the subtract and use it.. union it together with another shape... remove a few holes... and then what are the colors suppose to be?

Also, take that cube, move it, rotate it, and then use it again in union, subtract. OOPS! Those face colors are gone.

FYI, the focus of JSCAD is basically for 3D printing. And 3D printers don't care about faces with different colors. 3D printers can about meshes with different colors. That's why the support for colors at the geometry level is important, not poly3.

frading:
@z3dev thank you for your quick reply.

To answer your questions, ideally the colors should be propagated in every downstream operation. But I just saw a bug where they are not, so I'll improve this PR if you think this could potentially be interesting to merge it.

The reason I went down that route is that in the type definitions, [poly3 inherits from Colored](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/modeling/src/geometries/poly3/type.d.ts), so my understanding is that poly3 could carry colors. It may not be useful for 3D printers, but more for display before any export, or for debugging.

Curious to hear what your thoughts are based on this.

packages/modeling/src/geometries/poly3/invert.js
```diff
@@ -15,6 +15,7 @@ const invert = (polygon) => {
     // Flip existing plane to save recompute
     inverted.plane = plane.flip(plane.create(), polygon.plane)
   }
+  inverted.color = polygon.color;
```
hrgdavor, line 18: This should be conditional, so you dont attach undefined on all polygons

packages/modeling/src/operations/modifiers/reTesselateCoplanarPolygons.js
```diff
@@ -307,6 +308,7 @@ const reTesselateCoplanarPolygons = (sourcepolygons) => {
           const points2d = prevpolygon.outpolygon.rightpoints.concat(prevpolygon.outpolygon.leftpoints)
           const vertices3d = points2d.map((point2d) => orthobasis.to3D(point2d))
           const polygon = poly3.fromPointsAndPlane(vertices3d, plane) // TODO support shared
+          polygon.color = color
```
hrgdavor, line 312: This should be conditional, so you dont attach undefined on all polygons

hrgdavor:
@z3dev we have left forced colorPerVertex in regl shader, so I guess it something you intended to support. I am planing to make colorPerVertex optional at least in the rendering step(if during conversion no polygon has custom color). Maybe it would be good to test performance implications of allowing color preservation, and see if we can have it as optional to keep performance, and also allow ppl to use it if desired.

hrgdavor:
Another thing. Now that I looked at the code, I am suspecting that this is done ina mutable way, and could affect other operations unintentionally.

I would propose a new coloring function that applies color to all polygons like paint instead just defining default color of a whole mesh. But eqch polygon would need to be duplicated. This way booleans would copy color when splitting a polygon. Anyway, I am against making booleans slower, even if it means having a copy of boolean ops that jave builtin color preservation.

packages/modeling/src/geometries/poly3/invert.js
frading: Thank you, and agreed. I've just added a commit.

packages/modeling/src/operations/modifiers/reTesselateCoplanarPolygons.js
frading: Thank you, and agreed. I've just added a commit.

z3dev:
@frading lets backup a little...

These changes are changing the behavior of the booleans. It's probably a lot like half full versus half empty glasses of water... people have different opinions. Why does removing a hole leave different colors? Isn't the shape just one solid color? (My opinion)

Basically, we ask the user to add color AFTER completing all operations to create a complex shape. It makes everything very simple, and useful for 3D printing.

And there are all kinds of changes / decisions to make in other parts of the library.

FYI, I filled a new issue in order to fix the TS definitions.

frading:
@z3dev That's totally fair. I absolutely did not mean to suggest a change that would not fit with where you aim the library to go. (and thank you for opening the TS related issue, that will definitely help people like me who discover this library)

Thanks a ton for having taken the time to look into this on a weekend, I appreciate.

z3dev:
> @z3dev That's totally fair. I absolutely did not mean to suggest a change that would not fit with where you aim the library to go. (and thank you for opening the TS related issue, that will definitely help people like me who discover this library)
> 
> Thanks a ton for having taken the time to look into this on a weekend, I appreciate.

@frading No problem. We try very hard to understand all views. That's why I took a few days to think about this 'color' issue again. And who knows, someday JSCAD might be able to support some of the cool renderings possible via other libraries. If you'd like, please open a new issue or discussion to let us know how you want to use  / improve JSCAD.

PR #1085, opened 2022-05-08
perf(modeling): improved performance of reTesselateCoplanarPolygons by using maps or sets

z3dev:
These changes improve the performance of reTesselateCoplanarPolygons, which is used by the boolean operations underneath to optimize the polygon meshes.

I always have major issues with this code but here are some more changes. Hopefully, this improves the performance of the boolean operations. TO BE SEEN

FYI, there's about 97% coverage on this function via the test suites.

### All Submissions:


platypii (approved):
Changes look good.

Would be nice if these variables were camelCase (but that issue pre-dates this PR).

PR #1043, opened 2022-03-30
fix(modeling): ellipsoid manifolds

platypii:
This is the next series of geometry fixes, similar to the cylinder one (#1039), but instead focused on ellipsoids.

Because of rounding errors in the trig functions, there are microscopic gaps in the ellipsoid geometries. I made a script to visualize non-manifold edges as red cylinders inside JSCAD:
sphere-bad
<details>
<summary>manifold.js</summary>

```js
const jscad = require('@jscad/modeling')
const { geom3 } = jscad.geometries
const { translate, rotate } = jscad.transforms
const { cylinder, sphere } = jscad.primitives
const { colorize } = jscad.colors

const main = () => {
  const obs = sphere({ segments: 16 })
  return showNonManifold(obs)
}

const showNonManifold = (obj) => {
  try {
    geom3.validate(obj)
  } catch (e) {
    const edges = e.message.substring(e.message.indexOf('\n', 19) + 1)
      .split('\n')
      .map((e) => e.split(" -> ").map((v) => v.split(",").map(Number)))
      console.log('non-manifold edges', edges)
    return [
      colorize([0,.6,1,.9], obj),
      colorize([1,0,0], edges.map(([start, end]) => cylinderV1({ start, end, radius: 0.01 })))
    ]
  }
  return obj
}

const cylinderV1 = (opt) => {
  const rot = [0, 0, 0]
  const center = [0, 0, 0]
  if (opt.start && opt.end) {
    center[0] = (opt.start[0] + opt.end[0]) / 2
    center[1] = (opt.start[1] + opt.end[1]) / 2
    center[2] = (opt.start[2] + opt.end[2]) / 2
    const dx = opt.end[0] - opt.start[0]
    const dy = opt.end[1] - opt.start[1]
    const dz = opt.end[2] - opt.start[2]
    opt.height = Math.hypot(dx, dy, dz)
    rot[1] = Math.atan2(dx / opt.height, dz / opt.height)
    rot[0] = Math.asin(-dy / opt.height)
  }
  return translate(center, rotate(rot, cylinder(opt)))
}

module.exports = { main }
```

</details>

By switching to the util `sin` and `cos` functions, these gaps go away:
sphere-good

This results in manifold ellipsoid geometries. It also cleans up the tests a bunch, particularly the sphere expand test since a bunch of almost-identical edges are unified.

### All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
[...]

PR #1090, opened 2022-05-15
fix(modeling): use special trig functions for rotations

platypii:
This is the latest in a series of PRs to fix the geometries of primitive objects. See also #1028 #1039 #1041 #1043 #1046

This PR brings the number of non-manifold geometries in the tests down from 22 to 18.

Non-manifold geometries vs  Date

As in the previous cases, there are gaps in the generated geometries. They result directly from using `Math.sin` and `Math.cos` on approximations of 90 degree rotations.

For example, if you want to rotate a part 180 degrees, the way to do that in jscad is something like `rotateZ(Math.PI, obj)`. Except that produces a transform matrix of:
```
[-1, 1.2246467991473532e-16, -0, 0, -1.2246467991473532e-16, -1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]
```

Which is not exactly what was intended, because of the near-zero values. This can lead to cascading small errors as it is applied. It would be nice if I could rotate a part by exactly 90 or 180 degrees, a very simple and common operation, without introducing small errors.

The result of using the utility sin and cos functions is that a number of tests become valid geometries. Maybe not as important, but it also cleans up a lot of near-zero values from the expected test outputs. In fact, with use of new `sin` and `cos` functions, there are no more 1e-16 numbers in the primitive tests at all. This affects `arc`, `roundedCuboid`, and `extrudeRotate`.

extrudeRotate

### All Submissions:


z3dev:
@platypii this looks fine but I have a few concerns.

first, did we put those sin / cos functions into the correct utilities. It looks more like math utilities now.

second, although these functions help with calculations at specific PI rotation values, I'm wondering if this is actually correct. What happens at PI / 8, or PI * 0.10? Wouldn't a general 'rounding' be better?

platypii:
> first, did we put those sin / cos functions into the correct utilities. It looks more like math utilities now.

I'm fine with moving them if you like. I put them in `utils` because that is where functions like `degToRad` and `radToDeg` were located. Let me know if you'd like them moved.

> second, although these functions help with calculations at specific PI rotation values, I'm wondering if this is actually correct. What happens at PI / 8, or PI * 0.10? Wouldn't a general 'rounding' be better?

The other multiples of PI do not produce near-zero numbers, so the rounding doesn't really make sense on other values. For example, numbers that should go to 1 already work (because of floating point significant digits). PI/8 is transcendental so there's no benefit to rounding:
`Math.sin(Math.PI / 2) === 1`
`Math.sin(Math.PI / 4) === 0.7071067811865475`
`Math.sin(Math.PI / 8) === 0.3826834323650898`

z3dev:
@platypii sorry. let me try again.

the goal of these changes is improving the results, and insure that shapes are water-tight. i'm wondering if these changes are enough, give that starting and ending angles can be provided when creating several shapes. for example, if a cylinder is created with a different starting angle and ending angle which still produces a 2PI rotation.

platypii:
@z3dev ah now I see. Good point. In general, it's not just 90 degree angles that have floating point errors. When we are doing a rotation by a full 2 * Pi radians, we need to be careful to make sure that the ends line up, regardless of the startAngle.

```js
f = (x) => Math.sin(x) - Math.sin(x + Math.PI * 2)
```

If the math was perfect, then `f(x)` would always be zero. In reality, there is almost always a small error:

```js
f(0) == 2.4492935982947064e-16
f(0.1) == 5.967448757360216e-16
f(0.2) == 5.551115123125783e-17
f(0.3) == 3.885780586188048e-16
f(Math.PI) == -2.449293598294706e-16
f(Math.PI / 2) == 0
```

Let me go read the code, and maybe write some tests, to see if I can find any such bugs.

z3dev:
> > first, did we put those sin / cos functions into the correct utilities. It looks more like math utilities now.
> 
> I'm fine with moving them if you like. I put them in `utils` because that is where functions like `degToRad` and `radToDeg` were located. Let me know if you'd like them moved.

these are kind of borderline. Maths utilities should be for our internals, while general Utils should be for users. What do you think?

platypii:
> these are kind of borderline. Maths utilities should be for our internals, while general Utils should be for users. What do you think?

I agree. I moved them `utils` into `maths/utils`.

platypii:
@z3dev you were right, there were other cases worth checking. In the case where `startAngle` and `endAngle` make a full rotation, but `startAngle` is non-zero, I found a bug in `cylinderElliptic`. The other primitives with `startAngle` are fine, and I added some tests to be sure:

`extrudeRotate`: angle already adjusted by rotation check `totalRotation === Math.PI * 2`
`torus`: uses `extrudeRotate`
`circle`: uses ellipse
`ellipse`: closed by `geom2.fromPoints`
`star`: closed by `geom2.fromPoints`
`cylinderElliptic`: has a bug, now fixed

It should be good to go now. But let me know if you think of other possible issues.

z3dev (approved):
@platypii thanks for making the extra effort on these changes.

Issue #1086, opened 2022-05-08
TS definitions for geometries


z3dev:
## Expected Behavior

Proper TS definition of color as an attribute of geometries.

## Actual Behavior

it seems that there's a hierarchy of classes where Color is the base class. Totally wrong.

## Steps to Reproduce the Problem

  1. See code

## Specifications

  - Version: all
  - Platform: all
  - Environment: all

z3dev:
@platypii can you take a look at this? i'm not knowledgable about TypeScript definitions.

z3dev:
Just a quick look at the definitions... I don't understand why Colored is 'type' defined. I think the geometries should not define 'types' but rather 'interfaces' only. No hiearchies, no types.

Interfaces are more flexible and extendable. And users can define 'types' if they want to define objects with methods, etc.

platypii:
I'm experimenting with different type hierarchies to see what will be best. The current types are not _wrong_ but I do think there is room for improvement. I agree that `Colored` should probably be an interface not a type alias.

Here is a simplified version of the jscad geometries, for testing different types. Commented lines are type errors:

```typescript
type RGB = [number, number, number]
type RGBA = [number, number, number, number]

type Vec3 = [number, number, number]

type Colored = {
  color?: RGB | RGBA
}

interface Geom3 extends Colored {
  polygons: Array<Poly3>
}

interface Poly3 extends Colored {
  vertices: Array<Vec3>
}

type Geometry = Geom3 | Poly3

const g0 = { }
const g1 = { polygons: [] }
const g2 = { polygons: [], color: [1, 0, 0] }
const g3 = { vertices: [] }
const g4 = { vertices: [], color: [1, 0, 0] }
const g5: Colored = { }
const g6: Colored = { color: [1, 0, 0] }
const g7: Geom3 = { polygons: [] }
const g8: Geom3 = { polygons: [], color: [1, 0, 0] }
const g9: Poly3 = { vertices: [] }
const g10: Poly3 = { vertices: [], color: [1, 0, 0] }
const g11: Geometry = { polygons: [] }
const g12: Geometry = { polygons: [], color: [1, 0, 0] }
const g13: Geometry = { vertices: [] }
const g14: Geometry = { vertices: [], color: [1, 0, 0] }

function isColored(colored: Colored): boolean {
  return colored.color !== undefined
}

function isColoredGeometry(geometry: Geometry): boolean {
  return geometry.color !== undefined
}

isColored(g0)
[...]

PR #1098, opened 2022-05-30
refactor(modeling): refactor geometry types

platypii:
This is an attempt to improve the JSCAD geometry type definitions. See discussion from #1086

I made a few proposed changes:

- Changed geometries to not "extend" type `Colored`
- But all geometries have an _optional_ `color` field
- Changed `Colored` to _require_ that a color be present.
- Changed `Colored` from a type to an interface
- Defined a new `type Color = RGA | RGBA` to simplify the type definitions, and make it easier to change color types in the future

A concrete advantage is that a geometry is now only definitively colored after a call to `colorize` and this can be checked by typescript automatically. So for example this gives a type warning:

```ts
function example(geom: Geom3) {
  console.log(geom.color[0])
}
```

But this considered valid (correctly):

```ts
function example(geom: Geom3) {
  const colored = colorize([0,0,0], geom)
  console.log(colored.color[0])
}
```

I think that's pretty neat.

Feedback welcome though.

### All Submissions:

### DIFF
--- a/packages/modeling/src/colors/colorize.d.ts
+++ b/packages/modeling/src/colors/colorize.d.ts
@@ -5,9 +5,10 @@ import { RGB, RGBA } from './types'
 
 export default colorize
 
-declare function colorize<T extends Geometry>(color: RGB | RGBA, object: T): T
-declare function colorize<T>(color: RGB | RGBA, object: T): T & Colored
+// Single Geom3 returns Colored Geom3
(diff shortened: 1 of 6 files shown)


z3dev (approved):
@platypii these changes look really good. I was hoping that interfaces would be the main typing, as other types are now easy to add.


hopefully, @ftrading or @danmarshall will find these changes useful.

frading:
Thanks a lot @platypii and @z3dev . Yes, I think that'll definitely be useful.

Issue #1095, opened 2022-05-27
Nodejs Math.hypot performance is pathetic.


briansturgill:
Math.hypot is written to the IEEE standard for hypot... which forces the routine to handle very tiny/large (e-/+250) numbers by scaling the incoming numbers. It slows things down greatly, though not as much as in node... something very wrong there.
Anyway, a CAD package doesn't need to handle such tiny/huge numbers... the "simple" hypot would work just fine.
As I pointed out in a different issue, hypot is VERY critical path in boolean operations. And is used a lot in general.
Here's a benchmark:
-- Output --
```
node mathbench.js
hypot loop: 4059.450ms
simple hypot loop: 207.846ms
sqrt loop: 213.970ms
```
-- Code ---
```javascript
size=100000000;
console.time("hypot loop");
let sum = 0
let x = 1.3
let y = 2.1834
let z = 3.134
for(i=0; i<size; i++) {
	sum += Math.hypot(x, y, z)
	x++
}
console.timeEnd("hypot loop");
console.time("simple hypot loop");
x = 1.3
y = 2.1834
z = 3.134
for(i=0; i<size; i++) {
	sum += Math.sqrt(x * x+ y*y + z*z)
	x++
}
console.timeEnd("simple hypot loop");
console.time("sqrt loop");
x = 2
for(i=0; i<size; i++) {
	sum += Math.sqrt(x)
	x++
}
console.timeEnd("sqrt loop");
```

platypii:
The change to `Math.hypot` was my doing in PR #996. I was trying to minimize floating point error, but that is surprising how bad the performance is.

Maybe worth switching back to pythagorean distance. Or adding `maths.utils.trigonometry.hypot` using `Math.sqrt(...)` under the hood.

hrgdavor:
@briansturgill thanks for the insight. I was expecting that Math.hypot is faster since I expectig the performance was the reason begind adding it to Math. It is good to be aware of the performance hit it brings.

PR #1099, opened 2022-05-30
perf(modeling): remove use of slow Math.hypot

platypii:
It turns out that `Math.hypot` is much slower than `Math.sqrt`, as pointed out by @briansturgill in #1095

I liked using hypot for increased accuracy, and nicer to read. But since it is slower, I think it should go away.

I had to remove the tests for accuracy on very small, and very large numbers (1e200+).

### All Submissions:


hrgdavor (commented):
looks good

z3dev:
@platypii sorry. I'm not going for this without a good reason. For places where performance matters then the larger function can opt for using sqrt(). I think we should open an issue with gl matrix, and ask why that library uses hypot()

platypii:
@z3dev that's fair. I'll see if I can measure the speed difference in real models.

platypii:
Changing from `hypot` to `sqrt` makes a significant 30%+ speed up on unions. I think it's worth considering the change.

|  | hypot | no hypot | improvement |
| --- | --- | --- | --- |
| disjoint (x=4) | 480ms | 480ms | 0% |
| touching (x=2) | 8500ms | 5800ms | 31.8% |
| overlap (x=1) | 8900ms | 5850ms | 34.3% |
| inside (x=0) | 7300ms | 4800ms | 34.2% |

<details>
<summary>Test script</summary>

```js
const { union } = jscad.booleans

const main = () => {
  const segments = 100
  const x = 1
  let s1 = sphere({ segments })
  let s2 = sphere({ segments, center: [x, 0, 0] })
  const startTime = performance.now()
  const out = union(s1, s2)
  console.log(`Union time ${performance.now() - startTime} ms`)
  return out
}

module.exports = { main }
```
</details>

hrgdavor:
that is massive, definitely worth changing

z3dev:
> Changing from `hypot` to `sqrt` makes a significant 30%+ speed up on unions. I think it's worth considering the change.
> 
> hypot	sqrt	improvement
> disjoint (x=4)	480ms	480ms	0%
> touching (x=2)	8500ms	5800ms	31.8%
> overlap (x=1)	8900ms	5850ms	34.3%
> inside (x=0)	7300ms	4800ms	34.2%
> Test script

I'm not saying that the replacement shouldn't be done. I'm saying that we should be selective about the replacement. Booleans may benefit but we don't want to create other issues.

We should do some research, and ask the gl matrix geeks. And try to understand the math completely.

platypii:
https://github.com/toji/gl-matrix/issues/446

z3dev:
FYI, here's the js-core implementation of hypot()

https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.math.hypot.js

platypii:
The core-js polyfill implementation is somehow even slower than Math.hypot despite using sqrt. :vomiting_face: 

| environment | hypot | sqrt | core-js |
|-------------|-------|------|-------|
| chrome      | 235ms | 12ms | 328ms |
| firefox     | 64ms  | 12ms | 160ms |
| safari      | 390ms | 74ms | 1830ms |
| node        | 248ms | 11ms | 370ms |

packages/modeling/src/maths/mat4/rotate.js
```diff
@@ -14,14 +15,14 @@ const copy = require('./copy')
  */
 const rotate = (out, matrix, radians, axis) => {
   let [x, y, z] = axis
-  let len = Math.hypot(x, y, z)
+  const lengthSquared = x * x + y * y + z * z
 
-  if (Math.abs(len) < 0.000001) {
+  if (Math.abs(lengthSquared) < EPSILON) {
```
z3dev, line 20: I suggest that we use the EPS for normals, then we know that this function works for calculations with normals.

packages/modeling/src/maths/mat4/fromRotation.js
```diff
@@ -21,14 +21,14 @@ const { EPSILON } = require('./constants')
  */
 const fromRotation = (out, rad, axis) => {
   let [x, y, z] = axis
-  let len = Math.hypot(x, y, z)
+  const lengthSquared = x * x + y * y + z * z
 
-  if (Math.abs(len) < EPSILON) {
+  if (Math.abs(lengthSquared) < EPSILON) {
```
z3dev, line 26: Let's use the EPS for normals.

z3dev (changes requested):

z3dev (approved):
@platypii thanks again. Let's see if the general user community feels any performance improvement.

Issue #85, opened 2015-04-08
Feature Request: Support nested interactive parameters


thehans:
Currently the getParameterDefinitions() function accepts 4 types: float, int, text, and choice

I am proposing a fifth type that would allow for nesting / grouping of parameters.

This would be useful for scripts which have multiple independent pieces with their own parameters.

Here is an example of how I would picture it being used:
```javascript
{
  name: 'chassisBolt',     // a javacsript object will be created, filled with its own parameters
  type: 'parameterGroup',  // not sure best name for param type, other ideas include 'object', 'group', 'grouping'
  caption: 'Chassis Bolt', // optional, displayed left of the input field
					       // if omitted, the 'name' is displayed (i.e. 'chassisBolt')
  parameters: { // params go here in the same manner as at the 'top level', including possibly more parameter groups
	  {  name: 'headDiameter',  type: 'float', initial: 6, caption: 'Head Diameter:' }
	  {  name: 'shankDiameter',  type: 'float', initial: 3, caption: 'Shank Diameter:' }
	  {  name: 'length',  type: 'float', initial: 20, caption: 'Length:' }
  }
}
```

When a parameter group is defined the interactive parameter window should generate a bold heading with the given caption or name, and list all contained parameters beneath it, indented some amount.  

Then if I have a function that only needs the bolt related parameters I can pass in params.chassisBolt object and leave out all the other unrelated parameters.

z3dev:
I agree. A group of the parameters is more important then a title, and allows more flexibility.
I have this coded already but have to wait for some additional merges to occur first.

z3dev:
@thehans please look at some of the designs at www.z3d.jp This website has support for "groups" already.

Basically, the "group" parameter becomes a new row in the table of parameters, and contains a single header (HTML "th" element). And of course, a new style is added to the CSS file.

Good?

thehans:
I don't see any openjscad on that site, not sure what I'm supposed to be looking at.

z3dev:
Sorry. I didn't provide a complete URL. Please review some of the designs at http://www.z3d.jp/lab/

z3dev:
@thehans grouping of parameters has been implemented, and groups can be collapsed or expanded interactively.

please try the 'Birthday Ballons' example as www.openjscad.xyz

Issue #49, opened 2014-09-24
Params callback functions


ErikDeBruijn:
Perhaps it's a good idea to allow callback functions or at least regexes to validate content passed through the parameters div input controls or for the ones passed by URL. This will prevent parse errors in many cases. Right now javascript errors are thrown, but it's better to allow people to define friendly messages, perhaps even things like "The wall thickness cannot be negative".

```
{
  name: 'shape',
  type: 'choice',
  values: ["TRI", "SQU", "CIR"],               // these are the values that will be supplied to your script
  captions: ["Triangle", "Square", "Circle"],  // optional, these values are shown in the listbox
                                               // if omitted, the items in the 'values' array are used
  caption: 'Shape:',                           // optional, displayed left of the input field
  initial: "SQU",                              // optional, default selected value
                                               // if omitted, the first item is selected by default
                                               // NOTE: parameter "default" is deprecated
  validate: shapeValidate
}
```

An example to clarify:

``` javascript
function validate(parameterDefinitions,paramName,paramValue){
if(paramValue == "circular") && parameterDefinitions['height'].value > 10)
  status("The "+paramValue+" shape can only be used when the height is more than 10.");
}
```

paulftw:
This issue was opened in 2014, had some commits but no recent activity for past 2 years.
And it's not a bug.
I suggest we close it as stale.

z3dev:
Let’s let @ErikDeBruijn decide to close. There’s no solution yet.

ErikDeBruijn:
I don't have a strong need for this anymore but if my request is useful to others we can keep it alive.
Im fine with focus on other, more recent feature requests, though! Perhaps that is time better spent.

PR #1100, opened 2022-05-30
feat(modeling): allow zero size in primitives

platypii:
This PR allows primitives to "scale to zero" in size. Specifically, for rectangles and cuboids allow size equal to zero. And for ellipses and cylinders allow radius equal to zero.

Previously these primitives would throw an exception if the size was zero. This leads to a bad user experience in cases where you have a parametric design that has a parameter that might scale to zero and disappear. Right now, you need awkward special cases in your code to handle that (a user raised this issue on discord recently).

Another example where this is helpful: Using a slider parameter, today if you slide it to zero you get an exception and the design does not update on the screen. After this PR, you can scale these to zero, and it works fine:

```js
const { circle, cube, cylinder, square } = require('@jscad/modeling').primitives

const getParameterDefinitions = () => ([
  { name: "size", type: "slider", initial: 1, min: 0, max: 10 }
])

const main = ({ size }) => {
  return [
    circle({ radius: size, center: [0, 0, 0] }),
    cylinder({ radius: size, center: [5, 0, 0] }),
    cube({ size, center: [10, 0, 0] }),
    square({ size, center: [15, 0, 0] })
  ]
}

module.exports = { main, getParameterDefinitions }
```

### All Submissions:

### DIFF
--- a/packages/modeling/src/primitives/circle.js
+++ b/packages/modeling/src/primitives/circle.js
@@ -2,7 +2,7 @@ const { TAU } = require('../maths/constants')
 
 const ellipse = require('./ellipse')
 
-const { isGT } = require('./commonChecks')
+const { isGTE } = require('./commonChecks')
 
 /**
  * Construct a circle in two dimensional space where all points are at the same distance from the center.
@@ -28,7 +28,7 @@ const circle = (options) => {
   }
   let { center, radius, startAngle, endAngle, segments } = Object.assign({}, defaults, options)
 
-  if (!isGT(radius, 0)) throw new Error('radius must be greater than zero')
+  if (!isGTE(radius, 0)) throw new Error('radius must be positive')
 
   radius = [radius, radius]
 
--- a/packages/modeling/src/primitives/circle.test.js
+++ b/packages/modeling/src/primitives/circle.test.js
@@ -146,3 +146,10 @@ test('circle (options)', (t) => {
   t.deepEqual(pts.length, 5)
   t.true(comparePoints(pts, exp))
 })
+
+test('circle (radius zero)', (t) => {
+  const geometry = circle({ radius: 0 })
+  const pts = geom2.toPoints(geometry)
+  t.notThrows(() => geom2.validate(geometry))
+  t.is(pts.length, 0)
+})
--- a/packages/modeling/src/primitives/cube.js
+++ b/packages/modeling/src/primitives/cube.js
@@ -1,6 +1,6 @@
 const cuboid = require('./cuboid')
 
-const { isGT } = require('./commonChecks')
+const { isGTE } = require('./commonChecks')
 
 /**
  * Construct an axis-aligned solid cube in three dimensional space with six square faces.
@@ -20,7 +20,7 @@ const cube = (options) => {
   }
   let { center, size } = Object.assign({}, defaults, options)
 
-  if (!isGT(size, 0)) throw new Error('size must be greater than zero')
+  if (!isGTE(size, 0)) throw new Error('size must be positive')
 
   size = [size, size, size]
 
--- a/packages/modeling/src/primitives/cube.test.js
+++ b/packages/modeling/src/primitives/cube.test.js
@@ -46,3 +46,10 @@ test('cube (options)', (t) => {
   t.is(pts.length, 6)
   t.true(comparePolygonsAsPoints(pts, exp))
 })
+
+test('cube (zero size)', (t) => {
+  const obs = cube({ size: 0 })
+  const pts = geom3.toPoints(obs)
+  t.notThrows(() => geom3.validate(obs))
+  t.is(pts.length, 0)
+})
--- a/packages/modeling/src/primitives/cuboid.js
+++ b/packages/modeling/src/primitives/cuboid.js
@@ -23,7 +23,10 @@ const cuboid = (options) => {
 
   if (!isNumberArray(center, 3)) throw new Error('center must be an array of X, Y and Z values')
   if (!isNumberArray(size, 3)) throw new Error('size must be an array of width, depth and height values')
-  if (!size.every((n) => n > 0)) throw new Error('size values must be greater than zero')
+  if (!size.every((n) => n >= 0)) throw new Error('size values must be positive')
+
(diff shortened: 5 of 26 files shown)


hrgdavor (dismissed):
I like these changes. It does raise some additional questions.
 - should we define official api to tell if some geom is empty
 - should transform or colorize or other functions have some special handling of those zero size primitives
 - do we allow scale=0 ?

t1u1:
> This leads to a bad user experience in cases where you have a parametric design that has a parameter that might scale to zero and disappear. Right now, you need awkward special cases in your code to handle that (a user raised this issue on discord recently).

:100: agree with this

In my own app, I had to litter my code with many if conditions because some circle's radius would become <= 0 depending on a user-specified parameter.

Having these checks in jscad would simplify application code, while not significantly impacting performance.

platypii:
> Also, sliders only work on integers, so one(1) is the smallest viable size.

There is no reason you can't set the slider minimum to 0, I gave an example in the original post.

> What's the use case?

One of the reasons for parametric designs is to be able to experiment, and try out different options. Or maybe be able to generate designs with optional features.

Currently it is awkward to make a design where an element can "scale to zero and disappear". You would either need special handling in the code (`if (size > 0) objects.push(...)`). Or you would need a seperate parameter to enable or disable a design feature (eg- a check box). These are awkward workarounds, and users on discord have complained about them. And I agree.

There are choices of how JSCAD could handle zero sizes: throw an exception, or quietly produce an empty geometry. I feel that we should throw errors for things like negative radius, because it is unclear what that even means. But for radius zero, I think it is MUCH more user friendly to just produce an empty geometry.

> do we allow scale=0 ?

Yes, today we allow scale 0

> should transform or colorize or other functions have some special handling of those zero size primitives

I don't see why they need special handling? They work fine today on empty geometry. It was always possible to have empty geoemetries (eg- intersection of nonoverlapping cubes)

> What's the use case?

Here's a parametric crown with re-sizable point balls. This code would be more elegant with this PR change:

```js
const jscad = require("@jscad/modeling")
const { subtract } = jscad.booleans
const { cylinder, sphere } = jscad.primitives
const { rotateX, rotateZ, translate } = jscad.transforms

const getParameterDefinitions = () => ([
  {name: "size", type: "slider", min: 0, initial: 1, max: 10}
])

const main = ({ size }) => {
  const segments = 100
  const cutout = translate([0, 0, 6], rotateX(Math.PI / 2, cylinder({ height: 20, radius: 6, segments })))
  const ball = sphere({ center: [10, 0, 2.5], radius: size / 10 })
  return [
    subtract(
      cylinder({ height: 10, radius: 10, segments }),
      cylinder({ height: 10, radius: 9.8, segments }),
      cutout,
      rotateZ(Math.PI * 2 / 3, cutout),
      rotateZ(Math.PI * 4 / 3, cutout),
    ),
    ball,
    rotateZ(Math.PI * 1 / 3, ball),
    rotateZ(Math.PI * 2 / 3, ball),
    rotateZ(Math.PI * 3 / 3, ball),
    rotateZ(Math.PI * 4 / 3, ball),
    rotateZ(Math.PI * 5 / 3, ball),
  ]
}

module.exports = { main, getParameterDefinitions }
```

crown

z3dev:
I think we are actually taking about the same thing. Designs need to validate parameters. Of course, the 'defaults' are useful to control the UI but users can still input almost anything.

There have been a few requests in the past to add validations. One was a request for general validation, adding validation per parameter, or a general validation function.

https://github.com/jscad/OpenJSCAD.org/issues/49

Another was about validating the relationships between parameters, i.e. parameter A should be larger than parameter B, etc.

https://github.com/jscad/OpenJSCAD.org/issues/85

Also, there was a nice PR for creating dependencies between parameters as well, but against V1. Something like this would add some nice flexibility to the parameter definitions.

https://github.com/jscad/OpenJSCAD.org/pull/42

z3dev:
Just continuing, it would be best if the validation of parameters was performed  within the UI, rather then passing bad parameters into the main design logic. So, maybe another predefined function needs to be called, which can perform the validations and throw an appropriate error.

hrgdavor:
Or just thow a ValidationError from main with info on which fields failed and message for each

z3dev:
@platypii are you still looking to make these changes? V2 or V3?

platypii:
> @platypii are you still looking to make these changes? V2 or V3?

I still strongly believe this is a good change. It makes the designs more robust, for cases where size zero makes logical sense, it just makes it easier for users to "play" with designs more freely. This makes parameters much more useful.

I just rebased it onto V3 branch. How do feel about it as a V3 change @z3dev?

z3dev:
@platypii thanks.

I reviewed the changes again, and came to the same conclusion that these are really improving the usability of the API.

Making these changes to V2 is fine. I'll review seriously. ;)

platypii:
@z3dev I re-rebased it onto V2 master branch again. Please review :-)

packages/modeling/src/primitives/circle.js
```diff
@@ -28,7 +28,7 @@ const circle = (options) => {
   }
   let { center, radius, startAngle, endAngle, segments } = Object.assign({}, defaults, options)
 
-  if (!isGT(radius, 0)) throw new Error('radius must be greater than zero')
+  if (!isGTE(radius, 0)) throw new Error('radius must be greater than zero')
```
t1u1, line 31: The message could also be changed to "greater than or equal to".

Similarly in other places too.

z3dev: do agree

z3dev (changes requested):
@platypii thanks. this look good. just the one nit about the error messages for GTE

z3dev (commented):
@platypii thanks for the late changes. these look fine.

however... i was wondering about the 'rounded' primitives. wouldn't these follow the same?

also, not sure but maybe star as well.

platypii:
@z3dev good call. I added similar logic and tests for zero size for rounded primitives. Star has a confusing relationship of inner and outer radius, so I left it alone.

z3dev (approved):
@platypii well, well... that was easy. ha ha ha

all good. thanks again.

PR #1102, opened 2022-06-02
test(x3d-deserializer): expected values for createTransform test

platypii:
Remove `x3d-deserializer/tests/helpers` which contained unnecessary re-implementations of vector3, vector4, matrix4, etc.

Instead, just use hard-coded expected values for `createTransform.test.js`.

This removes 584 of lines of redundant code, which never matched jscad code style anyway. And the test is just as good as it was before.

### All Submissions:


z3dev (approved):
@platypii thanks for cleaning up this code, which was from ... ?? (I couldn't find the origin JS library, but I know there's one!)

Issue #1094, opened 2022-05-27
Problem with JSCAD measureBoundingBox/measureBoundingSphere caching.


briansturgill:
1. Because Geom2/Geom3/Path2 geometries are NOT immutable, any transformation on a Geometry after the first call to measureBoundingX will in most cases return an invalid result from the cache.

2. The caching of values makes the performance testing numbers very skewed. In real life the cache would be very seldom used as new geometries are usually created for each position in which they are used. One questions if the cache should be there at all.

3. The major measuring of bounding spheres is done by poly3.measureBoundingSphere (which calls poly3.measureBoundingBox), which is not included at all in the performance testing. It is also not cached... if caches actually are useful, this should likely be cached as poly3.measureBoundingSphere is VERY critical path in boolean operations.

You might want to look at the attached screen capture from my running profiling on the CSharp version I'm working on. The profiling you see was run on the union portion of the performance testing translated from JSCAD.
The need for absolute speed in the measurement of bounding boxes and spheres is amazing.
I used profiling via nodejs on JSCAD and got similar results, but the profiling is harder to read/understand there.

Brian
![Capture](https://user-images.githubusercontent.com/7350143/170785044-54f695a4-5188-44b3-86d3-ea8ba205d0d6.PNG)

hrgdavor:
I would not disregard caching of bounding boxes as easily. The bounding box can be safely recalculated for transforms that only translate and scale (no rotation). One of reasons I wanted to introduce caching is to allow for shape generators to prefill the cache. For example it is trivial to precalculate bounding box of a sphere, but is exponentialy more expensive wirh increasing segment count.

One more thing, if cache is to removed, It should not be because of geometries that are mutable. The direction of making geometries immutable is already a chosen path and I would not like to revert it.

briansturgill:
...
> One more thing, if cache is to removed, It should not be because of geometries that are mutable. The direction of making geometries immutable is already a chosen path and I would not like to revert it.

I think you misunderstand... all geometries ARE mutable. That is how Transform/ApplyTransforms works.
The "measurement.measureBounding... routines INCORRECTLY do the caching. They will return incorrect values
if the routine is called before a Transform, and then called again. The current caching simply is incorrect.
If you want caching, the the cached bounds should be on the geometry object and the transform/applyTransform
methods must invalidate the cached values. The current situation is simply broken.

Further, caching isn't really doing anything at all w.r.t. to usage by JSCAD internals.
[...]

PR #1106, opened 2022-06-10
perf(modeling): cache poly3.measureBoundingSphere for BSP performance

platypii:
A number of performance issues were pointed out by @briansturgill. One of them was caching of `measureBoundingSphere` in the BSP boolean operations (#1094). BSP operations use measureBoundingSphere to determine if objects overlap.

By adding caching to `poly3.measureBoundingSphere`, we can get a significant speed up in BSP operations. This uses the same caching code as in the `measurements` package, but applied to the `poly3` package.

For unioning two overlapping spheres with 100 segments, we get **49,366** cache misses (measurements), and **97,030,301** cache hits saved! That's a 99.95% cache hit rate. :grimacing: 

This results in major performance improvements as measured in BSP union operations. Comparing the last release 2.5.10, master (with no-hypot #1099), and measureBoundingSphere caching:

|  | 2.5.10 | master | caching |
| --- | --- | --- | --- |
| disjoint (x=4) | 480ms | 480ms | 480ms |
| touching (x=2) | 8500ms | 5800ms | 3800ms |
| overlap (x=1) | 8900ms | 5850ms | 4000ms |
| same (x=0) | 7300ms | 4800ms | 3700ms |
| inside (x=0) | 4850ms | 3300ms | 2500ms |

So roughly speaking, removing `hypot` gave a ~30% improvement, and caching gives ANOTHER ~30% performance improvement. This gives a total speedup in BSP operations of 48-55% this month! :sunglasses: 

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/poly3/measureBoundingSphere.d.ts
+++ b/packages/modeling/src/geometries/poly3/measureBoundingSphere.d.ts
@@ -1,6 +1,6 @@
 import Poly3 from './type'
-import Vec3 from '../../maths/vec3/type'
+import Vec4 from '../../maths/vec4/type'
 
 export default measureBoundingSphere
 
-declare function measureBoundingSphere(polygon: Poly3): [Vec3, Vec3]
+declare function measureBoundingSphere(polygon: Poly3): Vec4
--- a/packages/modeling/src/geometries/poly3/measureBoundingSphere.js
+++ b/packages/modeling/src/geometries/poly3/measureBoundingSphere.js
@@ -1,6 +1,8 @@
 const vec3 = require('../../maths/vec3')
 const measureBoundingBox = require('./measureBoundingBox')
 
+const cache = new WeakMap()
+
 /**
  * Measure the bounding sphere of the given polygon.
  * @param {poly3} polygon - the polygon to measure
@@ -8,12 +10,19 @@ const measureBoundingBox = require('./measureBoundingBox')
  * @alias module:modeling/geometries/poly3.measureBoundingSphere
  */
 const measureBoundingSphere = (polygon) => {
+  let boundingBox = cache.get(polygon)
+  if (boundingBox) return boundingBox
+
   const box = measureBoundingBox(polygon)
-  const center = box[0]
+  const center = vec3.create()
   vec3.add(center, box[0], box[1])
   vec3.scale(center, center, 0.5)
   const radius = vec3.distance(center, box[1])
-  return [center, radius]
+
+  boundingSphere = [center, radius]
+  cache.set(polygon, boundingSphere)
+
+  return boundingSphere
 }
 
 module.exports = measureBoundingSphere
--- a/packages/modeling/src/geometries/poly3/measureBoundingSphere.js
+++ b/packages/modeling/src/geometries/poly3/measureBoundingSphere.js
@@ -1,5 +1,4 @@
 const vec3 = require('../../maths/vec3')
-const measureBoundingBox = require('./measureBoundingBox')
 
 const cache = new WeakMap()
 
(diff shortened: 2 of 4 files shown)


hrgdavor:
woot !!! 😃

z3dev:
@platypii nice

I have some changes on branch perf-booleans which speed up the calculation of bounding sphere. You could merge those into these changes and improve even further.

platypii:
@z3dev that's interesting, I'm looking at that branch now, and it actually removes a call to `poly3.measureBoundingSphere` so I will have to test to see how these two branches interact with each other on performance.

Another thought I had: why are we using bounding spheres and not bounding boxes? At least in the case of poly3, the bounding sphere is calculated as the sphere containing the bounding box. So the sphere will always be larger than the box. If that's the case, it would be more precise if we checked the smaller bounding box instead of the sphere for checking whether two poly3's have overlap. I'll do some tests.

briansturgill:
> This results in major performance improvements as measured in BSP union operations. Comparing the last release 2.5.10, master (with no-hypot #1099), and measureBoundingSphere caching:
> 	2.5.10 	master 	caching
> disjoint (x=4) 	480ms 	480ms 	480ms
> touching (x=2) 	8500ms 	5800ms 	3800ms
> overlap (x=1) 	8900ms 	5850ms 	4000ms
> same (x=0) 	7300ms 	4800ms 	3700ms
> inside (x=0) 	4850ms 	3300ms 	2500ms

I'd like to get access to that test program if I could (to translate and run against CSharpCAD).
Brian

briansturgill:
Congratulations... it's not every day a mature package can get that kind of improvement!

platypii:
@briansturgill thanks! Here is the test script I used. It's a simple union of two spheres at different x distances. I also include a smaller sphere inside a larger sphere test since it exhibits slightly different performance. I'm measuring specifically the `union` call.

<details>
<summary>spheres.js</summary>

```js
const { union } = jscad.booleans

const segments = 100

const main = () => {
  test(4, 1, "disjoint")
  test(2, 1, "touching")
  test(1, 1, "overlap")
  test(0, 1, "same")
  test(0, 0.5, "inside")
  return sphere()
}

const test = (x, radius, name) => {
  const s1 = sphere({ segments })
  const s2 = sphere({ segments, center: [x, 0, 0], radius })
  const startTime = performance.now()
  const out = union(s1, s2)
  console.log(`${name} ${performance.now() - startTime} ms`)
  return out
}

module.exports = { main }
```

</details>

platypii:
@z3dev nice one on `perf-booleans` branch! The fastest BSP I've seen is a cached version of your `perf-booleans` branch. It required some tweaking, and loses some of your memory savings, but this gives the best performance I've seen yet!

|  | 2.5.10 | master | poly3-cache | perf-booleans | perf-booleans-cache |
| --- | --- | --- | --- | --- | --- |
| disjoint (x=4) | 480ms | 480ms | 480ms | 480ms | 480ms |
| touching (x=2) | 8500ms | 5800ms | 3800ms | 4530ms | 3570ms |
| overlap (x=1) | 8900ms | 5850ms | 4000ms | 4820ms | 3840ms |
| same (x=0) | 7300ms | 4800ms | 3700ms | 3980ms | 3160ms |
| inside (x=0) | 4850ms | 3300ms | 2500ms | 2580ms | 2170ms |

This would push the total improvement this month to 55-58% faster boolean operations :exploding_head:

z3dev:
> Another thought I had: why are we using bounding spheres and not bounding boxes? At least in the case of poly3, the bounding sphere is calculated as the sphere containing the bounding box. So the sphere will always be larger than the box. If that's the case, it would be more precise if we checked the smaller bounding box instead of the sphere for checking whether two poly3's have overlap. I'll do some tests.

Simply put... the polygon is being tested to see if it can be cut by the plane. It uses the distance away from 0,0,0 to see if there's an overlap. The sphere is being calculated to get the **largest** possible delta (radius).

z3dev:
> @z3dev nice one on `perf-booleans` branch! The fastest BSP I've seen is a cached version of your `perf-booleans` branch. It required some tweaking, and loses some of your memory savings, but this gives the best performance I've seen yet!
> 
> 2.5.10	master	poly3-cache	perf-booleans	perf-booleans-cache
> disjoint (x=4)	480ms	480ms	480ms	480ms	480ms
> touching (x=2)	8500ms	5800ms	3800ms	4530ms	3570ms
> overlap (x=1)	8900ms	5850ms	4000ms	4820ms	3840ms
> same (x=0)	7300ms	4800ms	3700ms	3980ms	3160ms
> inside (x=0)	4850ms	3300ms	2500ms	2580ms	2170ms
> This would push the total improvement this month to 55-58% faster boolean operations 🤯

Feel free to merge together  the new tweaks. 😃

briansturgill:
@platypii Thanks for the script, timing was wonderful as I was just getting around to doing the optimizing.
Note, I extended the script to do Subtract and Intersect too.

Results of optimization of by simplifying hypot, inlining critical checking and caching Geom2/3 BoundingBox and Poly3 BoundingSphere (CSharpCAD doesn't have measureBoundingBox, but rather puts it on the Geom2/Geom3 objects.)

image

platypii:
@z3dev I copied your memory-optimized measureBoundingSphere from #1107 back here into `poly3.measureBoundingSphere`. 

This is the fastest BSP, but does include a breaking change to the API. The performance benefit of changing the api to return `vec4` instead of `[vec3, number]` is:

|  | [vec3, number] | vec4 |
| --- | --- | --- |
| disjoint (x=4) | 474ms | 470ms |
| touching (x=2) | 3950ms | 3650ms |
| overlap (x=1) | 4000ms | 3800ms |
| same (x=0) | 3460ms | 3330ms |
| inside (x=0) | 2290ms | 2140ms |

Which means we get 4-8% improvement in performance from changing the measureBoundingSphere API. Do I think anyone is using `poly3.measureBoundingSphere`? No, almost certainly not. So I think we should change it, personally. But I am open to suggestions here @z3dev.

z3dev (approved):
@platypii lots of fun watching the two approaches being merged together. The performance improvement is really nice. Thanks again.

PR #1107, opened 2022-06-11
perf(modeling): rewrote measureBoundingSphere for speed and to reuse memory

z3dev:
WORK IN PROGRESS 
DO NOT MERGE

### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?

### DIFF
--- a/packages/modeling/src/geometries/poly3/measureBoundingSphere.d.ts
+++ b/packages/modeling/src/geometries/poly3/measureBoundingSphere.d.ts
@@ -3,4 +3,4 @@ import Vec3 from '../../maths/vec3/type'
 
 export default measureBoundingSphere
 
-declare function measureBoundingSphere(polygon: Poly3): [Vec3, Vec3]
+declare function measureBoundingSphere(polygon: Poly3): [Vec3, number]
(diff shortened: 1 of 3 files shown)


platypii:
@z3dev I added caching to this PR.

I had to refactor the measureBoundingSphere function a bit. Having a mutable `out` parameter breaks caching. I tried some variations, but this immutable function seemed like the cleanest and fastest solution.

I think we should merge this PR and close #1106

z3dev:
> @z3dev I added caching to this PR.
> I think we should merge this PR and close #1106

Ok. But I think measureBoundingSphere should be moved into poly3. The return value is slightly different but I doubt many users if any are using the functionality. Thoughts?

platypii:
> Ok. But I think measureBoundingSphere should be moved into poly3. The return value is slightly different but I doubt many users if any are using the functionality. Thoughts?

Yea after thinking about it, I agree, it would be really nice to not have duplicate measureBoundingSphere code just for booleans. Take a look at my PR #1106 which I just updated with your changes merged back into `poly3.measureBoundingSphere`.

PR #1110, opened 2022-06-12
perf(web): changed to retain webworker process

z3dev:
While working through some issues with rollup, I came across issues with the webworkify package. (It can only be packaged using Browserify. 😢  ) However, I tried to retain the webworker instead of terminating / initializing. It works just fine across the browsers. In addition. error handling has been improved so processing should be more robust as well.

Performance improves differently across the browsers, but basically the 100 ms delay of creating a new webworker is gone.

Manually tested these changes across Safari, Firefox, Chrome, and Edge on MacOS 10.14.6
Also Safari, Firefox, Chrome, and Edge on IOS

### All Submissions:

### DIFF
--- a/packages/web/src/sideEffects/worker/index.js
+++ b/packages/web/src/sideEffects/worker/index.js
@@ -5,25 +5,24 @@ const callBackToStream = require('../../most-utils/callbackToObservable')
 const makeWorkerEffect = (workerPath) => {
   const workerEventsCb = callBackToStream()
 
-  let _worker = WebWorkify(workerPath)
+  let _worker = WebWorkify(workerPath, {})
   _worker.onerror = (error) => workerEventsCb.callback({ error })
   _worker.onmessage = (message) => workerEventsCb.callback(message)
 
   const workerSink = (outToWorker$) => {
     // cancel whatever is going on in the worker by terminating it
     outToWorker$.filter(({ cmd }) => cmd === 'cancel')
-      .forEach((_) => _worker.terminate())
+      .forEach((task) => {
+        _worker.terminate()
+        _worker = WebWorkify(workerPath)
+        _worker.onerror = (error) => workerEventsCb.callback({ error })
+        _worker.onmessage = (message) => workerEventsCb.callback(message)
+      })
 
     // send other messages to the worker
     outToWorker$
       .filter(({ cmd }) => cmd !== 'cancel')
       .forEach((task) => {
-        // FIXME: sub optimal ! worker recreation is SLOW and should not be systematic
-        _worker.terminate()
-        _worker = WebWorkify(workerPath)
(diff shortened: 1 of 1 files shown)


z3dev:
Please try these changes across as many browsers as possible.

platypii (approved):
Looks good. I tested on various browsers and saw typical improvements of 60-80ms per model load. Awesome!

Tested on:
 - linux chrome
 - linux firefox
 - mac m1 chrome
 - mac m1 firefox
 - mac m1 safari

Since we are persisting the web worker, I also wanted to make sure that there wasn't a memory leak. I did this by updating the model repeatedly while monitoring memory usage. Even with large models, repeatedly loaded, the memory always freed back to baseline. :+1:

hrgdavor:
I know it is already merged, but not sure if I should make a separate issue

I was curious about the terminate of the worker and details how it would behave, and I added for my test purpose console log when terminating or sending a command.

I used a boolean on 2 spheres with enough segments to make it last more than 1 second, then I clicked parameter few times. 3 commands were sent to worker, and were all executed, there was no cancel happening.


image

```js
const { subtract } = jscad.booleans
const { translate } = jscad.transforms

function main({ // @jscad-params
  segments = 32 // {step:8}
}){
  return subtract(
    sphere({}),
    translate([0,0.5,0.5],sphere({segments}))
  )
}

module.exports = {main}
```

z3dev:
> I know it is already merged, but not sure if I should make a separate issue
> 
> I was curious about the terminate of the worker and details how it would behave, and I added for my test purpose console log when terminating or sending a command.
> 
> I used a boolean on 2 spheres with enough segments to make it last more than 1 second, then I clicked parameter few times. 3 commands were sent to worker, and were all executed, there was no cancel happening.
> 
> image
> 
> ```js
> const jscad = require('@jscad/modeling')
> const { sphere } = jscad.primitives
> const { subtract } = jscad.booleans
> const { translate } = jscad.transforms
> 
> function main({ // @jscad-params
>   segments = 32 // {step:8}
> }){
>   return subtract(
>     sphere({}),
>     translate([0,0.5,0.5],sphere({segments}))
>   )
> }
> 
> module.exports = {main}
> ```

What was the result of the viewer?

hrgdavor:
all get executed
image

I uses sequence of segments and manual update: 208,8,208,8 and let it play out
 - 208 was being executed
 - 8 was queued
 - 208 queued
 - 8 queued
 - 208 finished
 - 8 finished
 - 208 finished
 - 8 finished

I intentionally took large difference to see visual changes. Clearly no cancel is happening if new command is supplied
I tried 308 segments and it executed for 50 seconds, and no cancel happened. So I am not sure when cancel is triggered. This could be an issue with code before this PR, it probably is not caused by your recent changes.

PR #1115, opened 2022-06-22
cleanup old examples

platypii:
Remove old, broken examples in `packages/examples/old`. Some of these are V1, many are broken in both V1 and V2. I skimmed through them looking for examples that were worth preserving or translating to V2. I didn't see any which were particularly interesting, but open to suggestions.

There was also a `site` directory which contained a literal "lorem ipsum" placeholder page. I removed that too.

Red code is the best code :fire: 

### All Submissions:


z3dev:
@SimonClark might be interested in some of these. let's see if he is still following. ;)

z3dev (approved):
@platypii thanks for the cleanup. it saves a few bytes in the downloads.

FYI, those examples are available on the V1 and V2 branches, which will be preserved.

PR #1128, opened 2022-07-24
Refactor to use TAU constant

z3dev:
These changes introduce TAU to the modeling library, which includes...
- adding TAU constant
- reworking primitives to use TAU
- reworking tests to use TAU exclusively
- reworking internal logic to use TAU
- changed examples in documentation to use TAU

Testing was conducted with CLI and WEB UI as well. All browsers work just fine.

### All Submissions:


z3dev:
@platypii please review. Also, you should change your user I'd to @playtau 😃

packages/modeling/src/primitives/cylinderElliptic.js
```diff
@@ -17,7 +17,7 @@ const { isGT, isGTE, isNumberArray } = require('./commonChecks')
  * @param {Array} [options.startRadius=[1,1]] - radius of rounded start, must be two dimensional array
  * @param {Number} [options.startAngle=0] - start angle of cylinder, in radians
  * @param {Array} [options.endRadius=[1,1]] - radius of rounded end, must be two dimensional array
- * @param {Number} [options.endAngle=(Math.PI * 2)] - end angle of cylinder, in radians
+ * @param {Number} [options.endAngle=Math.TAI] - end angle of cylinder, in radians
```
platypii, line 20: Typo: TAI

platypii:
I have mixed feelings on this. I think it is actually very aesthetic in a lot of ways. I personally find the code more readable and understandable in most circumstances. But will this confuse users? I also don't love that it modifies the global `Math` object. Overall I am supportive, but need to think about it a bit.

hrgdavor:
I would vote to not modify global Math object. Especially if future will see us moving to es6 modules, such constant is more suited to be imported.

z3dev:
FYI, there is now adoption of tau by the Rust language, Microsoft’s .NET framework, Boost C++ library, and Python.

it's only a matter of time that JavaScript gains Math.TAU

z3dev:
> I have mixed feelings on this. I think it is actually very aesthetic in a lot of ways. I personally find the code more readable and understandable in most circumstances. But will this confuse users? I also don't love that it modifies the global `Math` object. Overall I am supportive, but need to think about it a bit.

I adjusted the code to add Math.TAU if not present. Hopefully, TAU will be added in a few years. :)

z3dev:
@platypii (alias @platytau) any thoughts?

platypii:
I still think TAU should be exported from constants, NOT added to the global Math object. I don't like modifying the global math object. It breaks encapsulation of the jscad code, and is generally considered bad practice.

There are no current proposals to add Math.TAU to the js standard, and the idea was generally opposed [here](https://esdiscuss.org/topic/math-tau). So this doesn't feel like a polyfill for future TAU support. It just feels like JSCAD intruding into the standard library.

z3dev:
> I still think TAU should be exported from constants, NOT added to the global Math object. I don't like modifying the global math object. It breaks encapsulation of the jscad code, and is generally considered bad practice.

Done. TAU is now defined in the maths contants, and exported.

@platypii if fine then I will refactor the other parts of modeling.

platypii (approved):
Looks good
 ~platytau :smile:

z3dev:
@platypii I completed all the changes / refactoring. So, the changes are now ready for a full review. All comments are welcome.

I like the code changes. The TAU constant makes all rotations easier to conceptualize, and of course the code just looks cleaner. I'm hoping that performance gains slightly as well.

packages/modeling/src/maths/mat4/fromTaitBryanRotation.js
```diff
@@ -13,7 +13,7 @@ const { sin, cos } = require('../utils/trigonometry')
  * @returns {mat4} out
  * @alias module:modeling/maths/mat4.fromTaitBryanRotation
  * @example
- * let matrix = fromTaitBryanRotation(create(), Math.PI / 2, 0, Math.PI)
+ * let matrix = fromTaitBryanRotation(create(), TAU / 4, 0, TAU / 2)
```
t1u1, line 16: I see a lot of PI/2 computations. You could declare a constant for that as well (no cool sounding name, unfortunately). It would have even more performance impact since division is much more costly than multiplication.

packages/modeling/src/maths/mat4/isOnlyTransformScale.test.js
```diff
@@ -1,9 +1,11 @@
 const test = require('ava')
 
+const { TAU } = require('../constants')
+
 const { isOnlyTransformScale, create, fromTranslation, fromTaitBryanRotation, fromScaling, invert, multiply } = require('./index')
 
 test('mat4: isOnlyTransformScale() should return true for right angles', (t) => {
-  let someRotation = fromTaitBryanRotation(create(), Math.PI, 0, 0)
+  let someRotation = fromTaitBryanRotation(create(), TAU / 2, 0, 0)
```
t1u1, line 8: This is bad for performance. Division is a very costly operation. Replacing 2π with ⲧ makes sense though.

t1u1: In case the TAU/2 expression is kept, it is much faster to multiply by 0.5  than divide by 2.

z3dev:
> @platypii I completed all the changes / refactoring. So, the changes are now ready for a full review. All comments are welcome.
> 
> I like the code changes. The TAU constant makes all rotations easier to conceptualize, and of course the code just looks cleaner. I'm hoping that performance gains slightly as well.

@platypii can you take another look at this PR? thanks

platypii (approved):
Looks good! Happy to see JSCAD leading the way for clearer mathematical notation :-)

z3dev:
> Looks good! Happy to see JSCAD leading the way for clearer mathematical notation :-)

@platypii 
HA HA HA! The idea came initially from your post. ;)
By the way, see your email... I've been pinging you to set up a time to discuss V3.

Issue #1135, opened 2022-08-08
Malformed svg path at x:y. Path closed itself with command.


hrgdavor:
## Expected Behavior
closed path from svg

## Actual Behavior
error because last line ends at the begining

## Steps to Reproduce the Problem

  1.
create simple script:
```
const { deserializers } = require("@jscad/io")
let svg = `<svg><g><path d="M0 0 L10 10L10 0L0 0" id="path4544" /></g></svg>`

const jscad = require('@jscad/modeling')
function main() {
    return deserializers.svg({ output: "geometry", segments: 128 }, svg)
}

module.exports = {main}
```
  2. run it on openjscad.xyz
  3. error is generated "Error: Malformed svg path at 2:50. Path closed itself with command # 3 L0 0"

This is what inkscape produces and is not unexpectd path in general. I was part of adding this error report for cases when svg path closes on itself somewhere in the middle, but failed to account for this valid case when start point and end point meet.

workaround is to replace last command `L 0 0` with `Z`
```js
let svg = `<svg><g><path d="M0 0 L10 10L10 0Z" id="path4544" /></g></svg>` 
```

## Specifications

  - Version: 2.5.12
  - Platform: all
  - Environment: all

z3dev:
@hrgdavor I believe the fix has been released. Please verify, and close this issue if satisfied.

hrgdavor:
works :)

PR #1142, opened 2022-09-01
fix 1135 svg start == end point throw unecessary error

hrgdavor:
fix for #1135

there is no need to throw error when last point is same as first point in svg.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ objects/
 
 # project ignores
 packages/web/examples/
+.history
--- a/packages/io/svg-deserializer/src/shapesMapGeometry.js
+++ b/packages/io/svg-deserializer/src/shapesMapGeometry.js
@@ -428,7 +428,6 @@ const expandPath = (obj, svgUnitsPmm, svgUnitsX, svgUnitsY, svgUnitsV, svgGroups
 
     if (pc !== true && paths[pathName] && paths[pathName].isClosed) {
       let coNext = obj.commands[j + 1]
-
       if (!coNext || !isCloseCmd(coNext.c)) {
         if (pathSelfClosed === 'trim') {
           while (coNext && !isCloseCmd(coNext.c)) {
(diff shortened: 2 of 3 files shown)


packages/io/svg-deserializer/src/shapesMapGeometry.js
```diff
@@ -437,7 +436,7 @@ const expandPath = (obj, svgUnitsPmm, svgUnitsX, svgUnitsY, svgUnitsV, svgGroups
           }
         } else if (pathSelfClosed === 'split') {
           newPath()
-        } else {
+        } else if(coNext){ // allow self close in the last command #1135
```
z3dev, line 439: @hrgdavor I know that this is correct syntax, however the addition of if(coNext) expression hides the fact this this is the case where pathSelfClosed === 'error'. I had to review the code in order to understand this. So, I suggest the following...
- Add a comment that the else clause is for 'error'
- Add a nested if statement to test if(coNext)

packages/io/svg-deserializer/tests/issue.1135.test.js
```diff
@@ -0,0 +1,10 @@
+const test = require('ava')
+
+const deserializer = require('../src/index.js')
+
+test('deserialize issue 885 do not fail on close at the end', (t) => {
```
z3dev, line 5: super.

packages/io/svg-deserializer/src/shapesMapGeometry.js
hrgdavor: I have changed it bit more, check if this variant is more understandabla

hrgdavor:
Hope it is ok I add a small change to .gitignore
```
.history
```
I use vscode a lot now, and there is a nice local history plugin but it prefers to store localhistory in `.history` fodler

.gitignore
```diff
@@ -17,3 +17,4 @@ objects/
 
 # project ignores
 packages/web/examples/
+.history
```
z3dev, line 20: I don't mind adding more to gitignore but what is .history?

hrgdavor: it is a folder where vscode plugin saves file copy on every save. This way when doing a lot of changes before commit, deleting piece of code, undo redo ... from time to time saves the day by going back to a prev version of a file:)

https://marketplace.visualstudio.com/items?itemName=xyz.local-history

z3dev (approved):
@hrgdavor for the nice fix and test case.

PR #1147, opened 2022-09-23
build(all): switching to PNPM as the package manager, and PNPM workspaces

z3dev:
Build related only.

### All Submissions:


z3dev:
@platypii @hrgdavor please try the changes on this branch. you should clone the repository, and checkout this branch. you should **use a new new file system** as this will use PNPM to manage the packages.

z3dev:
FYI, lerna bootstrap in no longer necessary. And will throw an error. ;)

I updated the documentation.

The installation steps are now:

- pnpm install
- pnpm test

I do have one question about PNPM. Is the installation of NPM required?

z3dev:
> do you already have a favorite for using instead of lerna (I remember you mentioning moving from lerna)?

Not really.

The next PR will move to Lerna-Lite which has been put together from pieces of Lerna, converted to Typescript, and reworked to support PNPM workspaces. Lots of functionality is now optional.

https://github.com/ghiscoding/lerna-lite

The basic requirements for the automated publishing are:
- change determination
- independent versioning
- automatic change logs

hrgdavor:
Lerna lite sounds nice and is I am guessing a simpler move. I randomly picked rush and am sticking with it in my project to learn it. I am in favor of lerna-lite for jscad because it is most likely less hassle to move, and it sounds nice.

who knows maybe after some time using both I will also prefer lerna-lite, or be undecided.

hrgdavor:
@z3dev I had issues runnung tests until I remove global lerna install that I had.
I am not 100% sure it was related, but it is wort mentioning to users about migration
https://github.com/ghiscoding/lerna-lite#migration-for-existing-lerna-users

z3dev:
> @z3dev I had issues running tests until I remove global lerna install that I had. I am not 100% sure it was related, but it is worth mentioning to users about migration.

Good point. I added V3-CHANGES.md to document the changes between V2 and V3.

z3dev:
@hrgdavor thanks for the nice feedback.

@platypii any chance that you have some band width?

hrgdavor (approved):
nice step forward :)

PR #1152, opened 2022-10-10
feat(modeling)!: converted code to ES6 import / export

z3dev:
**WORK IN PROGRESS DO NOT MERGE**

### BREAKING CHANGES

The changes move JSCAD from CommonJS to ES6 modules. Summary of progress.
* [X] modeling
* [X] core
* [X] io
* [X] cli (converted but cannot be used yet due to differences in ES6 and CommonJS)
* [X] array utils
* [X] regl-renderer

NOTE: Due to incompatibility of the 'most' libraries, the WEB UI will not be converted. A new WEB UI will be developed as part of V3.

### All Submissions:


z3dev:
Technical Notes:
- "type": "module" must be added to package.json
- all require paths need to be reviewed and changed, ending with *.js module name
- all functions are exported by name, as well as exported as default

hrgdavor:
this is a huge task. Let me know when/if you want me to test something. I am guessing this is in progress currently.

z3dev:
> this is a huge task. Let me know when/if you want me to test something. I am guessing this is in progress currently.

@hrgdavor Thanks. I'll let you know when some of the libraries are ready for testing.

There's no way to do this in pieces so ONE HUGE set of changes.

hrgdavor:
I know you have asked abou changes to jscad for v3, there is one thing that was nagging me

```js
```
and then I need to look into `jscad.booleans` to get subtract ... and 99% of the functions in to primitives, booleans etc have unique name.

it would be nicer if they were declared on top level instead:
```js
const { cube, subtract, translate} = require('@jscad/modeling')
// or import for es6 ofc
import { cube, subtract, translate} from '@jscad/modeling'
```

as a bonus, this would also improve current issue I have with vscode that prefers to generate automatic import like this
```js
const {subtract} = require('@jscad/modeling/booleans')
```
which then breaks the script

z3dev:
> I know you have asked abou changes to jscad for v3, there is one thing that was nagging me
> 
> ```js
> const jscad = require('@jscad/modeling')
> ```
> 
> and then I need to look into `jscad.booleans` to get subtract ... and 99% of the functions in to primitives, booleans etc have unique name.
> 
> it would be nicer if they were declared on top level instead:
> 
> ```js
> const { cube, subtract, translate} = require('@jscad/modeling')
> // or import for es6 ofc
> import { cube, subtract, translate} from '@jscad/modeling'
> ```

Good idea. However, the API changes will have to be completed in another set of changes.

Can you create a new issue for this, and label with V3?

z3dev:
@hrgdavor @platypii @SimonClark the conversion is complete, moving all code (except web) to ES6. here's the state of play...
- array-utile converted, all tests pass
- modeling converted, all tests pass
- io packages converted, all tests pass. HOWEVER deserializers translate objects into CommonJS code
- core converted, all tests pass. HOWEVER changes are required support designs with ES6 code bases. It may be possible to support designs with CommonJS code bases as well.
- cli converted, all tests fail. Once core is corrected then the CLI should work once again.

hrgdavor:
can we run tests from root of the monorepo ?
```
 @jscad/gcode-deserializer@1.0.0 test C:\hrg\3dp_dev\OpenJSCAD.v3\packages\io\gcode-deserializer
> ava './test.js' --verbose --timeout 2m


  ✖ Based on your configuration, 1 test file was found, but did not match the CLI arguments:

  * './test.js'

 ELIFECYCLE  Test failed. See above for more details.
```

I tried and have issue with @jscad/gcode-deserializer

hrgdavor:
One option for cli to work would be to have a cmdline option to say if it is es6 script that needs to run, or "old" .

mister-jscad (approved):
DO IT!

Issue #1162, opened 2022-11-07
Helix extrude


Loosetooth:
I was unable to find a function to help me extrude a helix. Currently, both the `extrudeRotate` and `extrudeLinear` functions are not suitable for this task.

Would it be interesting to add this functionality to JSCAD?
I think it's a pretty basic function, for example if you want to model something like a bolt or a nut.

Here is my current naive implementation:
```js
const helicalExtrude = (base, pitch, numberOfSlices, totalRotation) => {

    if (geom2.isA(base)) baseSlice = slice.fromSides(geom2.toSides(base))
    if (poly3.isA(base)) baseSlice = slice.fromPoints(poly3.toPoints(base))

    const sliceCallback = (progress, index, base) => {
        const zRotation = totalRotation / numberOfSlices * index
        const zOffset = zRotation / (2 * Math.PI) * pitch

        // create rotation matrix
        const matrix = mat4.multiply(
            // then rotate around Z axis
            mat4.fromZRotation(zRotation),
            // first rotate "flat" 2D shape from XY plane to XZ plane
            mat4.fromXRotation(Math.PI / 2)
        )

        const completeMatrix = mat4.multiply(
            // add translation
            mat4.fromTranslation([0, 0, zOffset]),
            matrix
        )

        return slice.transform(completeMatrix, base)
    }

    return extrudeFromSlices(
        {
            numberOfSlices: numberOfSlices,
            capStart: false,
            capEnd: false,
            close: false,
            callback: sliceCallback
        },
        baseSlice
    )
}
```

Which you can then use as:
```js
const pitch = 10
const totalRotation = 4 * Math.PI
const totalSlices = 2 * 32

const myCircle = translate([5, 0, 0], circle({ radius: 2.5 }))
return helicalExtrude(myCircle, pitch, totalSlices, totalRotation)
```

With the following result:
![image](https://user-images.githubusercontent.com/66420755/200348499-810fa12f-e545-450b-b60c-f0c5312cfdef.png)

This feature would improve user experience, as users do not have to 'dig' into the code any more and learn how to use `extrudeFromSlices` in order to create a helix. It would also then be possible to improve the "nuts and bolts" code example in the web UI.

The implementation could be done either as a separate function like `extrudeHelical`, or it would be possible to implement an optional `pitch` parameter for the existing `extrudeRotate` function.

What do you think?
I'm willing to work on a pull request in the coming days.

z3dev:
[...]

Issue #1012, opened 2022-02-24
extrudeRotate is broken when input polygon is clockwise


paulftw:
## Steps to Reproduce the Problem
Paste below code into `openjscad.xyz` (two triangles are rotateExtruded, but in the first case the vertices  are counterclockwise).

You'll see rendering artifacts on the bottom extrusion. Boolean operations with an "inside out" extrusion also give weird results.

```
const {
  booleans,
  extrusions,
  geometries,
  maths,
  primitives,
  transforms,
} = require('@jscad/modeling')

const { circle } = primitives
const { geom2 } = geometries
const { mat4 } = maths

const main = () => {
  return [
    transforms.translate(
      [0, 0, 20],
      extrusions.extrudeRotate(
        { segments: 32, startAngle: 0, angle: Math.PI * 0.5, overflow: 'cap' },
        primitives.polygon({
          points: [
            [-10, -0],
            [-10, -10],
            [-15, -5],
          ],
        }),
      ),
    ),

    extrusions.extrudeRotate(
      { segments: 32, startAngle: 0, angle: Math.PI * 0.5, overflow: 'cap' },
      primitives.polygon({
        points: [
          [10, 0],
          [10, 10],
          [15, 5],
        ],
      }),
    ),
  ]
}

module.exports = {main}
```

hrgdavor:
My opinion is that library should not defensively call all sorts of fixes for the input.
This would allow code like above  to work, but would punish with performance degradation those that provide good input geometry.

I would be against input cleanup, and in favor of making clear in docs that input must be ccw.

platypii:
This is expected behavior. Most computational geometry assumes that polygons are oriented in a counter-clockwise direction. This is even more important in 3D where a polygon can face in or out. However even in 2D, counterclockwise is considered a solid, and clockwise is used to represent holes, or negative space. That is why extruding a negative space does not totally make sense.

I am against "automatic flipping" of inverted polygons. It might be better to throw an exception, rather than produce a flawed extrusion. But overall this is expected. More docs are a good idea though. (for the record, I ran into the same confusion #808)

z3dev:
@paulftw right or wrong, there might be cases where someone wants to create an inverted extrusion. i know that some users are using inverted 3D shapes as 'holes' when using the booleans.

i also looked at the user guide. it doesn't mention anything about the 'handed-ness' of shapes, so probably a document change would be good.

paulftw:
[...]

PR #1164, opened 2022-11-08
add extrudeHelical

Loosetooth:
Adds the `extrudeHelical` functionality as discussed in #1162

This supposes that the `geom2` that is to be extruded is laying fully in the positive x plane. Otherwise you get issues with inside-out polygons, as discussed in #1012

It allows for the generating of a helix, tapered helix (`endRadiusOffset !== 0`) and spiral (`pitch === 0, endRadiusOffset !== 0`)

Simple example:
```js
const jscad = require("@jscad/modeling")
const { extrudeHelical } = jscad.extrusions

const main = () => {
    return extrudeHelical(
            {
                angle: Math.PI * 4,
                pitch: 10,
                endRadiusOffset: -5,
                segments: 64
            },
            circle({ size: 3, center: [10, 0] })
        )
}

module.exports = { main }
```

Resulting in a tapered helix:
image

I will need some extra guidance with adding the documentation, as I could not immediately find where I should add such documentation. For example, there's the [docs](https://openjscad.xyz/docs/) website, and the [user guide](https://openjscad.xyz/dokuwiki/doku.php). I'm not sure which one is supposed to be edited, and how I do that. Thanks for the help.

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
+++ b/packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
@@ -4,8 +4,10 @@ export default extrudeHelical
 
 export interface ExtrudeHelicalOptions {
   angle?: number
+  startAngle?: number
   pitch?: number
-  endRadiusOffset?: number
+  height?: number
+  endOffset?: number
   segments?: number
 }
 
--- a/packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
+++ b/packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
@@ -8,7 +8,7 @@ export interface ExtrudeHelicalOptions {
   pitch?: number
   height?: number
   endOffset?: number
-  segments?: number
+  segmentsPerRotation?: number
 }
 
 declare function extrudeHelical(options: ExtrudeHelicalOptions, geometry: Geom2): Geom3
--- a/packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
+++ b/packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
@@ -1,0 +1,12 @@
+import { Geom2, Geom3 } from '../../geometries/types'
+
+export default extrudeHelical
+
+export interface ExtrudeHelicalOptions {
+  angle?: number
+  pitch?: number
+  endRadiusOffset?: number
+  segments?: number
+}
+
+declare function extrudeHelical(options: ExtrudeHelicalOptions, geometry: Geom2): Geom3
--- a/packages/modeling/src/operations/extrusions/extrudeHelical.js
+++ b/packages/modeling/src/operations/extrusions/extrudeHelical.js
@@ -76,7 +76,7 @@ const extrudeHelical = (options, geometry) => {
     mat4.multiply(
       step1,
       // then apply offsets
-      mat4.fromTranslation(mat4.create(), [xOffset, 0, zOffset]),
+      mat4.fromTranslation(mat4.create(), [xOffset, 0, zOffset * Math.sign(angle)]),
       // first rotate "flat" 2D shape from XY to XZ plane
       mat4.fromXRotation(mat4.create(), -TAU / 4 * Math.sign(angle)) // rotate the slice correctly to not create inside-out polygon
     )
--- a/packages/modeling/src/operations/extrusions/extrudeHelical.js
+++ b/packages/modeling/src/operations/extrusions/extrudeHelical.js
@@ -13,7 +13,7 @@ const geom2 = require('../../geometries/geom2')
  * @param {Number} [options.pitch=10] - elevation gain for each turn
  * @param {Number} [options.height] - total height of the helix path. Ignored if pitch is set.
  * @param {Number} [options.endOffset=0] - offset the final radius of the extrusion, allowing for tapered helix, and or spiral
- * @param {Number} [options.segmentsPerRotation=32] - number of segments of the extrusion
+ * @param {Number} [options.segmentsPerRotation=32] - number of segments per full rotation of the extrusion
  * @param {geom2} geometry - the geometry to extrude
  * @returns {geom3} the extruded geometry
  * @alias module:modeling/extrusions.extrudeHelical
--- a/packages/modeling/src/operations/extrusions/extrudeHelical.js
+++ b/packages/modeling/src/operations/extrusions/extrudeHelical.js
@@ -27,55 +27,55 @@ const geom2 = require('../../geometries/geom2')
  * )
  */
 const extrudeHelical = (options, geometry) => {
-    const defaults = {
-        angle: TAU,
-        pitch: 10,
-        endRadiusOffset: 0,
-        segments: 32
-    }
-    const { angle, pitch, endRadiusOffset, segments } = Object.assign({}, defaults, options)
+  const defaults = {
+    angle: TAU,
+    pitch: 10,
+    endRadiusOffset: 0,
+    segments: 32
+  }
+  const { angle, pitch, endRadiusOffset, segments } = Object.assign({}, defaults, options)
 
-    const baseSlice = slice.fromSides(geom2.toSides(geometry))
+  const baseSlice = slice.fromSides(geom2.toSides(geometry))
 
-    const sliceCallback = (progress, index, base) => {
-        const zRotation = angle / segments * index
-        const xOffset = endRadiusOffset / segments * index
-        const zOffset = zRotation / TAU * pitch
+  const sliceCallback = (progress, index, base) => {
+    const zRotation = angle / segments * index
+    const xOffset = endRadiusOffset / segments * index
+    const zOffset = zRotation / TAU * pitch
 
-        // TODO: check for valid geometry after translations
-        // ie all the points have to be either x > -xOffset or x < -xOffset
-        // this would have to be checked for every transform, and handled
-        //
-        // not implementing, as this currently doesn't break anything,
-        // only creates inside-out polygons
+    // TODO: check for valid geometry after translations
+    // ie all the points have to be either x > -xOffset or x < -xOffset
+    // this would have to be checked for every transform, and handled
+    //
+    // not implementing, as this currently doesn't break anything,
+    // only creates inside-out polygons
 
-        // create transformation matrix
-        const step1 = mat4.create()
-        mat4.multiply(
-            step1,
-            // then apply offsets
-            mat4.fromTranslation(mat4.create(), [xOffset, 0, zOffset]),
-            // first rotate "flat" 2D shape from XY to XZ plane
-            mat4.fromXRotation(mat4.create(), -TAU / 4) // putting TAU/4 here creates inside-out polygon
-        )
-        
-        const matrix = mat4.create()
-        mat4.multiply(
-            matrix,
-            // finally rotate around Z axis
-            mat4.fromZRotation(mat4.create(), zRotation),
-            step1
-        )
-        return slice.transform(matrix, base)
-    }
-
-    return extrudeFromSlices(
-        {
-            numberOfSlices: segments,
-            callback: sliceCallback
-        },
-        baseSlice
+    // create transformation matrix
+    const step1 = mat4.create()
+    mat4.multiply(
+      step1,
+      // then apply offsets
+      mat4.fromTranslation(mat4.create(), [xOffset, 0, zOffset]),
+      // first rotate "flat" 2D shape from XY to XZ plane
+      mat4.fromXRotation(mat4.create(), -TAU / 4) // putting TAU/4 here creates inside-out polygon
     )
+
+    const matrix = mat4.create()
+    mat4.multiply(
+      matrix,
+      // finally rotate around Z axis
+      mat4.fromZRotation(mat4.create(), zRotation),
+      step1
+    )
+    return slice.transform(matrix, base)
+  }
+
+  return extrudeFromSlices(
+    {
+      numberOfSlices: segments,
+      callback: sliceCallback
+    },
+    baseSlice
+  )
 }
 
 module.exports = extrudeHelical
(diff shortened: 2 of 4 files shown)


z3dev:
@Loosetooth WOW! That's was super fast! Nice job!

I'll look over the code on the weekend. :)

The API documentation is generated from the source code. You can generate a local copy by...

```
nom run docs
open docs/index.html
```

Loosetooth:
Thanks @z3dev , looking forward to your feedback.

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,81 @@
...
+const mat4 = require('../../maths/mat4')
+const extrudeFromSlices = require('./extrudeFromSlices')
+const geom2 = require('../../geometries/geom2')
+
+/**
+ * Perform a helical extrude of the geometry, using the given options.
+ *
+ * @param {Object} options - options for extrusion
+ * @param {Number} [options.angle=TAU] - angle of the extrusion (RADIANS) positive for right-hand rotation, negative for left-hand
+ * @param {Number} [options.pitch=10] - elevation gain for each turn
```
z3dev, line 12: I like both 'pitch' and 'height'. I can see some users wanting to specify the pitch for each turn, but some users may want to set a final 'height'. I'm wondering if we could support both of these.

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param {Number} [options.endRadiusOffset=0] - offset the final radius of the extrusion, allowing for tapered helix, and or spiral
```
z3dev, line 13: I hate names... endRadiusOffset seems a little long. Does something like 'endOffset' fit?

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,81 @@
...
+const slice = require('./slice')
+const mat4 = require('../../maths/mat4')
+const extrudeFromSlices = require('./extrudeFromSlices')
+const geom2 = require('../../geometries/geom2')
+
+/**
+ * Perform a helical extrude of the geometry, using the given options.
+ *
+ * @param {Object} options - options for extrusion
+ * @param {Number} [options.angle=TAU] - angle of the extrusion (RADIANS) positive for right-hand rotation, negative for left-hand
```
z3dev, line 11: @Loosetooth is it possible to support 'startAngle'? both extrudeRotate and torus have this option.

z3dev (changes requested):
@Loosetooth code looks good but please adjust the formatting.

Also, lots of questions / thoughts about the parameters. Please provide some nice ideas. ;)

z3dev:
@Loosetooth it seems that full rotation is not being realized. here are some screen shots. i suspect that you will have a lot of fun with these 'rotations'. the extrudeRotate function has some logic to calculate rotations, which might help.

z3dev:
issue1
issue2

Loosetooth:
Hi @z3dev , thanks for the feedback.

Implementing the `startAngle` parameter shouldn't be a problem.
Full rotations should indeed be fixed, good catch. It's doing one step too little.

> I like both 'pitch' and 'height'. I can see some users wanting to specify the pitch for each turn, but some users may want to set a final 'height'. I'm wondering if we could support both of these.

Sure, this is also a parameter that some implementations in other CAD software accept. The only reason I did not implement it immediately was that it's slightly ambiguous. How do we measure the height of the helix? Do we measure from the center of the start sketch, until the center of the final extruded sketch? Or do we measure from the "bottom" of the start sketch, until the top of the final sketch?

The most straightforward implementation would be the first option, then we automatically calculate the pitch in this manner:
```js
const pitch = height / (angle / TAU)
```
In this case the total height of the helix would however be larger than the provided `height` parameter, which is perhaps confusing to users:
```js
const totalHelixHeight = height + sketchHeight
```

What do you think about this solution?

>  hate names... endRadiusOffset seems a little long. Does something like 'endOffset' fit?

Agree that the name is too long. But I'm not sure if `endOffset` would be a good choice as it is not as descriptive. Offset can be interpreted in multiple ways.  I have no other ideas right now for a name though :D If you have any other ideas please share, if we don't find anything else I don't mind to name it endOffset.

Loosetooth:
Please have a look @z3dev 

Implemented all the requested changes.
The height parameter, when specified refers to the total height of the extrusion path, not the height of the extruded solid, so should be clear to users.

Renamed `endRadiusOffset` to `endOffset`, so the basic example should now be:
```js
const jscad = require("@jscad/modeling")
const { extrudeHelical } = jscad.extrusions

const main = () => {
    return extrudeHelical(
            {
                angle: Math.PI * 4,
                pitch: 10,
                endOffset: -5,
                segments: 64
            },
            circle({ size: 3, center: [10, 0] })
        )
}

module.exports = { main }
```

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,95 @@
...
+
+  let pitch
+  // ignore height if pitch is set
+  if(!options.pitch && options.height) {
+    pitch = options.height / (angle / TAU)
+  } else {
+    pitch = options.pitch ? options.pitch : defaults.pitch
+  }
+
+  if (segments < 2) throw new Error('segments must be greater than 1')
```
z3dev, line 49: hmm... i don't see valid results coming from 2 segments. maybe minimum is 3.

z3dev (changes requested):
@Loosetooth really close now. i'm looking forward to using this. nice.

just a few more suggestions.

packages/modeling/src/operations/extrusions/extrudeHelical.d.ts
```diff
@@ -0,0 +1,12 @@
+import { Geom2, Geom3 } from '../../geometries/types'
+
+export default extrudeHelical
+
+export interface ExtrudeHelicalOptions {
+  angle?: number
+  pitch?: number
+  endRadiusOffset?: number
```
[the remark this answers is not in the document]
Loosetooth: Fixed in 86ed105

packages/modeling/src/operations/extrusions/extrudeHelical.js
Loosetooth: Thinking about this, it needs at least 3 segments for each complete revolution, in order to get a valid geometry. Even if it's a partial revolution, so we round up.

To see this, we can find an angle $\alpha$ for every number of segments $n \in \mathbb{Z}^{+}$ so that

$$
{\alpha \over n} = \pi
\Leftrightarrow
\alpha = \pi n
$$

Which means that for every number of segments we can find an angle that will project all the slices in the same plane, resulting in incorrect (flat) geometry. Requiring that the number of segments is larger than any arbitrary value can never solve this problem.

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,95 @@
...
+    const step1 = mat4.create()
+    mat4.multiply(
+      step1,
+      // then apply offsets
+      mat4.fromTranslation(mat4.create(), [xOffset, 0, zOffset]),
+      // first rotate "flat" 2D shape from XY to XZ plane
+      mat4.fromXRotation(mat4.create(), -TAU / 4) // putting TAU/4 here creates inside-out polygon
+    )
+
+    const matrix = mat4.create()
```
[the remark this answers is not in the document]
Loosetooth: Added in 25f3b7b, please let me know is this is what you meant.

z3dev:
@Loosetooth try this... i think you will have to flip the slice depending on the direction of the rotation.

```
  const s1 = extrudeHelical(
            {
                angle: constants.TAU * -1.25,
                startAngle: constants.TAU * 0.2,
                pitch: -3,
                endRadiusOffset: 0,
                segments: 64
            },
            // circle({ size: 1, center: [10, 0] })
            ellipse({ radius: [5, 1], center: [5, 0] })
        )
```

Loosetooth:
> @Loosetooth try this... i think you will have to flip the slice depending on the direction of the rotation.
> 
> ```
>   const s1 = extrudeHelical(
>             {
>                 angle: constants.TAU * -1.25,
>                 startAngle: constants.TAU * 0.2,
>                 pitch: -3,
>                 endRadiusOffset: 0,
>                 segments: 64
>             },
>             // circle({ size: 1, center: [10, 0] })
>             ellipse({ radius: [5, 1], center: [5, 0] })
>         )
> ```

True, for negative angles, the slice has to be flipped in the other direction.

This should now work without inside out polygons:
```js
const jscad = require("@jscad/modeling")
const { ellipse } = jscad.primitives
const { extrudeHelical } = jscad.extrusions


const main = () => {
    return extrudeHelical(
        {
            angle: Math.PI * 2 * -1.25,
            startAngle: Math.PI * 2 * 0.2,
            pitch: -3,
            endOffset: 0,
            segments: 64
        },
        // circle({ size: 1, center: [10, 0] })
        ellipse({ radius: [5, 1], center: [5, 0] })
    )
}

module.exports = { main }
```

It will still give bad results when the sketch is not in the positive x plane though. I should fix that :)

Loosetooth:
> @z3dev , is there an easy way to invert the 'direction' of `geom2` shapes? What I mean is that in one orientation (clockwise?) extruding the shape results in a normal geom3 polygon. When the orientation is opposite (anti clockwise?) the extrusion results in an inverted polygon.
> 
> If we can easily invert this orientation, we can detect if the shape is in the negative x plane. Then we invert the orientation, instead of changing the rotation transformations, to get a correctly extruded polygon.

Found it :)

Geom2's with all points fully negative X (zero is counted as positive) can now also be helically extruded.

z3dev:
@Loosetooth this is working really well now. super thanks.

However, while playing around with some of the different types of extrusions, I created this combination.

  ```
const s1 = extrudeHelical(
            {
                angle: constants.TAU * 10.25,
                startAngle: constants.TAU * 0.2,
                height: 20,
                endRadiusOffset: 0,
                segments: 64 * 10.25  // STRANGE??
            },
            ellipse({ radius: [2, 1], center: [5, 0] })
        )
```

It seems that the user has to calculate the number of segments for the full rotation. Not wrong, but I wanted to point this out for discussion. I was wondering if a user would rather specify the segments PER rotation instead.

Thoughts?

**Comment by Loosetooth** [comment]
Sure, let's do segments per rotation. This would also simplify some error handling in the code.

How would you like to name this parameter? I suppose `segmentsPerRotation` is too long for your liking. If we keep it `segments`, we might confuse users because it means something else in all other extrusion functions.

**Comment by z3dev** [comment]
> Sure, let's do segments per rotation. This would also simplify some error handling in the code.
> 
> How would you like to name this parameter? I suppose `segmentsPerRotation` is too long for your liking. If we keep it `segments`, we might confuse users because it means something else in all other extrusion functions.

Ha ha ha! SegmentsPerRotation is fine.

File: packages/modeling/src/operations/extrusions/extrudeHelical.js [hunk 7]
```diff
@@ -13,7 +13,7 @@ const geom2 = require('../../geometries/geom2')
  * @param {Number} [options.pitch=10] - elevation gain for each turn
  * @param {Number} [options.height] - total height of the helix path. Ignored if pitch is set.
  * @param {Number} [options.endOffset=0] - offset the final radius of the extrusion, allowing for tapered helix, and or spiral
- * @param {Number} [options.segments=32] - number of segments of the extrusion
+ * @param {Number} [options.segmentsPerRotation=32] - number of segments of the extrusion
```
z3dev: ... number of segments per rotation of the extrusion

z3dev (changes requested):
@loosetooth just a small comment change then this is ready.

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,104 @@
...
+    pitch = options.height / (angle / TAU)
+  } else {
+    pitch = options.pitch ? options.pitch : defaults.pitch
+  }
+
+  // needs at least 3 segments for each revolution
+  const minNumberOfSegments = Math.ceil(Math.abs(angle) / TAU) * 3
+
+  if (segments < minNumberOfSegments)
+    throw new Error(`For a rotation of ${angle - startAngle} radians there need to be at least ${minNumberOfSegments} segments.
```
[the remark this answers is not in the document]
Loosetooth: This is already changed, you're looking at an old commit.

packages/modeling/src/operations/extrusions/extrudeHelical.js
Loosetooth: Is already in, looking at old commit. Still had some commits after my comment, apologies.

packages/modeling/src/operations/extrusions/extrudeHelical.js
```diff
@@ -0,0 +1,104 @@
...
+  Only ${segments} segments were specified.`)
+
+  const baseSlice = slice.fromSides(geom2.toSides(geometry))
+
+
+  // define transform matrix variables for performance increase
+  let step1
```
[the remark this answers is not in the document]
Loosetooth: Thanks for the clarification, added in [f5eb972](https://github.com/jscad/OpenJSCAD.org/pull/1164/commits/f5eb9729c223ed732bc12b302a916265d04bc600)

z3dev (approved):
@Loosetooth super. This looks good now.

I don't have any more suggestions or issues. Let's see if @platypii or @hrgdavor  have any suggestions.

hrgdavor (approved):
A nice addition. Thanks for the contribution. I have been following the conversation and changes. @z3dev is much better at noticing details, and to me it looks ready for merge :)

z3dev:
@Loosetooth Thanks again for the nice addition. Sorry that it took extra time but the final outcome was really nice.

PR #1175, opened 2022-11-29
add render hotkey message in editor

### LINKED ISSUE #1174: mention `shift + enter` render keys in UI
When using https://openjscad.xyz/ it's currently not clear for new users that they have to press `shift + enter` to render their code.

Quick fix would be to mention it somewhere in the UI. For example in the editor when it's empty.
Another option would be to add a `render` button somewhere.

Not sure how easy this is. What do you think?

Loosetooth:
Fixes #1174, by adding a subtle message in the editor on how to render the code.

Currently looks like this:
image

### DIFF
--- a/packages/web/css/demo.css
+++ b/packages/web/css/demo.css
@@ -444,3 +444,11 @@ input.error::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
   width: 16px;
   height: 16px;
 }
+
+.renderHotkeyHelp {
+  position: absolute;
+  top: 0rem;
+  right: 1rem;
+  color: gray;
+  user-select: none;
+}
--- a/packages/web/src/ui/views/editor.js
+++ b/packages/web/src/ui/views/editor.js
@@ -34,7 +34,11 @@ const createWrapper = (state, callbackToStream) => {
   if (!wrapper) {
     wrapper = html`
     <section class='popup-menu' id='editor' key='editor' style='visibility:${state.activeTool === 'editor' ? 'visible' : 'hidden'}'>
-      <textarea></textarea>
+    <textarea>
+    </textarea>
+    <p style='position:absolute;top:0rem;right:1rem;color:gray;user-select:none'>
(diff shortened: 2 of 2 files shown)


z3dev:
@Loosetooth can you provide a list of browsers tested? thanks

these changes should be tested on as many browsers as possible.

Loosetooth:
I justed tested it with
* chromium
* firefox
* brave

Looks fine in all of them.

As it's just an added <p> tag with some styling, wouldn't it be supported in almost any browser? Or is that a naive thought of me?

hrgdavor (commented):
looks good, should work in any browser. Nothing exotic in css or html

hrgdavor (approved):
looks good, should work in any browser. Nothing exotic in css or html

z3dev:
> I justed tested it with
> 
> * chromium
> * firefox
> * brave
> 
> Looks fine in all of them.
> 
> As it's just an added p tag with some styling, wouldn't it be supported in almost any browser? Or is that a naive thought of me?

@Loosetooth not at all. just don't assume that every browser acts the same, even with simple tags / css. there are too many variations of layout, screen sizes, etc, and you probably cannot test everything. we will fill in where you cannot.

platypii (commented):
I'm not a fan of inline CSS styles. It would be nicer if there was a class applied to the `<p>` tag, with the styles defined in `packages/web/demo.css`

That being said, this is a really nice improvement for new users. I also tested in Chrome mobile and it looks good.

z3dev:
@Loosetooth lets move the styling into the CSS. I know that V2 is kind of on the way out, but some users may want to create a slightly different layout for a website.

Loosetooth:
@z3dev I moved the styling to the css file.

Now, there's still a problem with this implementation, as I just found out that there is a settings button, where you can edit the render hotkey :D

image

So ideally we should change the helper text to the custom render hotkey that the user has set. What do you think?

> I know that V2 is kind of on the way out...

If there's any way that these types of improvements could go directly into V3, let me know :)

PR #1176, opened 2022-12-18
additional rollups

z3dev:
These changes add rollup configurations to packages, allowing both UMD and ES ready packages to be built. The focus is on packages that are competing all tests and stable.

For the IO packages, only the modeling library is expected to be 'external'. The other dependencies are rolled up into the IO package. For example, the svg-deserializer library includes array-utils and saxes packages.

### All Submissions:


z3dev:
@platypii @Loosetooth can you take a look at these initial rollup configurations. it's all new ground for me so some feedback would be good BEFORE making additional changes.

z3dev:
@hrgdavor @platypii can you review these changes quickly? this is the next step towards publishing V3 libraries.

let me know if any other libraries need packaging. i'm actually starting to understand rollup now. ;)

hrgdavor:
@z3dev what would be a good way to test ? try to build ?

As always, I also have something unrelated too. 

One thing I discovered while trying to make vscode to work nicely with my libraries is to include source in npm publishing, and not just bundled cjs and es.
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files

this way I do not need to produce d.ts files to have nice code hints in vscode, and I can navigate library code from within a project that uses the library. That is very much in-line with how I navigate source code of libraries when coding in Java. Many times I can get more useful info from source than from the documentation.

platypii:
I also have questions of how to run / test this. I was able to build bundles by doing:

```
pnpm i
cd packages/io/x3d-deserializer
npm run build
```

I don't know if this is how @z3dev expected it to be run though, it seems like there should be a better way.

packages/io/3mf-serializer/package.json
```diff
@@ -34,8 +37,12 @@
     "fflate": "0.7.3"
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^23.0.4",
+    "@rollup/plugin-node-resolve": "^15.0.1",
```
platypii, line 41: Why does the 3mf-serializer depend on these two plugins, but most other packages don't?

packages/io/x3d-deserializer/rollup.config.js
```diff
@@ -0,0 +1,32 @@
...
+      file: 'dist/jscad-x3d-deserializer.es.js',
+      format: 'es',
+    }
+  ],
+
+  plugins: [
+    commonjs(),
+    banner('<%= pkg.description %>\n<%= pkg.name %>\nVersion <%= pkg.version %>\n<%= pkg.license %> License'),
+    versionInjector({injectInComments:{fileRegexp: /\.(html)$/},logLevel: 'warn'}),
+    nodeResolve()
```
platypii, line 30: Why does `x3d-deserializer` depend on these 4 plugins, but for example, `dxf-deserializer` only depends on 2 plugins?

packages/io/json-deserializer/rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default {
+  input: 'src/index.js',
+
+  output: [
+    {
+      file: 'dist/jscad-json-deserializer.min.js',
+      format: 'umd',
+      name: 'jscadJsonDeserializer',
+    },
```
platypii, line 13: Most other io packages have a global `jscadModeling` export. Why the inconsistency? I would prefer to not use globals if not necessary.

packages/io/3mf-serializer/package.json
z3dev: So, the 3mf serializer depends on a library called "", which is CommonJS (require based imports). Node of course understands both CommonJS and ES6 imports. But Rollup needs to find modules using plugin-node-resolve, and translate the CommonJS code to ES6 using plugin-commonjs.

Most of the packages depend on modeling and/or array-utils. These are both ES6 modules now. :)

packages/io/x3d-deserializer/rollup.config.js
z3dev: It has a dependency on 'saxes' with is CommonJS.

packages/io/json-deserializer/rollup.config.js
z3dev: Hmm... only depends on array-utils.

z3dev:
> I also have questions of how to run / test this. I was able to build bundles by doing:
> 
> ```
> pnpm i
> cd packages/io/x3d-deserializer
> npm run build
> ```
> 
> I don't know if this is how @z3dev expected it to be run though, it seems like there should be a better way.

Sorry. I haven't plugged the 'build' into lerna yet. There are still a few more things to do before publishing starts.

platypii (approved):
Seems like reasonable configs. I'm not a rollup expert, but it seems like a good starting point. I'm in favor of merging into V3 branch so we can continue making progress.

hrgdavor (approved):
looks good :)

z3dev:
@hrgdavor @platypii thanks for the reviews. Once all tests start passing then the publishing of V3 libraries can start. For V2, publishing started as 'alpha' versions. I hope to follow the same with V3.

PR #1177, opened 2022-12-20
Martinez algorithm for 2D booleans

platypii:
This PR brings the martinez polygon clipping algorithm to JSCAD v3.

The benefits include more precision:

```js
const geometry = subtract(
  square({ size: 6 }),
  square({ size: 4 })
)
```

Previously would give sides like:
```json
[
    [3.000013333333334, 3.000013333333334],
    [-3.000013333333334, -3.000013333333334]
]
```

But now gives nice clean:
```json
[
    [3, 3],
    [-3, -3]
]
```

Importantly, it never produces geometries with a "gap". All geometries produced by martinez pass validation. There are no more "skipped" validate tests for 2D operations!

### Performance:

For the union of 10k segment circles:

|  | V2 | martinez |
| --- | --- | --- |
| disjoint (x=4) | 1900ms | 50ms |
| touching (x=2) | 2900ms | 90ms |
| overlap (x=1) | 2400ms | 85ms |
| inside (x=0) | 1800ms | 55ms |

MASSIVE speedup!

### All Submissions:


hrgdavor:
Great work, looks amazing. And performance numbers are juicy :) .

I am not very good at reviewing, especially when I see u write cleaner code than I do. 

would you mind adding an issue or discussion where you could share some resources you used and your own insights while implementing this ?

platypii:
Good point @hrgdavor thanks

The majority of this is just a copy of the martinez algorithm from https://github.com/w8r/martinez

My first step after copying the code was to adjust the styles to match JSCAD. This was done using `standardx --fix` on the martinez directory. 

After the basic styles were updated, I made some tweaks to things like the import statements to match JSCAD. I was also able to remove some redundant code. For example `equals.js` in martinez is the same thing as `vec2.equals` in JSCAD so I used our version instead.

Then I needed to hook the JSCAD 2D booleans into the martinez package. The input format for martinez is the same as the output of `geom2.toOutlines`. The output of martinez is a little different than JSCAD. They have the concept of a `Contour` which contains information on the polygon outlines, in addition to information like which holes are assigned to which solids. I wrote some glue code in `martinez/index.js` and `martinez/martinez.js` which allows JSCAD geom2 to be passed in and out from the martinez library.

Once things were wired up, I started working through the tests. Most passed already, but some were failing due to small changes in precision. I went through each failed test and updated as needed. In every case, the test changed because the martinez algorithm was more accurate and cleaner geometry. In places where `geom2.validate` tests were skipped, I confirmed that all geometries now validated successfully.

The final step I took was adding attribution headers to the martinez files. The martinez repo is MIT licensed.

packages/modeling/src/operations/booleans/martinez/segment_intersection.js
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Finds the magnitude of the cross product of two vectors (if we pretend
+ * they're in three dimensions)
+ *
+ * @param {Object} a First vector
+ * @param {Object} b Second vector
+ * @private
+ * @returns {Number} The magnitude of the cross product
+ */
+const crossProduct = (a, b) => (a[0] * b[1]) - (a[1] * b[0])
```
hrgdavor, line 16: I may be wrong but think I saw this one elsewhere in jscad, even multiple times. Not sure if it is wise to hunt for it now or maybe do as a separate PR.

packages/modeling/src/operations/booleans/martinez/segment_intersection.js
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Finds the dot product of two vectors.
+ *
+ * @param {Object} a First vector
+ * @param {Object} b Second vector
+ * @private
+ * @returns {Number} The dot product
+ */
+const dotProduct = (a, b) => (a[0] * b[0]) + (a[1] * b[1])
```
hrgdavor, line 26: I may be wrong but think I saw this one elsewhere in jscad, even multiple times. Not sure if it is wise to hunt for it now or maybe do as a separate PR.

hrgdavor (approved):
I am a bit lazy reviewer. I have seen a small thing to comment on. but in general I am for this PR to go in.

z3dev:
@platypii COOL! Nice job! Thanks for the nice XMAS present.

This is going to be a huge functionality boost in the 2D space! Oh... those wicked SVG paths... prepare to be mastered.

I'll try to get this reviewed this weekend.

hrgdavor:
@platypii if you are willing to also give it a go after this one. The dude also made a offset alg that seems to work well.
https://github.com/w8r/polygon-offset . From what I remember offset in jscad is messed up so this could be a better solution that also has links to research papers. this would be great steps forward to maybe reliable extrude.

z3dev:
@platypii just one suggestion... I'm following the original projects of everything 'assimilated' into JSCAD. We should do the same for the Martinez project. I doubt there will be any major fixes but who knows.

packages/modeling/src/operations/booleans/martinez/segment_intersection.js
platypii: `vec2.cross` returns a `vec3` whereas this function actually returns the magnitude of the cross product. I am leaving this function for now to avoid unnecessary array allocations.

packages/modeling/src/operations/booleans/martinez/segment_intersection.js
platypii: Good call. I've replaced this function with `vec2.dot`.

packages/modeling/src/operations/booleans/martinez/segment_intersection.js
hrgdavor: ok, nice

packages/modeling/src/operations/booleans/martinez/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+const fromMartinez = (outlines) => {
+  if (outlines) {
+    return geom2.fromOutlines(outlines.flat())
+  } else {
+    return geom2.create()
+  }
+}
+
+export const union = (subject, clipping) => {
```
z3dev, line 26: OK... I don't mind this but just to follow what's happening with geom3 booleans, shouldn't this be called...

- booleans/unionGeom2Sub
- booleans/intersectGeom2Sub
- booleans/subtractGeom2Sub

These three little functions (and conversion utilities) are really part of JSCAD, not the martinez library.

packages/modeling/src/operations/booleans/martinez/martinez.js
```diff
@@ -0,0 +1,104 @@
+/*
```
z3dev, line 1: let's maintain the original file name; index.js

packages/modeling/src/operations/booleans/martinez/segment_intersection.js
```diff
@@ -0,0 +1,140 @@
...
+  /* eslint-disable arrow-body-style */
+
+  const toPoint = (p, s, d) => {
+    return [
+      p[0] + s * d[0],
+      p[1] + s * d[1]
+    ]
+  }
+
+  /* eslint-enable arrow-body-style */
```
z3dev, line 59: ditto

z3dev (changes requested):
@platypii wow... that was a lot of work. thanks again.

i have a few suggestions on the organization. basically, let's keep the same file names as the original. this will allow us to track / make changes IF there are any fixes from the martinez library.

which means, a slightly different approach to the unionGeom2, etc.

packages/modeling/src/geometries/geom2/fromOutlines.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Array} points - list of outlines in 2D space
+ * @returns {geom2} a new geometry
+ * @alias module:modeling/geometries/geom2.fromOutlines
+ */
+export const fromOutlines = (outlines) => {
+  if (!Array.isArray(outlines)) {
+    throw new Error('the given outlines must be an array')
+  }
+  if (outlines.length > 0 && !Array.isArray(outlines[0])) {
+    throw new Error('the given outlines must be an array of points')
```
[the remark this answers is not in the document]
platypii: updated phrasing

packages/modeling/src/operations/booleans/martinez/connect_edges.js
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * @param  {Number} pos
+ * @param  {Array.<SweepEvent>} resultEvents
+ * @param  {Object>}    processed
+ * @return {Number}
+ */
+const nextPos = (pos, resultEvents, processed, origPos) => {
+  let newPos = pos + 1
+  const p = resultEvents[pos].point
+  let p1
```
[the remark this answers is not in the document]
platypii: I agree that the code is structured in an unclear way, but I think it works. It is the original logic from martinez.

The reason it works is because `p1` is initialized if `newPos < length`, but the same condition exists on the while loop. So it won't enter the loop unless `p1` is initialized.

packages/modeling/src/operations/booleans/martinez/contour.js
```diff
@@ -0,0 +1,12 @@
+export default class Contour {
+  constructor () {
+    this.points = []
+    this.holeIds = []
+    this.holeOf = null
+    this.depth = null
```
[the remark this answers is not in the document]
platypii: This is the same logic as in martinez [[1]](https://github.com/w8r/martinez/blob/master/src/contour.js#L12).

If you look in `initializeContourFromContext` you can see that the depth always gets set. I would prefer a more immutable style personally, but doesn't seem worth rewriting.

packages/modeling/src/operations/booleans/martinez/martinez.js
platypii: Agreed

packages/modeling/src/operations/booleans/martinez/index.js
platypii: I refactored this code now to get rid of these functions

packages/modeling/src/operations/booleans/martinez/compare_events.js
```diff
@@ -0,0 +1,47 @@
...
+  if (p1[0] < p2[0]) return -1
+
+  // Different points, but same x-coordinate
+  // Event with lower y-coordinate is processed first
+  if (p1[1] !== p2[1]) return p1[1] > p2[1] ? 1 : -1
+
+  return specialCases(e1, e2, p1, p2)
+}
+
+/* eslint-disable no-unused-vars */
```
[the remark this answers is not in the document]
platypii: Removed eslint statements, although it would be nice to use `eslint` consistently on the code.

packages/modeling/src/operations/booleans/martinez/divide_segment.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param  {Queue} queue
+ * @return {Queue}
+ */
+export default function divideSegment (se, p, queue) {
+  const r = new SweepEvent(p, false, se, se.isSubject)
+  const l = new SweepEvent(p, true, se.otherEvent, se.isSubject)
+
+  /* eslint-disable no-console */
+  if (equals(se.point, se.otherEvent.point)) {
+    console.warn('what is that, a collapsed segment?', se)
```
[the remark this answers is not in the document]
platypii: I just commented this out. Let me know if you prefer removal, or some other option.

packages/modeling/src/operations/booleans/martinez/index.js
```diff
@@ -0,0 +1,139 @@
+/*
+ * Implementation of the Martinez 2D polygon clipping algorithm
+ * Copyright (c) 2018 Alexander Milevski
+ * https://github.com/w8r/martinez
```
z3dev, line 4: @platypii i think you should add a note that this was converted by YOU. this wasn't trivial and some credit should be given. ;)

z3dev (changes requested):
@platypii thanks for the changes. i think this looks ready now, and so much more readable as well. just one comment about 'credits'.

And just one more small thing... please add a small change to the README of modeling, and add the Martinez copyright notice. See the Copyrights section.

packages/modeling/src/operations/booleans/martinez/index.js
platypii: done, thanks :-)

z3dev (approved):
@platypii looks good. thanks for the late changes.

PR #1179, opened 2022-12-22
fix 1178 transparent object overlap solid objects

### LINKED ISSUE #1178: transparent object overlap solid objects
## Expected Behavior
image

## Actual Behavior
lines are drawn over the sphere and transparent cube too. This is not grid from jscad, these are actual transparent lines.
image

## Specifications
I actually introduced this bug in #1060 as I thought wrongly transparent shapes do not need depth buffer.
without depth buffer multiple overlaping transparent objects look better, but it is not actual solution to transparency,
as it causes issues like this, and it is not a viable compromise.

  - Version:
  - Platform:
  - Environment: (all)

hrgdavor:
### All Submissions:

this fixes #1178 by removing code I added previously for transparent entities

### DIFF
--- a/packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
+++ b/packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
@@ -12,7 +12,7 @@ const drawMesh = (regl, params = { extras: {} }) => {
   const { geometry, dynamicCulling, useVertexColors, color, visuals } = Object.assign({}, defaults, params)
   // let ambientOcclusion = vao(geometry.indices, geometry.positions, 64, 64)
   const ambientOcclusion = regl.buffer([])
-  
+
   // vertex colors or not ?
   const hasIndices = !!(geometry.indices && geometry.indices.length > 0)
   const hasNormals = !!(geometry.normals && geometry.normals.length > 0)
--- a/packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
+++ b/packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
@@ -53,8 +53,7 @@ const drawMesh = (regl, params = { extras: {} }) => {
     cull: {
       enable: true,
       face: cullFace
-    },
-    depth: { enable: !transparent }
+    }
   }
 
   // blending is a bit tricky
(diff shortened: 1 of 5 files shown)


platypii:
I'm trying to test this locally, but I'm not seeing the same thing you are with transparency. And when I checkout this branch, I'm not seeing any change in behavior? Can you give an example, and maybe screenshots of JSCAD (not your prototype) showing the difference?

I tried:

```js
const main = () => {
  const s1 = colorize([0.8, 0.8, 0, 0.4], cube({ }))
  const s3 = colorize([0, 0.1, 0.9, 0.6], sphere({ center: [1.2, 0.2, 1] }))
  return [s1, s3]
}
```

And got:
Screenshot at 2022-12-22 13-07-20

hrgdavor:
ok, nice test. looks like more things needed fixing because I probably worked around the issues in my prototype.

current jscad
image

with fix
image

code:
```js

const jscad=require('@jscad/modeling')
const {colorize} = jscad.colors
const {sphere,cube} = jscad.primitives

const main = () => {
  const s1 = colorize([0.8, 0.8, 0, 0.5], cube({ }))
  const s3 = colorize([0, 0.1, 0.9], sphere({ center: [1.2, 0.2, 1] }))
  return [s1, s3]
}

module.exports = {main}
```

platypii (dismissed):
Beautiful! This is a significant improvement in transparency quality! Thanks @hrgdavor 

Screenshot at 2022-12-22 14-04-27

platypii (approved):

packages/utils/regl-renderer/src/rendering/commands/drawMesh/index.js
```diff
@@ -9,14 +9,14 @@ const drawMesh = (regl, params = { extras: {} }) => {
     geometry: undefined,
     color: meshColor
   }
-  const { geometry, dynamicCulling, useVertexColors, color, transparent } = Object.assign({}, defaults, params)
-
+  const { geometry, dynamicCulling, useVertexColors, color, visuals } = Object.assign({}, defaults, params)
```
z3dev, line 12: hmm... it seems that we never had a 'default' value for transparent. let's assign a proper 'default' for visuals. maybe {}?

z3dev (changes requested):
@hrgdavor looks like a very nice fix. i just have one small suggestion. then this is ready to go.

z3dev (approved):
@hrgdavor thanks again for the nice fixes!

hrgdavor:
np, I introduced this bug 😄 in #1060

PR #1181, opened 2022-12-30
Update Usage instructions for stl-deserializer

ntrrgc:
The instructions in the README no longer work with the current codebase, as they don't match the `deserialize()` function signature.


platypii (approved):
Good catch thanks

Issue #1120, opened 2022-07-04
2D booleans are an absolute mess.


briansturgill:
## Expected Behavior
Reasonable retention of precision.

## Actual Behavior
Some of these numbers are off by 4 times EPS!!!
```javascript
[
  [ 2.999997997037038, 1.499998998518519 ],
  [ 3.999997329382717, 1.499998998518519 ],
  [ 3.999997329382717, 2.499998330864198 ],
  [ 2.999997997037038, 2.499998330864198 ]
]
[
  [ 0, 0 ],
  [ 8.000044658732051, 0 ],
  [ 8.000044658732051, 0.9999993323456793 ],
  [ 8.999993991111113, 0.9999993323456793 ],
  [ 8.999993991111113, 2.999997997037038 ],
  [ 8.000044658732051, 2.999997997037038 ],
  [ 8.000044658732051, 3.999997329382717 ],
  [ 0, 3.999997329382717 ]
]
[
  [ 1.9999986646913586, 2.999997997037038 ],
  [ 5.999995994074076, 2.999997997037038 ],
  [ 5.999995994074076, 0.9999993323456793 ],
  [ 1.9999986646913586, 0.9999993323456793 ]
]
[
  [ 3.1499978968888898, 1.8499987648395066 ],
  [ 3.349997763358026, 1.8499987648395066 ],
  [ 3.349997763358026, 1.6499988983703708 ],
  [ 3.1499978968888898, 1.6499988983703708 ]
]
[
  [ 3.3999977299753095, 2.39999839762963 ],
  [ 3.6999975296790133, 2.39999839762963 ],
  [ 3.6999975296790133, 1.9999986646913586 ],
  [ 3.3999977299753095, 1.9999986646913586 ]
]
```

## Steps to Reproduce the Problem
```javascript
const { geom2 } = jscad.geometries
const { rectangle } = jscad.primitives
const { union, subtract } = jscad.booleans
const { translate } = jscad.transforms
const { extrudeLinear } = jscad.extrusions

const main = () => {
	let g2 = subtract(rectangle({size:[8, 4], center:[4,2]}), translate([2, 1], rectangle({size:[4, 2], center:[2, 1]})));
	g2 = union(g2, translate([7, 1], rectangle({size:[2, 2], center:[1,1]})), translate([3, 1.5], rectangle({size:[1, 1], center:[0.5,0.5]})));
[...]

PR #1182, opened 2023-01-03
V3 geom2 no sides

platypii:
This PR changes geom2 from using `sides` to using `outlines`. This will improve performance, reduce memory allocations, and simplify a bunch of operations.

The main benefit is that for operations like booleans, extrusions, and expansion, the first step is to produce outlines from a geom2. To produce outlines from sides requires running a depth first search of the sides. With this PR, there is no need to run a graph traversal, because the data structure is already stored as outlines.

Credit to @briansturgill for the idea in #1120 

To make these changes work, I needed to update the following:
 - Change internal data structure of geom2
 - Move geom2.toOutlines to geom2.fromSides
 - Refactor compact binary representation of outlines
 - Update tests to use new geom2.create with outlines
 - Update tests for shuffling of the output points
 - Reduce usage of geom2.toSides
 - Fix a martinez bug that was revealed by the JSCAD test suite [[1]](https://github.com/w8r/martinez/issues/155)

Sorry for the large PR. This will help prepare for future improvements for things like the expansions which I want to look at next. Let’s make the V3 geom2 engine into a ferrari :-D

### All Submissions:


hrgdavor (approved):
Nice effort. I am very much in favor of this. 

Unrelated comment. Also for some next PR it will be I think ok to remove compactBinary. during my explorations of worker code, I think it will be just fine to do all the calculations in worker, and also IO for export, so there is no need to transfer models to main thread. TypedArrays that are used to provide webgl data for displaying is transferable and already optimal for postMessage.

packages/modeling/src/operations/booleans/martinez/connect_edges.js
```diff
@@ -99,7 +99,10 @@ const initializeContourFromContext = (event, contours, contourId) => {
     // result".
     const lowerContourId = prevInResult.outputContourId
     const lowerResultTransition = prevInResult.resultTransition
-    if (lowerResultTransition > 0) {
+    if (lowerContourId < 0) {
```
z3dev, line 102: @platypii I hate to ask this but I really feel that there will be other fixes to the Martinez code base. Therefore, let's start this off right with some a test case for this fix. I saw the new issue at the Martinez repo so hopefully this will be easy.

z3dev:
@platypii This is so cool. This is like one of the most talked about change for V2 but we opted out in order to focus on fixing the broken API (V1 was a mess). And, I think this would never be possible without the improvements made over the last year to V2. The result of this PR will be huge, and allow the 2D API to become much more useful.

packages/modeling/src/operations/booleans/martinez/connect_edges.js
platypii: Already done!

One new test in `project.test.js` that demonstrates the bug using `torus({ innerSegments: 8, outerSegments: 4})`

The other is in `unionGeom2.test.js` which is a simplified example extracted from the torus test. This is the same one that I submitted to the `martinez` repo.

I intend to improve both JSCAD and Martinez as a result of this integration :smiley:

packages/modeling/src/geometries/geom2/fromSides.js
```diff
@@ -0,0 +1,118 @@
...
+    } else {
+      vertexMap.set(edge[0], [edge])
+    }
+  })
+  return vertexMap
+}
+
+/**
+ * Create the outline(s) of the given geometry.
+ * @param {geom2} geometry - geometry to create outlines from
```
z3dev, line 43: param is sides

returns geom2

I guess this is being used internally by the booleans. is this still necessary?

packages/modeling/src/geometries/geom2/reverse.js
```diff
@@ -12,11 +11,9 @@ import toSides from './toSides.js'
  * let newgeometry = reverse(geometry)
  */
 export const reverse = (geometry) => {
-  const oldsides = toSides(geometry)
-
-  const newsides = oldsides.map((side) => [side[1], side[0]])
-  newsides.reverse() // is this required?
-  return create(newsides)
+  const reversed = clone(geometry)
+  reversed.outlines.forEach((outline) => outline.reverse())
```
z3dev, line 15: array.reverse() does an 'in-place' reversal of the contents. i think this will change the original geometry as well.

packages/modeling/src/geometries/geom2/clone.test.js
```diff
@@ -4,7 +4,7 @@ import { clone, create, fromPoints } from './index.js'
 
 test('clone: Creates a clone on an empty geom2', (t) => {
   const expected = {
-    sides: [],
+    outlines: [],
```
z3dev, line 7: Didn't clone() change?

packages/modeling/src/geometries/geom2/toSides.js
```diff
@@ -11,6 +11,15 @@ import applyTransforms from './applyTransforms.js'
  * @example
  * let sharedsides = toSides(geometry)
  */
-export const toSides = (geometry) => applyTransforms(geometry).sides
+export const toSides = (geometry) => {
```
z3dev, line 14: I think that applyTransforms should be called before constructing the sides. Every to* function should be calling applyTransforms() to make sure the outlines(points) are ready for export.

packages/modeling/src/geometries/geom2/type.d.ts
```diff
@@ -1,11 +1,11 @@
-import Vec2 from '../../maths/vec2/type'
+import Poly2 from '../../maths/poly2/type'
 import Mat4 from '../../maths/mat4/type'
 import { Color } from '../types'
 
 export default Geom2
 
 declare interface Geom2 {
-  sides: Array<[Vec2, Vec2]>
+  outlines: Array<Poly2>
```
z3dev, line 8: probably not correct.

packages/modeling/src/geometries/geom2/validate.js
```diff
@@ -18,13 +18,16 @@ export const validate = (object) => {
...
-  // check for closedness
-  toOutlines(object)
-
-  // check for self-edges
-  object.sides.forEach((side) => {
-    if (vec2.equals(side[0], side[1])) {
-      throw new Error(`geom2 self-edge ${side[0]}`)
+  object.outlines.forEach((outline) => {
+    if (outline.length < 3) {
+      throw new Error('geom2 outlines must contain at least 3 points')
```
z3dev, line 23: how about adding the 'index' of the bad outline to the error message.

packages/modeling/src/geometries/geom2/validate.js
```diff
@@ -18,13 +18,16 @@ export const validate = (object) => {
...
+    }
+    // check for duplicate points
+    for (let i = 0; i < outline.length; i++) {
+      const j = (i + 1) % outline.length
+      if (vec2.equals(outline[i], outline[j])) {
+        throw new Error(`geom2 self-edge ${outline[i]}`)
```
z3dev, line 29: kind of criptic. something like... geom2 outline N found duplicate point

packages/modeling/src/geometries/poly2/create.js
```diff
@@ -18,7 +18,7 @@ export const create = (vertices) => {
   if (vertices === undefined || vertices.length < 3) {
     vertices = [] // empty contents
   }
-  return { vertices: vertices }
+  return { vertices }
```
z3dev, line 21: let's revert and fix in another PR

packages/modeling/src/operations/booleans/martinez/index.js
```diff
@@ -76,14 +77,31 @@ const compareBBoxes = (subject, clipping, sbbox, cbbox, operation) => {
...
+/*
+ * Convert martinez data structure to geom2
+ */
+const fromOutlines = (outlines) => {
+  outlines.forEach((outline) => {
+    if (vec2.equals(outline[0], outline[outline.length - 1])) {
+      outline.pop() // first == last point
+    }
+  })
+  outlines = outlines.filter((o) => o.length >= 3)
```
z3dev, line 101: How about a note on why the filter is required?

packages/modeling/src/geometries/geom2/fromPoints.js
```diff
@@ -20,16 +20,9 @@ export const fromPoints = (points) => {
...
   }
   // adjust length if the given points are closed by the same point
-  if (vec2.equals(points[0], points[length - 1])) --length
+  if (vec2.equals(points[0], points[length - 1])) points.length--
 
-  const sides = []
-  let prevpoint = points[length - 1]
-  for (let i = 0; i < length; i++) {
-    const point = points[i]
-    sides.push([vec2.clone(prevpoint), vec2.clone(point)])
```
z3dev, line 29: hmm... i'm wondering if this function should clone the points. i saw the cloning of vectors in project() below, so wanted to point this out for discussion.

packages/modeling/src/operations/extrusions/project.js
```diff
@@ -48,7 +49,12 @@ const projectGeom3 = (options, geometry) => {
   projpolys = projpolys.sort((a, b) => poly3.measureArea(b) - poly3.measureArea(a))
 
   // convert polygons to geometry, and union all pieces into a single geometry
-  const projgeoms = projpolys.map((p) => geom2.fromPoints(p.vertices))
+  const projgeoms = projpolys.map((p) => {
+    // This clones the points from vec3 to vec2
+    const cloned = p.vertices.map(vec2.clone)
```
z3dev, line 54: Why doesn't geom2.fromPoints() clone the vectors?

packages/modeling/src/operations/modifiers/snap.js
```diff
@@ -20,11 +20,12 @@ const snapPath2 = (geometry) => {
...
-  const sides = geom2.toSides(geometry)
-  let newsides = sides.map((side) => [vec2.snap(vec2.create(), side[0], epsilon), vec2.snap(vec2.create(), side[1], epsilon)])
-  // snap can produce sides with zero (0) length, remove those
-  newsides = newsides.filter((side) => !vec2.equals(side[0], side[1]))
-  return geom2.create(newsides)
+  const outlines = geom2.toOutlines(geometry)
+  const newOutlines = outlines.map((outline) => {
+    return outline.map((point) => vec2.snap(vec2.create(), point, epsilon))
+  })
+  // TODO: remove duplicate points, and zero-area outlines
```
z3dev, line 27: this should probably be completed before moving on to the IO library modifications.

packages/modeling/src/primitives/polygon.js
```diff
@@ -58,13 +58,13 @@ export const polygon = (options) => {
   const allpoints = []
   listofpolys.forEach((list) => list.forEach((point) => allpoints.push(point)))
 
-  let sides = []
+  let outlines = []
   listofpaths.forEach((path) => {
     const setofpoints = path.map((index) => allpoints[index])
     const geometry = geom2.fromPoints(setofpoints)
-    sides = sides.concat(geom2.toSides(geometry))
+    outlines = outlines.concat(geometry.outlines)
```
z3dev, line 65: I suggest just appending each set of points to the outlines, and return / create the geom2

z3dev (changes requested):
@platypii great start but i have a few change requests, and a few questions as well.

packages/modeling/src/geometries/geom2/fromSides.js
platypii: geom2.fromSides is still needed in 7 places, for expansion and extrusion. My goal is to remove these uses, but it will require refactoring, and this PR was already long enough. I could remove it from the index, so that it is internal use only?

packages/modeling/src/geometries/geom2/clone.test.js
mister-jscad: The test changed so I thought that the clone() function must have changed as well.

z3dev: I assumed that clone() changed because the test has changed.

platypii: Clone didn't need to change because it just copies the whole object structure, regardless of the internal representation.
```js
export const clone = (geometry) => Object.assign({}, geometry)
```

packages/modeling/src/operations/booleans/martinez/index.js
```diff
@@ -76,14 +77,32 @@ const compareBBoxes = (subject, clipping, sbbox, cbbox, operation) => {
...
+      outlines.push([...outline, outline[0]])
+    }
   })
   return [outlines]
 }
 
+/*
+ * Convert martinez data structure to geom2
+ */
+const fromOutlines = (outlines) => {
```
z3dev, line 95: Just asking... but is there an issue with geom2 if the first point == last point?

platypii: There is no "issue". But martinez always produces outlines with first == last. So if you run a bunch of boolean operations in sequence, you can end up with many duplicate points at the end of the array. Since JSCAD does not require first == last, I remove the duplicates in order to produce the cleanest, simplest geometry that we can.

packages/modeling/src/geometries/geom2/toSides.js
platypii: The `toSides` function calls `toOutlines`, and that function calls `applyTransforms`. So this is already handled.

packages/modeling/src/geometries/geom2/fromPoints.js
platypii: My opinion is that the correct action is never copy/clone unless necessary. That is kind of the core philosophy of an immutable data structure library. If we clone on every create call, we will never get benefits of shared vertex references, and there will be a huge number of unnecessary allocations.

hrgdavor: I would agree with that. With immutability we must copy on write, but no need to copy defensively. So copy is made of parts that would otherwise mutate.

z3dev: The usage is slightly different, as the points provided are those of the user, i.e. not part of a geometry. The create() function definitely expects the list of outlines to be immutable, which implies that the points are also immutable. But this is not the case for fromPoints(), where the provided points are outside the domain of the geometry.

By the way, we have changed some of the internal logic to create points, and use geom.create() instead of geom.fromPoints(). There's a significant performance boost due to those changes.

Also, the geometries should be consistent. The other geometries are cloning vectors.

packages/modeling/src/geometries/geom2/index.js
```diff
@@ -15,8 +15,8 @@
  */
```
z3dev, line 15: Doc needs changes.

packages/modeling/src/operations/booleans/martinez/index.js
z3dev: @platypii cool. let's make sure the documentation mentions that each outline is closed between first and last points.

packages/modeling/src/operations/extrusions/project.js
z3dev: this should be geom2.create(cloned)

packages/modeling/src/geometries/geom2/fromPoints.js
platypii: There are many places internal to JSCAD which also use `geom2.fromPoints`.

Also, I don't see documentation anywhere telling the user that `create` can use its input as part of its internal structure, but that `fromPoints` will clone everything. Both are available to users.

In fact, for `geom3`, both `create` and `fromPoints` will reuse user provided data without cloning. So I think the consistent behavior would be to assume immutability as I've done here.

I feel pretty strongly that we should assume the user will NOT modify the inputs they pass to JSCAD. We should document that assumption, and if a user really wants to modify things, they should clone them before passing to JSCAD. This gives us opportunity for maximum performance. And I think it is reasonable for users to assume if they modify things, it might affect the geometry.

I'll change it if you want in order to get the PR merged, but I really think it's better as-is.

z3dev: @platypii let's be consistent as least in functionality/design/documentation for the geometries.

- create(contents) expects the contents to be immutable (DONE)
- from*(data) functions should convert the data to into contents (the data is reused if possible and should not be modified), and then call create
- get rid of from*(data) functions that are just duplicating create()

IF another PR is required to correct the geometries then please work on that next as consistent geometry data structures / functions will improve V3 in many ways.

P.S. slice() needs to become a geometry as well, moving this into geometries.

platypii: I have a subsequent PR ready which removes all uses of `fromPoints` which I'll put up next

z3dev (changes requested):
@platypii getting close now. sorry for the nits.

packages/modeling/src/operations/extrusions/project.js
platypii: Martinez doesn't crash, but some of the points returned will be improperly Vec3 instead of Vec2

packages/modeling/src/geometries/geom2/fromPoints.js
platypii: I think it should be removed entirely in the next PR because it is redundant with create.

packages/modeling/src/geometries/geom2/reverse.js
```diff
@@ -1,5 +1,4 @@
-import create from './create.js'
-import toSides from './toSides.js'
+import clone from './clone.js'
 
 /**
  * Reverses the given geometry so that the sides are flipped in the opposite order.
```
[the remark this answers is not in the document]
platypii: grepped and fixed some other "sides" :+1:

packages/modeling/src/operations/extrusions/project.js
z3dev: @platypii i suspect that this will have to be addressed later. but for now, let's go with this.

z3dev (approved):
@platypii super job. thanks again.

i think those little discussions were worth the extra time as well.

platypii:
> i think those little discussions were worth the extra time as well.

Definitely. Thanks for the detailed reviews!

hrgdavor:
@z3dev thanks for the detailed reviews. I am bit lazy to look that much in detail, and I am more messy than @platypii so I rarely find things to improve :)

PR #1187, opened 2023-01-09
feat(io): rework io package to incorporate 'core' serialize and deserialize functions

z3dev:
These changes rework the 'io' package to expose to new functions; deserialize() and serialize(). Previously, these functions were mashed into CLI / CORE / WEB packages as different functions returning different things. These are part of the new building blocks for applications.

Test suites were added to make sure the new functions work as expected.

In support of the new function, the deserializers and serializers have been modified to export mimeType and de/serialize functions. By using mimeType, users can quickly convert to any desired format.

An additional utility function was added to assist users to find the correct mime type using an extension; getMimeType()

### All Submissions:


z3dev:
@platypii please review. this is part of the reorg of packages. io needs to stand alone, and provide both serialize and deserialize functionality. previously, these functions were mashed into CLI and CORE code. these changes try to expose an easy API to do the same, via one package; io

z3dev:
@hrgdavor please review. i'm wondering if the new deserialize() and serialize() functions will help with the new WEB UI. i hope so. all feedback is welcome.

packages/io/io/src/serialize.js
```diff
@@ -0,0 +1,36 @@
+import { flatten } from '@jscad/array-utils'
+
+import { serializers } from './serializers.js'
+
+/**
+ * Serialize the given objects as per the given mimeType into the external format.
+ * Options can be provided to over-ride or suppliment the defaults used during serialization.
+ * See each serializer package for available options.
+ * @param {Object} options - options used during deserializing
+ * @param {String} mimeType - MIME type of the file
```
platypii, line 10: Should mimeType be one of the options fields?

packages/io/io/README.md
```diff
@@ -26,7 +26,6 @@ This package is a metapackage and includes all the input/output format handling
 ie: file data => JSCAD code or JSCAD geometry
 
 Following formats are supported as inputs
-- [AMF](../amf-deserializer)
```
platypii, line 29: Why no AMF? I know it's not very popular. But it is meant to be a "better" STL file. And the converter is already written. I'm fine with removal, just curious.

packages/io/io/src/serialize.js
```diff
@@ -0,0 +1,36 @@
...
+  options = Object.assign({}, defaults, options)
+
+  objects = flatten(objects)
+
+  if (mimeType in serializers) {
+    const serializer = serializers[mimeType]
+    const data = serializer(options, ...objects)
+    return { data, mimeType }
+  }
+  throw new Error(`Unknown mime type (${mimeType})`)
```
platypii, line 33: nit: I would put the `throw` in an else clause

packages/io/stl-serializer/src/index.js
```diff
@@ -49,6 +49,6 @@ const serialize = (options, ...objects) => {
 }
 
 export {
-  mimeType,
-  serialize
+  serialize,
```
platypii, line 52: It was alphabetical, why change order?

packages/io/io/src/formats.js
```diff
@@ -113,22 +104,6 @@ const supportedFormats = {
   }
 }
 
-// handled input formats that can be converted to jscad code
-const conversionFormats = [
```
platypii, line 117: conversionFormats is exported from `formats.js` and re-exported in `index.js` but does not exist. This throws an error when I try to run tests.

platypii (changes requested):
Looks like cool changes. But tests threw an error.

packages/io/io/README.md
hrgdavor: My knowledge is limited about AMF, but I have heard from dude that worked on 3mf, that AMF failed in the sense to deliver on those promises in practice. Still I am also curious why remove it ?

packages/io/stl-deserializer/src/index.js
```diff
@@ -410,9 +410,9 @@ const solid${index} = () => {
   return src
 }
 
-const extension = 'stl'
+const mimeType = 'application/sla'
```
hrgdavor, line 413: Hm, mimeType is important, but having extension available conceptually is not bad, to allow for automatic addition of extension when exporting to a file (with save dialog). Is extension info removed, or just moved elsewhere ?

hrgdavor (dismissed):
I am for separation even though for WEB ui to work I need both modeling and io packages inside worker. Still, separation is better. If needed, I can start worker with modeling, and load IO package first time export is called.

By the way, new WEB  worker can run `import` es module,s, old `require` cjs, and also TypeScript. Io is done in the worker, so there is no need for comapctBinary. Worker sends weblg buffers to main, but all the processing is in the worker.

packages/io/stl-deserializer/src/index.js
z3dev: @hrgdavor excellent question.

so, the deserializers/serializers work on a specific file format, which is designated by the 'mimeType'. it's kind of a standard.

the file extensions can be found in the supportedFormats, and there are functions to look up formats for input. output, etc.

there's a new function called getMimeType(extension). would the opposite be useful? getExtension(mimeType)?

packages/io/io/README.md
z3dev: > Why no AMF? I know it's not very popular. But it is meant to be a "better" STL file. And the converter is already written. I'm fine with removal, just curious.

The simple answer is "maintenance". I don't want to waste time and effort maintaining this further. We have done this for other formats (gcode, scad) so there's nothing new here.

FYI, I will be removing gcode and scad from the V3 branch. If someone wants these then let them port V1 to V3, with tests suites as well.

packages/io/io/src/serialize.js
z3dev: I considered that but thought the options should be more focused on the deserialiing / serializing. For example, segments for deserilizing, or unit for serializing.

z3dev:
@platypii what do you think about moving this to another namespace? i've been wondering how to reorg the current 'core' package, and think that some smaller packages for parameters, fake file system, loading, io, etc. should be create.

this 'io' package really feels like 'core-io' now, creating the glue between APPS and IO libraries

platypii (dismissed):
Changes look good now.

As far as moving to another package, I don't have a strong opinion on that. I've never really liked "core" because it's not clear exactly what that means. I think the serialize/deserialize makes sense in IO package. I don't want to have a million tiny packages, I think its better to have a couple well defined standalone packages. :man_shrugging:

packages/io/stl-deserializer/src/index.js
hrgdavor: getExtension(mimeType) would be useful, but it is not mission critical. It is not a big deal when setting up download menu in UI to define myself for each serializer->extension. But a thing like this would maybe allow more generic approach.

hrgdavor (dismissed):
looks good. Also with es-modules, tree-shaking works well even with aggregate packages. So in that sense It leaves for more flexibility with this type of decisions.

z3dev:
> As far as moving to another package, I don't have a strong opinion on that. I've never really liked "core" because it's not clear exactly what that means. I think the serialize/deserialize makes sense in IO package. I don't want to have a million tiny packages, I think its better to have a couple well defined standalone packages. 🤷‍♂️

Ok. Let me ask this... are any functions in io-utils out of place? I think some of those would be better inside this package, and make io-utils remain with only common utility functions.

z3dev:
@hrgdavor @platypii please review again

the getExtension(mimeType) function was added

several mime types were corrected to match those of the IANA media types, which has replaced mime types in the exchange of data.

https://www.iana.org/assignments/media-types/media-types.xhtml

platypii (approved):
Looks good.

I don't have specific suggestions for package changes. I'll think about it though as I'm hacking on the codebase.

Issue #215, opened 2017-01-10
Error trying to render a cookie cutter


Moosieus:
The Error displayed is the following:
```
Syntax error
Line: 10,col: 8
(blob:1E053920-8CA1-4E3A-910C-C6F37A44857D)
```
The object is this: 
```
//!OpenSCAD
edgeThickness = 0.4;
cutter_sizes = 4;
wallHeight = 17;
wallFlareWidth = 14;
wallFlareThickness = 2;
minWallWidth = .8;
maxWallWidth = .8;
insideWallFlareWidth = 1;
insideWallFlareThickness = 1.5;
minInsideWallWidth = 1;
maxInsideWallWidth = 3;
featureHeight = 14;
minFeatureThickness = 0.8;
maxFeatureThickness = 0.8;
connectorThickness = 2;
size = 25.4*cutter_sizes;

module dummy() {}

scale = size/25.108;

module ribbon(points, thickness=1, closed=false) {
    p = closed ? concat(points, [points[0]]) : points;
    
    union() {
        for (i=[1:len(p)-1]) {
            hull() {
                translate(p[i-1]) circle(d=thickness, $fn=8);
                translate(p[i]) circle(d=thickness, $fn=8);
            }
        }
    }
}

module cookieCutter() {

path0=scale*[[-2.604,305.284],[-2.961,290.904],[-3.050,283.568],[-3.006,280.906],[-2.961,280.545],[2.805,288.678],[1.814,291.470],[-0.099,297.214],[-0.906,299.960],[-1.349,301.962],[-1.373,302.531],[-1.231,302.729],[-0.898,302.494],[0.269,300.934],[0.810,300.408],[1.276,300.152],[1.672,300.132],[2.005,300.313],[2.280,300.661],[2.678,301.717],[2.911,303.022],[3.057,305.653],[-2.604,305.284]];
render(convexity=10) linear_extrude(height=(wallHeight)) ribbon(path0,thickness=min(maxWallWidth,max(0.001,minWallWidth)));
render(convexity=10) linear_extrude(height=(wallHeight+1)) ribbon(path0,thickness=edgeThickness);
difference() {
 render(convexity=10) linear_extrude(height=wallFlareThickness) ribbon(path0,thickness=wallFlareWidth);
 translate([0,0,-0.01]) linear_extrude(height=wallFlareThickness+0.02) polygon(points=path0);
}
}

translate([0,-280.545*scale,0]) cookieCutter();

```    
[...]

PR #1191, opened 2023-01-12
test(modeling): V3 check for self-intersecting geom2

platypii:
Adds a geom2 validation check for self-intersecting polygons. This is a helpful check as I continue to improve the 2D operations.

I added validate.test.js with validation tests for geom2 as well.

I had to add an extra parameter to maths/utils/intersect.js. This actually makes it consistent with the intersect routine in martinez/segment_intersection.js. Hopefully they can be merged one day, but not yet.

### All Submissions:

### DIFF
--- a/packages/modeling/src/maths/utils/intersect.js
+++ b/packages/modeling/src/maths/utils/intersect.js
@@ -6,11 +6,11 @@
  * @param {vec2} p2 - second point of first line segment
  * @param {vec2} p3 - first point of second line segment
  * @param {vec2} p4 - second point of second line segment
- * @param {Boolean} noEndpointTouch - skip intersections at segment endpoints
+ * @param {Boolean} endpointTouch - include intersections at segment endpoints
  * @returns {vec2} intersection point of the two line segments, or undefined
  * @alias module:modeling/maths/utils.intersect
  */
-export const intersect = (p1, p2, p3, p4, noEndpointTouch) => {
+export const intersect = (p1, p2, p3, p4, endpointTouch = true) => {
   // Check if none of the lines are of length 0
   if ((p1[0] === p2[0] && p1[1] === p2[1]) || (p3[0] === p4[0] && p3[1] === p4[1])) {
     return undefined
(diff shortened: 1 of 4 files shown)


packages/modeling/src/maths/utils/intersect.js
```diff
@@ -6,10 +6,11 @@
  * @param {vec2} p2 - second point of first line segment
  * @param {vec2} p3 - first point of second line segment
  * @param {vec2} p4 - second point of second line segment
+ * @param {Boolean} noEndpointTouch - skip intersections at segment endpoints
  * @returns {vec2} intersection point of the two line segments, or undefined
  * @alias module:modeling/maths/utils.intersect
  */
-export const intersect = (p1, p2, p3, p4) => {
+export const intersect = (p1, p2, p3, p4, noEndpointTouch) => {
```
z3dev, line 13: I'm not sure which is better... another parameter or another function.

If parameter then the default should be the previous behavior; includeEndPoints = false

If functions then there could be 'intersectWithin()' and 'intersect()', or something like this.

By the way, i think the original description is wrong. '..., end points **excluded**' is correct.

z3dev (commented):
@platypii sorry to be trouble but wondering about the correct approach for intersect(). let's discuss as the proposed change has raised a few questions about the current function.

packages/modeling/src/maths/utils/intersect.js
platypii: The previous behavior is to INCLUDE endpoints. That is why I made the parameter `noEndpointTouch` because the default previous behavior was YES endpoint touch. So the original description was right.

I can imagine a few ways to approach this:

1. `noEndpointTouch = false` as the default. This is essentially how I made this PR (since the default is actually undefined). I could add the default `= false` though.
2. `endpointTouch = true` as the default. This is fine, just inverted from `segment_intersection` in martinez.
3. Split into `intersect` and `intersectInside` or some other variation. It would be nice if we can re-use the same code though for both.

I'm personally fine with any of the options. I think my preference is for option 2 or option 1. But let me know what you prefer and I'll change it. They are all fine to me.

z3dev: Ok. Let's go with option #215

platypii: Ok I updated to use the `endpointTouch` optional parameter. Let me know what you think.

z3dev (approved):
@platypii ok. let's go with this.

packages/modeling/src/maths/utils/intersect.js
youssefsharief: @platypii Did we update the typescript definitions to accept the extra boolean?

platypii: Good catch, I missed that. I will update in a new PR. Thanks @youssefsharief

PR #1193, opened 2023-01-15
refactor(modeling)!: V3 convert slice to use contours

platypii:
This is an exciting one! This converts slice to use a list of outlines instead of edges. This will save a bunch of time and memory when converting from geom2 to slice, especially now that geom2 uses outlines.

In the process, I removed the `slice.repair` function since there can't be gaps in a point list.

I'm open to suggestions on the name `parts`. I just don't want to re-use `outlines` since it conflicts with geom2. Another option could be `contours`?

I know the plan is to move slice into geometries. I suggest we do that in a followup PR. Doing it in this PR would make it hard to read the diff.

### Performance

Here's the fun part! When extruding a 50k segment circle, the performance goes from:

| 2.5.2 (1 year ago) | V2 | V3 | V3 with slice.parts |
| --- | --- | --- | --- |
| 24500ms | 580ms | 570ms | 360ms |

About a 40% improvement in extrusion speed!

### All Submissions:


z3dev:
> I'm open to suggestions on the name `parts`. I just don't want to re-use `outlines` since it conflicts with geom2. Another option could be `contours`?

I like contours as that at least gives a hint on what the structure contains. parts is too generic. Can you rework?

platypii:
> I like contours as that at least gives a hint on what the structure contains. parts is too generic. Can you rework?

Done. I agree that `contours` is a better name.

packages/modeling/src/operations/extrusions/slice/equals.js
```diff
@@ -1,27 +1,29 @@
...
 
-  const isEqual = aedges.reduce((acc, aedge, i) => {
-    const bedge = bedges[i]
-    const d = vec3.squaredDistance(aedge[0], bedge[0])
-    return acc && (d < Number.EPSILON)
-  }, true)
+  const len = a.contours.length
+  for (let i = 0; i < len; i++) {
+    for (let j = 0; j < len; j++) {
+      const aPoint = a.contours[i]
```
z3dev, line 18: nit. move above the j for loop for performance

packages/modeling/src/operations/extrusions/slice/reverse.js
```diff
@@ -3,23 +3,14 @@ import create from './create.js'
...
-  } else {
-    out = params[0]
-    slice = params[1]
-  }
-  // reverse the edges
-  out.edges = slice.edges.map((edge) => [edge[1], edge[0]])
+export const reverse = (slice) => {
+  const out = create()
+  // reverse the contours
+  out.contours = slice.contours.map((contour) => contour.slice().reverse())
```
z3dev, line 13: it would be cleaner just to return create(contours)

z3dev (changes requested):
@platypii just a few minor requests.

also, i just wanted to check that the extrusions are still water tight. have you checked the results in meshlab?

platypii:
Yes the extrusions are water tight. Can see from the tests that they pass validation and produce the same points as before.

z3dev (approved):
@platypii nice first set of changes. thanks.

PR #1194, opened 2023-01-15
refactor(modeling)!: V3 convert poly2 vertices to points

platypii:
This PR changes `poly2.vertices` to `poly2.points`.

The problems is that otherwise there is no difference between the structure of poly2 and poly3. So you would have the following weirdness:

```js
poly2.isA( poly3.create() ) == true
poly3.isA( poly2.create() ) == true
```

By using `points` instead of `vertices` we can tell the difference between a poly2 and a poly3.

More generally, I like the idea of using "points" to refer to 2D points, and "vertices" to refer to 3D points. I think it could help make JSCAD code easier to understand if we are consistent about that.

I also added tests for `poly2.reverse` and `poly2.toPoints` since I noticed that they were not covered by existing tests.

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/extrusions/earcut/assignHoles.js
+++ b/packages/modeling/src/operations/extrusions/earcut/assignHoles.js
@@ -1,6 +1,6 @@
 import area from '../../../maths/utils/area.js'
-import toOutlines from '../../../geometries/geom2/toOutlines.js'
-import arePointsInside from '../../../geometries/poly2/arePointsInside.js'
+import { toOutlines } from '../../../geometries/geom2/index.js'
+import * as poly2 from '../../../geometries/poly2/index.js'
 
 /*
  * Constructs a polygon hierarchy of solids and holes.
@@ -41,7 +41,7 @@ export const assignHoles = (geometry) => {
     holes.forEach((h, j) => {
       const hole = outlines[h]
       // check if a point of hole j is inside solid i
-      if (arePointsInside([hole[0]], { vertices: solid })) {
+      if (poly2.arePointsInside([hole[0]], poly2.create(solid))) {
         children[i].push(h)
         if (!parents[j]) parents[j] = []
         parents[j].push(i)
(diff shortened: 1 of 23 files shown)


packages/modeling/src/operations/extrusions/earcut/assignHoles.js
```diff
@@ -1,6 +1,6 @@
 import area from '../../../maths/utils/area.js'
-import toOutlines from '../../../geometries/geom2/toOutlines.js'
-import arePointsInside from '../../../geometries/poly2/arePointsInside.js'
+import { toOutlines } from '../../../geometries/geom2/index.js'
```
z3dev, line 2: In general, internal logic should import functions directly, which prevents circular dependencies. (It happens)

why the change?

z3dev (changes requested):
@platypii just one little nit.

packages/modeling/src/operations/extrusions/earcut/assignHoles.js
platypii: I was trying to be more consistent with what I have seen elsewhere in JSCAD. It seems like the preferred pattern is to use destructuring to import from index.js files. This seems better because it is using the JSCAD standard API, not making assumptions about where files are located. I can revert this change if you prefer though.

platypii: If you look at existing imports of geometry functions inside modeling, almost all of them import from index.js. In fact `assignHoles.js` is basically the only place that imported directly. That's why I changed it.

```bash
grep -rn import.*from.*geometries packages/modeling/src
```

z3dev: Yup. My mistake.

It's the general index that internals should not use, i.e. geometries/index.js, etc.

z3dev (approved):
@platypii do you ever sleep? :)

PR #1197, opened 2023-01-17
test(modeling): tests for area and volume of operations

platypii:
Suppose you're a developer for JSCAD. And you are making changes to the modeling library, and they are good changes. But as a result of changes, the order of points change, without actually changing the shape of the geometry. But the JSCAD test suite makes assumptions about point order. So you have to then update a bunch of tests with new point data to match.

This is all fine. But wouldn't it be nice to test against properties of the geometries that do not depend on point order? So you can be more confident that we aren't breaking the geometries. If you see points change, it's not necessarily a problem. If the volume changes, something is definitely wrong.

So this PR adds `measureArea` and `measureVolume` assertions to the existing tests on all operations. [V3]

### All Submissions:


z3dev (approved):
@platypii these look fine to me. thanks.

PR #1199, opened 2023-01-22
feat(cli): reworked logic to use new io package exports and functions

z3dev:
These changes do a few things:
- fixed a few of the io package functions
- fixed core package imports of io package functions
- reworked CLI logic to use the new io package exports and functions
- converted CLI tests to use import
- HACK of core to use dynamic import

### So... tests pass now. But it's a hack. But it's a start.

What was learned...
- dynamic import() is the only way to dynamically load designs with import / export
- dynamic import() is asynchronous, therefore calling functions must be async as well... as well as...

What's missing...
- no idea if dynamic import() works in all browsers
- no idea if dynamic import() works in web workers
- no idea if dynamic import() can import whole projects
- no idea if designs with require / module and import / export can be supported

### All Submissions:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -39,6 +39,7 @@
     "@jscad/array-utils": "workspace:2.1.4",
     "@jscad/core": "workspace:2.6.4",
     "@jscad/io": "workspace:2.4.3",
+    "@jscad/io-utils": "^2.0.23",
     "@jscad/modeling": "workspace:2.10.0"
   },
   "devDependencies": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -39,7 +39,7 @@
     "@jscad/array-utils": "workspace:2.1.4",
     "@jscad/core": "workspace:2.6.4",
     "@jscad/io": "workspace:2.4.3",
-    "@jscad/io-utils": "^2.0.23",
+    "@jscad/io-utils": "workspace:2.0.22",
     "@jscad/modeling": "workspace:2.10.0"
   },
   "devDependencies": {
--- a/packages/core/src/code-evaluation/rebuildGeometryCli.js
+++ b/packages/core/src/code-evaluation/rebuildGeometryCli.js
@@ -7,7 +7,7 @@ import requireDesignFromModule from '../code-loading/requireDesignFromModule.js'
 import getAllParameterDefintionsAndValues from '../parameters/getParameterDefinitionsAndValues.js'
 import makeWebRequire from '../code-loading/webRequire.js'
 
-export const rebuildGeometryCli = (data) => {
+export const rebuildGeometryCli = async (data) => {
   const defaults = {
     apiMainPath: '@jscad/modeling'
   }
@@ -34,7 +34,10 @@ export const rebuildGeometryCli = (data) => {
   }
 
   // rootModule should contain exported main and getParameterDefinitions functions
-  const rootModule = requireDesignFromModule(mainPath, requireFn)
+  // const rootModule = requireDesignFromModule(mainPath, requireFn)
+  // FIXME HACK for designs with import / export
+  const rootModule = await import(mainPath)
+
   // the design (module tree) has been loaded at this stage
   // now we can get our usefull data (definitions and values/defaults)
   const parameters = getAllParameterDefintionsAndValues(rootModule, parameterValues)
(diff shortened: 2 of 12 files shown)


z3dev:
@platypii not so clean, but at least the tests pass. this will let us reorganize the core package into a better structure, reorg functionality, and test everything.

the challenge will be supporting designs with require/modules as well as designs with import/export.

packages/core/src/code-evaluation/rebuildGeometryCli.js
```diff
@@ -34,7 +34,10 @@ export const rebuildGeometryCli = (data) => {
   }
 
   // rootModule should contain exported main and getParameterDefinitions functions
-  const rootModule = requireDesignFromModule(mainPath, requireFn)
```
platypii, line 37: Looks like `requireDesignFromModule` is not used anywhere anymore

packages/cli/package.json
```diff
@@ -39,6 +39,7 @@
     "@jscad/array-utils": "workspace:2.1.4",
     "@jscad/core": "workspace:2.6.4",
     "@jscad/io": "workspace:2.4.3",
+    "@jscad/io-utils": "^2.0.23",
```
platypii, line 42: should this have a `workspace:` prefix like the others?

platypii (approved):
Changes look like a good step toward V3! I think there will need to be additional refactoring, but it will be great to have the tests passing.

I did find some bugs while testing. Loading designs that use `require` does work. Using designs with ES module `import` does not always work. At first I was getting the following error:

```
$ node cli.js /tmp/cube-es6.js -o /tmp/cube.stl
JSCAD: generating output 
 from:  /tmp/cube-es6.js 
 to:  /tmp/cube.stl (STereoLithography, ASCII) 

(node:3935069) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/tmp/cube-es6.js:1
import * as jscad from '@jscad/modeling'
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

To fix it, I had to first add a `package.json` file in the same folder as the model. But then I still had problems importing `@jscad/modeling`. When I copied my model file into the `packages/cli` directory and ran from there, it finally worked.

```
$ node cli.js cube-es6.js -o /tmp/cube.stl
JSCAD: generating output 
 from:  cube-es6.js 
 to:  /tmp/cube.stl (STereoLithography, ASCII)
```

hrgdavor:
@z3dev I have a prototype for the web that runs all 3 variants
 - require (cjs)
 - import
 - typescript

in the worker a localized require is generated for each script that loads dependencies using synchronous IO. And for situations when a folder is drag and dropped, I am not reading all the files (folder could be huge in case of node projects) but I use serviceworkers to intercept http requests from those require calls and read files on demand. This also serves as a tracking which files are used and need to be watched for changes. So watching does not consume more resources based on projects size, but on amount of related imports/requires.

it is not yet in master branch there but: https://github.com/hrgdavor/jscadui/tree/threejs-with-cube-ctrl

@platypii was able to run it, but we both had samll issue with servise worker, that is misbehaved in normal chrome, but worked in incognito

to run the proto, checkout the code
in root run `npm install` then go into `cd apps/engine-test` and then just `npm run dev`

I am looking forward to ironing out any issues with different project types, so soon I will be testing all jscad examples.

One thing I did not implement yet is fallback to `.js` when ppl use `require('./localFile')` 

Maybe it is possible to do this without babel in CLI, but it is not possible to support import in worker or typescript. Another option is to use a faster transformer like esbuild or sth else

z3dev:
> To fix it, I had to first add a `package.json` file in the same folder as the model. But then I still had problems importing `@jscad/modeling`. When I copied my model file into the `packages/cli` directory and ran from there, it finally worked.

Yeah. NODE does not look outside of the project (cli) directory, so the design/project has to be re-loacted under cli. This is mentioned in the README.

> Note: The CLI will search for the design entry point just like NPM.
> - if there is a package.json file in the project, then try to load the 'main' property
> - if not, then try to load from 'index.js'

packages/core/src/code-evaluation/rebuildGeometryCli.js
z3dev: Yeah... more re-organization is required.

platypii (approved):

Issue #1204, opened 2023-02-06
V3: optimize extrudeLinearGeom2


z3dev:
## Expected Behavior

same... code needs some optimizations...
- correct function names / exports to match file name
- optimize to use options as provide as this is a private internal function

There many be other private / internal functions that need changes as well.

PR #1206, opened 2023-02-06
refactor(modeling): filenames match exports

platypii:
This PR renames files and exported functions to match each other. I also renamed martinez files to use camelCase so that it matches the standard linter.

This addresses half of #1204 by making the function names match their filenames. (but doesn't not touch the input validation, so this PR does NOT close that issue)

To find files where the default export did not match the filename, I used this carefully crafted grep command:
```bash
grep -rn --include='*.js' 'export default' packages/*/src | grep -Eiv '.?*/([a-z0-9]+)\.js.*\1'
```

### All Submissions:


packages/modeling/src/operations/booleans/martinez/compareEvents.js
```diff
@@ -4,14 +4,14 @@
...
 
 /**
- * @param  {SweepEvent} e1
- * @param  {SweepEvent} e2
+ * @param {SweepEvent} e1
+ * @param {SweepEvent} e2
  * @return {Number}
  */
-export default function compareEvents (e1, e2) {
+const compareEvents = (e1, e2) => {
```
z3dev, line 14: export const ... ?

i think that exporting the named function is also important unless the function will only be used internally.

z3dev (commented):
@platypii thanks again. i just have one question about exports otherwise these changes are nice.

actually, i submitted a set of changes to 'style' the martinez code. of course, the authors of martinez just complained, ranted, and then did nothing. SIGH

packages/modeling/src/operations/booleans/martinez/compareEvents.js
platypii: Good question. I don't think this is about internal vs external. This comes down to an opinion on JSCAD style. An ES module can export individual members, or a single default export. They can also be mixed. So there are a couple cases to consider:

 - a module with multiple exports. This is actually pretty rare in JSCAD. Example: `mirror.js` individually exports `mirror`, `mirrorX`, `mirrorY`, and `mirrorZ`.
 - a module which exports an object. Example: `constants.js` with `export { EPS, NEPS, TAU }`
 - a module that exports exactly one function or const. Example: 99% of files in jscad.

On the last, most common case, there is exactly one export. Now there is a choice:

 - export individually as `export const cube = ...`
 - export as a default import `const cube = ...` and `export default cube`
 - export default on one line `export default const cube = ...`
 - export both individually and as a default export

There is impact on how you import things. Depending how things are exported, you can import as either of:

```
import cube from './cube.js'
import { cube } from './cube.js'
import { EPS, TAU } from './constants.js'
```

It seems like the pattern you choose was to export both. This allows importing in either style, and effectively results in an export object that looks like:

```
{ default: cube, cube: cube }
```

There is nothing wrong with this, and it does give the most flexibility when importing. One could argue it's better to pick one way any be consistent. It seems like the logical option would be to always use named exports, not the default export. But I don't really feel strongly one way or the other. Here's a blog talking about it: [Module Exports vs. Export Default: Why Not Both?](https://bignerdranch.com/blog/default-exports-or-named-exports-why-not-both/)

For this PR, I made the change to be consistent with the rest of JSCAD v3, by including `export const compareEvents`.

z3dev (approved):
@platypii thanks

i'm still not certain why javascript added 'default' export. SIGH

packages/modeling/src/operations/booleans/martinez/compareEvents.js
hrgdavor: I personally also prefer named exports. I probably could find reasons to justify it :)

PR #1208, opened 2023-02-12
refactor(all): use named import/export for ES modules

platypii:
This is what it would look like if we got rid of all the `export default` in the V3 branch, and instead use named import/exports for everything.

I think it's a good best practice. Instead of there being two ways to do something, this makes it so that we are consistent in always doing it one way. No need to think about how to import or export. This removes the need to add "export default" at the bottom of each file, so it saves hundreds of lines of code. It feels like cleaner use of ES modules.

Everything is exactly the same for users. The exported API doesn't change at all. This just simplifies how modules are imported internally in the codebase.

Let me know what you guys think.

### All Submissions:


z3dev:
@garyhodgson what do you think?

personally, I like simple concise code. Which makes maintenance far easier.

@platypii is there a way to configure JSLint to flag default exports?

z3dev (dismissed):
@platypii thanks for the massive amount of cleanup. the indexes look really clean now as well.

in general, i think these changes improve the code by simplifying the exports, adjusting code to follow the same style, and eliminating a whole lot of redundant wasteful code.

Any last comments before I merge?

platypii:
Just pushed some last minute cleanup of the imports. I think it's good now.

There is an ESLint rule `no-default-export` that we could set, but eslint appears to not be setup currently?

z3dev:
> Just pushed some last minute cleanup of the imports. I think it's good now.
> 
> There is an ESLint rule `no-default-export` that we could set, but eslint appears to not be setup currently?

The basic rules are from sandardx but additional rules can be set. See package.json

platypii:
Added lint check. Also made some more updates: updated import/port for packages outside of modeling, and removed default exports from typescript .d.ts files.

z3dev:
reviewing... but there's a lot of file... ;)

z3dev (approved):
@platypii THAT WAS MASSIVE!!!

z3dev:
@platypii anything else pending? if not then i will merge.

platypii:
Yup looks good!

PR #1211, opened 2023-02-27
refactor(all): lint V3 branch

platypii:
This is an attempt at linting the V3 branch. Sorry for the mega PRs lately. Changes include:

 - `camelCase` variables instead of `alllowercase`
 - `standardx --fix`
 - Remove unused code
 - Spellcheck

### All Submissions:


z3dev:
@platypii how did you locate unused code?

platypii:
> @platypii how did you locate unused code?

A combination JetBrains IDE and running lint:

```
npm run lint | grep never
```

platypii:
> how did you locate unused code?

A combination JetBrains IDE and running lint:

```
npm run lint | grep never
```

packages/utils/regl-renderer/src/rendering/render.js
```diff
@@ -56,8 +56,8 @@ export const prepareRender = (params) => {
                 drawCmd = drawCache2.get(visuals.cacheId)
               } else {
                 visuals.cacheId = drawCache2.size
-console.log(visuals.drawCmd)
-console.log(context)
+                console.log(visuals.drawCmd)
```
z3dev, line 59: let's remove the debug

packages/core/src/code-evaluation/rebuildGeometryCli.js
```diff
@@ -1,34 +1,19 @@
...
-  // we need to update the source for our module
-  let requireFn = createRequire(import.meta.url)
+  let { mainPath, parameterValues, useFakeFs } = Object.assign({}, defaults, data)
 
   // source came from conversion, i.e. not from file system
   if (useFakeFs) {
     const pathParts = path.parse(mainPath)
-    const fakeName = `${pathParts.name}.js`
     const fakePath = `/${pathParts.name}.js`
-    const filesAndFolders = [
```
z3dev, line 23: hmm... i think this is used when converting from STL to JS. maybe we don't have a test case?

z3dev (changes requested):
@platypii thanks again. this is going to make the code stinky clean. ;)

i saw some debugging... let's remove those. maybe we should scan code again for console logging.

z3dev:
@platypii can you finish this off? i have a few changes queued up.

platypii:
Yup sorry was slammed with work this week. Looking at it now.

packages/core/src/code-evaluation/rebuildGeometryCli.js
platypii: It was used by `makeWebRequire` but that got replaced with a call to `await import(mainPath)`, so `filesAndFolders` became unused.

I reverted this change though, since I think we will need to revisit the code loading.

packages/utils/regl-renderer/src/rendering/render.js
platypii: I removed these console.log statements, as well as a bunch of other ones. Many more remain, for another day.

z3dev (approved):
@platypii thanks a lot. The code is now very clean.

PR #1216, opened 2023-03-06
Added length and arcLengthToT functions to bezier curves

achirita:
Added 3 new methods for bezier curves:
- `lengths` - **_private utility_** function used by the next 2
- `length` - approximates the length of the bezier curve by sampling it at a sequence of points, then adding up all the distances
- `arcLengthToT` - converts a given arc length to a t value; useful for generating equally spaced points along a bezier curve

### All Submissions:


hrgdavor (approved):
looks good to me. This will definitely be useful.

packages/modeling/src/curves/bezier/length.js
```diff
@@ -0,0 +1,20 @@
...
+ * @example
+ * const b = bezier.create([[0, 0], [0, 10]]);
+ * console.log(length(b)) // output 10
+ * 
+ * @param {Object} bezier a bezier curve.
+ * @param {Number} [segments=100] the number of segments to use when approximating the curve length.
+ * @returns an approximation of the curve's length.
+ * @alias module:modeling/curves/bezier.length
+ */
+const length = (bezier, segments = 100) => {
```
z3dev, line 16: wondering if the parameters should be reversed as math/geometry functions always place the object at the end.

the only down side is that the segments would then be required. however, segments is really the driver for the approximate length, and important.

@platypii thoughts?

platypii: Another option would be to make segments an option inside a an options object the way we do elsewhere. So you could do:
```js
bezier.length({}, b)
bezier.length({ segments: 200 }, b)
```

That would make it future proof in case we want to add some other options later.

achirita: I thought of this as well, but this `bezier.length({}, b)` just hurts my eyes :).
If you think this is the better option I can make the changes.

platypii: None of the solutions seem great here. I will defer to @z3dev

z3dev:
```
bezier.length(200, b)
bezier.length({ segments: 200 }, b)
```
For length(), the first option is best. There are several geometry functions with the same signature.

```
bezier.arcLengthToT(0.7, b, 100)
bezier.arcLengthToT({arcLength: 0.7, segments: 100}, b)
```
For arcLength(), let's go with the options signature to align with library standards. And of course, some defaults can be supplied for the parameters; arcLength and segments.

However... now there are two 'arcLength' in the call signature. So, i'm wondering if a different name for the function or a different name for the parameter would be in order.

achirita: If you have any naming suggestions I'm happy to make the changes

z3dev: How about...

`bezier.arcLengthToT({distance: 0.7, segments: 100}, b) // preferred`

if understood correctly, distance is between 0 and L. Is that correct?

z3dev:
@achirita there are just a few changes pending. if you have any questions then let us know.

achirita:
@z3dev I've updated the method signatures.

packages/modeling/src/curves/bezier/arcLengthToT.d.ts
```diff
@@ -0,0 +1,5 @@
+import Bezier from './type'
+
+export default arcLengthToT
+
+declare function arcLengthToT(arcLength: number, bezier: Bezier, segments?: number): number
```
z3dev, line 5: please correct

packages/modeling/src/curves/bezier/length.d.ts
```diff
@@ -0,0 +1,5 @@
+import Bezier from './type'
+
+export default length
+
+declare function length(bezier: Bezier, segments?: number): number
```
z3dev, line 5: please correct

packages/modeling/src/curves/bezier/lengths.d.ts
```diff
@@ -0,0 +1,5 @@
+import Bezier from './type'
+
+export default lengths
+
+declare function lengths(bezier: Bezier, segments?: number): Array<number>
```
z3dev, line 5: please correct after moving segments to start of parameters

z3dev (changes requested):
@achirita please correct the TS definitions as well.

also, there's a pending changes to lengths() as well.

thanks again

z3dev (approved):
@achirita super. this looks ready, done, dusted.

PR #1218, opened 2023-03-15
export retessellate

FishOrBear:


platypii (approved):
Make sense to me. Why not export this. Thanks @FishOrBear for the contribution!

Side note, this PR got it right, but there are places we misspell `retesselate` with one `l` in the codebase.

z3dev:
Considering that no one knows what retesselate means... and spell checkers fail to recognize retesselate... there should be a few misspelling of retesselate :)

PR #1219, opened 2023-03-19
reorg(io): dropping gcode deserializer

z3dev:
The GCODE deserializer has never been touched in V1 or V2. This code is now unmaintainable and unsupported. Time to remove this relic.

### All Submissions:


z3dev:
P.S. I won't remove AMF or SCAD yet as someone might port these to V3.

platypii (approved):

hrgdavor (approved):
Yeah, it is a nice idea, but without a dedicated developer that would actually maintain it ....

Probably there is something like this standalone ... short google search found this https://github.com/aligator/gcode-viewer ... anyway, I do not see internal use-case for dedicated g-code viewer.

Finally, if ever wanted to be resurrected, can be done by taking code from V1/V2 codebase.

PR #1226, opened 2023-04-08
fix(modeling): fix line2.closestPoint

### LINKED ISSUE #1225: line2.closestPoint output is wrong
## Expected Behavior
line2.closestPoint(..) should return a point on the line which is closest to the provided point.

## Actual Behavior
In some cases line2.closestPoint(..) returns a point which is not on the line (see screenshots below).
If we consider a line going from [10, 0] to [0, 10] and try to get the point on the line which is closest to [0, 0] we get [-0.5952027266534774, 0.7912811580260264]. The correct result would be [5, 5].

If we switch to the line3.closestPoint(..) function and add 0 as the Z value for all the points we will get the correct result.

Get the closest point and draw a circle around it. The result is not correct
```javascript
const { line, circle } = jscad.primitives
const { line2 } = jscad.maths;

const main = () => {
  const start = [10, 0];
  const end = [0, 10];
  const point = [0, 0];
  
  const l = line2.fromPoints(line2.create(), start, end);
  const closest = line2.closestPoint(l, point);
  
  return [line([start, end]), circle({radius: 0.2, center: closest})];
}

module.exports = { main }
```
wrongPoint

Get the closest point using line3. This one returns the correct result.
```javascript
const { line, circle } = jscad.primitives
const { line3 } = jscad.maths;

const main = () => {
  const start = [10, 0, 0];
  const end = [0, 10, 0];
  const point = [0, 0, 0];
  
  const l = line3.fromPoints(line3.create(), start, end);
  const closest = line3.closestPoint(l, point);
  console.log(closest);
  
  return [line([start, end]), circle({radius: 0.2, center: closest})];
}

module.exports = { main }
```
correctPoint

## Specifications

  - Version: modeling 2.11.0
  - Platform: all
  - Environment: all

platypii:
Fixes #1225

Thanks @achirita for catching that, and for suggesting the fix! I used your suggested fix, but refactored it a bit to avoid allocating intermediate vec2's.

I also added tests for `origin` and `direction` and `closestPoint` which cover this case.

Screenshot at 2023-04-08 12-25-31

### All Submissions:


hrgdavor (approved):
looks good !

PR #1239, opened 2023-04-29
x3d-serializer: specular highlights and shape names

andreasplesch:
### All Submissions:

Rather subtle improvement to rendering by adding some specular highlighting similar to the default renderer. See #1234.

### DIFF
--- a/packages/io/x3d-serializer/src/index.js
+++ b/packages/io/x3d-serializer/src/index.js
@@ -48,6 +48,7 @@ const mimeType = 'model/x3d+xml'
  * Serialize the give objects to X3D elements (XML).
  * @param {Object} options - options for serialization, REQUIRED
  * @param {Array} [options.color=[0,0,1,1]] - default color for objects
+ * @param {Number} [options.shininess=8/256] - x3d shininess for specular highlights
  * @param {Boolean} [options.smooth=false] - use averaged vertex normals
  * @param {Number} [options.decimals=1000] - multiplier before rounding to limit precision
  * @param {Boolean} [options.metadata=true] - add metadata to 3MF contents, such at CreationDate
@@ -63,6 +64,7 @@ const mimeType = 'model/x3d+xml'
 const serialize = (options, ...objects) => {
   const defaults = {
     color: [0, 0, 1, 1.0], // default colorRGBA specification
+    shininess: 8 / 256,
     smooth: false,
     decimals: 1000,
     metadata: true,
@@ -180,7 +182,13 @@ const convertAppearance = (object, colorField, options) => {
   const colorRGB = object.color.slice(0, 3)
   const color = colorRGB.join(' ')
   const transparency = roundToDecimals(1.0 - object.color[3], options)
-  return ['Appearance', ['Material', { [colorField]: color, transparency }]]
+  const materialFields = { [colorField]: color, transparency }
+  if (colorField === 'diffuseColor') {
+    Object.assign(
+      materialFields,
+      { specularColor: '0.2 0.2 0.2', shininess: options.shininess })
+  }
+  return ['Appearance', ['Material', materialFields]]
 }
 
 /*
--- a/packages/io/x3d-serializer/src/index.js
+++ b/packages/io/x3d-serializer/src/index.js
@@ -43,6 +43,7 @@ const stringify = require('onml/lib/stringify')
 // https://x3dgraphics.com/examples/X3dForWebAuthors/Chapter13GeometryTrianglesQuadrilaterals/
 
 const mimeType = 'model/x3d+xml'
+const defNames = new Map()
 
 /**
  * Serialize the give objects to X3D elements (XML).
@@ -143,7 +144,7 @@ const convertObjects = (objects, options) => {
 const convertPath2 = (object, options) => {
   const points = path2.toPoints(object).slice()
   if (points.length > 1 && object.isClosed) points.push(points[0])
-  const shape = ['Shape', {}, convertPolyline2D(poly2.create(points), options)]
+  const shape = ['Shape', shapeAttributes(object), convertPolyline2D(poly2.create(points), options)]
   if (object.color) {
     shape.push(convertAppearance(object, 'emissiveColor', options))
   }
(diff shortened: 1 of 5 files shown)


andreasplesch:
[Without highlights (current)](https://create3000.github.io/x_ite/playground/?url=https://gist.githubusercontent.com/andreasplesch/dc3595c5cf5198bedd7dbdb119a7a059/raw/ffb746984bcd31340ac9d2b2a88650c7ff37ab57/jscad.x3d)
[With highlights, this PR](https://create3000.github.io/x_ite/playground/?url=https://gist.githubusercontent.com/andreasplesch/dc3595c5cf5198bedd7dbdb119a7a059/raw/ffb746984bcd31340ac9d2b2a88650c7ff37ab57/basicColorsSpecular.x3d)

z3dev:
@andreasplesch this seems reasonable but i think colors and other material attributes should be provided at the shape level.

several of the serializers already support other attributes on shapes, for example 'name', 'id', etc.

if the other X3D material attributes are important then how about adding support for those?

https://doc.x3dom.org/developer/x3dom/nodeTypes/Material.html

andreasplesch:
Yes, the Material fields correspond to the shape level. Only diffuseColor is important for 3d printing, I believe. The other fields are important for screen rendering. The initial idea was to get closer to the rendering of the regl renderer by default which also uses specular highlights. Further refinement could be done in postprocessing outside of jscad. For that it would be useful to support a 'name' attribute which would correspond to the X3D DEF attribute.

It would be possible to look for object properties with the Material field names such as object.ambientIntensity and then use those for the x3d serialization. This way other serializers or the regl renderer could also use these object properties.

Should these properties be grouped in a material (or x3d) subobject: object.material.ambientIntensity ?
Should these properties be managed or propagated by jscad in operations which create new objects ? This seems out of scope at this point.

There are some decisions which then would follow:

- Use x3d defaults values, except for specularColor and shininess.
- geom3 uses color as diffuseColor. If both are given use diffuseColor.
- line geom uses color as emissiveColor. If both are given use emissiveColor.
- If line geom has properties other than color, include all in the material node. This is legal although x3d will only use emissive.
- Allow options for all fields to be used if object property does not exist.

So, it may suffice to support only a 'name' attribute and expect postprocessing outside of jscad. Adding support for all material fields as object attributes is also possible, potentially allowing use of those elsewhere although they are somewhat x3d specific.

I am fine with both options.

andreasplesch:
I added support for a .name attribute. There is no default option. Because X3D requires unique DEF names, I added a function which adds suffixes for Shapes which have identical .names. This could be left out and become a responsibility for the designer.

With regards to Material attributes, this may not be a priority but please let me know and I can add support.

packages/io/x3d-serializer/src/index.js
```diff
@@ -165,6 +168,24 @@ const convertGeom2 = (object, options) => {
   return group
 }
 
+/*
+ * generate attributes for Shape node
+ */
+
+const shapeAttributes = (object, attributes = {}) => {
+  if (object.name) {
+    Object.assign(attributes, { DEF: uniqueDefName(object.name) })
```
z3dev, line 177: The 'name' should be decided by the user. So, let's just use whatever the user provides.

packages/io/x3d-serializer/src/index.js
```diff
@@ -180,14 +201,20 @@ const convertAppearance = (object, colorField, options) => {
   const colorRGB = object.color.slice(0, 3)
   const color = colorRGB.join(' ')
   const transparency = roundToDecimals(1.0 - object.color[3], options)
-  return ['Appearance', ['Material', { [colorField]: color, transparency }]]
+  const materialFields = { [colorField]: color, transparency }
+  if (colorField === 'diffuseColor') {
+    Object.assign(
+      materialFields,
+      { specularColor: '0.2 0.2 0.2', shininess: options.shininess })
```
z3dev, line 208: Can something similar be done here for material attributes?

packages/io/x3d-serializer/src/index.js
andreasplesch: ok, I can remove the unique function. Would it be possible to provide a warning, perhaps to the console, in case the same name is used for multiple objects ?

packages/io/x3d-serializer/src/index.js
andreasplesch: Ok, we can look for all Material attributes as outlined above.

packages/io/x3d-serializer/src/index.js
z3dev: that sounds reasonable. by definition, are DEF (names) expected to be unique?

packages/io/x3d-serializer/src/index.js
z3dev: Cool.

By the way, what ever becomes supported in the serializer should be supported in the deserializer. We should support the round trip...

andreasplesch:
Yes, DEF names in x3d are required to be unique. x3d processors may
complain.

---on the phone---

packages/io/x3d-serializer/src/index.js
andreasplesch: I noticed that the deserializer uses obect.id for DEF names. Let me then use .id here as well
 rather than .name

andreasplesch: Would material attributes in a sub property be appropriate: object.material.ambientIntensity ?

z3dev: That seems reasonable. Please add some notes to the README file.

z3dev (commented):
@andreasplesch these changes look fine.

just two things...
- please run 'npm run lint' from the main directory and see if any issues are highlighted
- please set the status of your branch to NOT draft

andreasplesch:
Thanks for taking a look. I fixed the lint reports and requested review to convert from draft to PR. If  this is ready to be merged, I can address remaining items later in another PR.

z3dev (approved):
@andreasplesch YES! Thanks much for the cleanup.

All looks good!
Thanks for the changes. Looking forward to the next set of changes.

PR #1250, opened 2023-05-18
Performance improvements for retessellate

achirita:
### All Submissions:

Based on this discussion https://github.com/jscad/OpenJSCAD.org/discussions/1247 I have refactored `retessellate.js` to improve performance.
This function is used by all boolean operations and is the main reason why sometimes they're really slow.
To measure the performance I used a geometry with ~15k polygons which gets reduced to ~7.5k polygons.
Results for current implementation of `retessellate.js` :
- total execution time ~580ms of which:
  -  ~465ms used to determine coplanar polygons (lines 29-39) 
  - ~115ms used to call `reTesselateCoplanarPolygons`

Results for new implementation:
- total execution time ~120ms of which 
  - ~18ms used to determine coplanar polygons (lines 21-22) 
  - ~102ms used to call `reTesselateCoplanarPolygons`

There is an ~5x overall performance improvement, but if we exclude the calls to `reTesselateCoplanarPolygons` there's more like an ~25x improvement.

I haven't added this test to the existing suite because I didn't want it slowing down build times. If you consider it should be added, I can work something out.

### DIFF
--- a/packages/modeling/src/operations/modifiers/retessellate.js
+++ b/packages/modeling/src/operations/modifiers/retessellate.js
@@ -1,17 +1,9 @@
 const geom3 = require('../../geometries/geom3')
 const poly3 = require('../../geometries/poly3')
 
-const aboutEqualNormals = require('../../maths/utils/aboutEqualNormals')
-
 const reTesselateCoplanarPolygons = require('./reTesselateCoplanarPolygons')
 
-const coplanar = (plane1, plane2) => {
-  // expect the same distance from the origin, within tolerance
-  if (Math.abs(plane1[3] - plane2[3]) < 0.00000015) {
-    return aboutEqualNormals(plane1, plane2)
-  }
-  return false
-}
+const tolerance = 0.000000015
 
 /*
   After boolean operations all coplanar polygon fragments are joined by a retesselating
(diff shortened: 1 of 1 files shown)


packages/modeling/src/operations/modifiers/retessellate.js
```diff
@@ -26,29 +18,77 @@ const retessellate = (geometry) => {
...
+  // restore the original order of the polygons
+  const result = []
+  // polygons inside the cluster should already be sorted by index
+  clusters.forEach(cluster => result[cluster[0]?.index] = cluster)
+  nonCoplanar.forEach(polygon => result[polygon.index] = polygon)
+
+  return result
+}
+
+const byPlaneComponent = (component) => {
```
z3dev, line 83: @achirita i'd like to split the comparisons of the distance component from the comparisons of the normal components (X Y Z). take a look at src/maths/utils/aboutEqualNormals.js

basically, the tolerance should be different.

packages/modeling/src/operations/modifiers/retessellate.js
```diff
@@ -26,29 +18,77 @@ const retessellate = (geometry) => {
...
+        maybeCoplanar.push(cluster.slice(startIndex))
+      }
+    })
+    // replace previous clusters with the new ones
+    clusters = maybeCoplanar
+  }
+  // restore the original order of the polygons
+  const result = []
+  // polygons inside the cluster should already be sorted by index
+  clusters.forEach(cluster => result[cluster[0]?.index] = cluster)
```
z3dev, line 77: This is too cryptic. Can you expand the expression?

I think that the full function signature should also be provided. Same for the next line.

z3dev (dismissed):
@achirita pretty cool. just a few comments.

packages/modeling/src/operations/modifiers/retessellate.js
achirita: I know it's a bit confusing, I'll try to explain the best way I can.
Before calling the `classifyPolygons()` function, on line 19, all polygons are remapped to contain vertices, the plane object and the original index.
The original index is important because during classification the polygons get sorted based on the components of the plane object so they're out of order from the perspective of their index.
For geom3 objects this is not really important because the order of the polygons doesn't really matter.
geom2 is where things start to get wrong. Boolean operations for geom2 objects have a "special" behavior - they transform geom2 to fake geom3, do the boolean operation and then turn the result back to a geom2 object. In this scenario, if the polygons are out of order then the resulting points of the geom2 object will be out of order.
Now that I've explained why we need to sort the result based on index, let's take a look at those lines of code you highlighted.
After classification we have two resulting arrays (`nonCoplanar` and `clusters`) which need to be put back together in a single array (`result`), ordered by index. 
Let's start with `nonCoplanar` as it's the easiest to explain. This is an array of polygons which are not coplanar with any others.
Line 77 iterates through each polygon and places it in the `result` array at it's original index.
For `clusters` there's a similar process with some small differences. `clusters` is an array of arrays of polygons. We iterate though each cluster and place it in the `result` array at the index of it's first polygon.
To make it even easier to understand, let's take a look at a small example. Let's assume we have a geometry with 10 polygons, some of which are coplanar. These would be the indexes of the polygons before classification `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`. At the end of the classification process we end up with something like this: `nonCoplanar`: `[9, 4]` and `clusters`: `[ [2, 7], [0, 5], [1, 6], [3, 8] ]`.
When putting things back to the `result` array we would have: 
```
result[9] = {index: 9, vertices: [...], plane: [...]}
result[4] = {index: 4, vertices: [...], plane: [...]}
result[2] = [{index: 2, vertices: [...], plane: [...]}, {index: 7, vertices: [...], plane: [...]}]
result[0] = [{index: 0, vertices: [...], plane: [...]}, {index: 5, vertices: [...], plane: [...]}]
result[1] = [{index: 1, vertices: [...], plane: [...]}, {index: 6, vertices: [...], plane: [...]}]
result[3] = [{index: 3, vertices: [...], plane: [...]}, {index: 8, vertices: [...], plane: [...]}]
```
This is exactly the same behavior as the previous implementation.
Sorry for the wall of text, hope it helps.

In terms of providing full function signature - are you suggesting I should extract these 2 lines in separate methods?

z3dev: @achirita got it. I understand the sorting / combing into grouped.

it not about that... the planes are normals plus distance from origin. Grouping distances is not the same as grouping normals . The tolerances are different.

achirita: I just updated that based on your comments

packages/modeling/src/operations/modifiers/retessellate.js
```diff
@@ -26,29 +16,78 @@ const retessellate = (geometry) => {
...
 }
 
-module.exports = retessellate
+const classifyPolygons = (polygons) => {
+  let clusters = [polygons] // a cluster is an array of potentially coplanar polygons
+  const nonCoplanar = [] // polygons that are known to be non-coplanar
+  // go through each component of the plane starting with the last one (the distance from origin)
+  for (let component = 3; component >= 0; component--) {
+    const maybeCoplanar = []
+    const tolerance = component == 3 ? 0.000000015 : NEPS
```
platypii, line 44: Why is the tolerance different for component 3?

platypii (approved):
I tested against a number of different examples. I found this branch to be slightly faster in some cases, and about the same performance in most cases. I did not find any cases that were substantially worse performance.

It helped the most with union of spheres 260ms -> 45ms
tess-sph

It helped a little with inline cylinders 145ms -> 80ms
tess-inline

Minor improvement for union of intersecting cylinders 1190ms -> 1050ms
tess-union

And roughly the same for adjacent intersecting cylinders 235ms -> 240ms
tess-adj

Code structure is cleaner. Overall nice change, thanks @achirita!

packages/modeling/src/operations/modifiers/retessellate.js
achirita: Thanks for the additional testing, those results look encouraging. 
In the old implementation there was this function which checked if two polygons are coplanar and it used `0.00000015` as tolerance.
```javascript
const coplanar = (plane1, plane2) => {
  // expect the same distance from the origin, within tolerance
  if (Math.abs(plane1[3] - plane2[3]) < 0.00000015) {
    return aboutEqualNormals(plane1, plane2)
  }
  return false
}
```
The other components were checked by calling `aboutEqualNormals(..)` which internally uses `NEPS` as tolerance.

z3dev:
@platypii wouldn't this be more appropriate in V3?

platypii:
> wouldn't this be more appropriate in V3?

Based on the tests being unchanged, it seems like @achirita implemented this in a way which does not change the output result at all. So it is not a breaking API change.

I think anything without breaking changes should go in V2 still. Why? Because V3 is not ready yet, who knows when it will be ready. But users are actively using V2, so improvements are nice, and usage also let's us find and fix bugs sooner.

z3dev:
@platypii can you follow with lint changes for V2 modeling?

PR #1252, opened 2023-05-27
feat(modeling): reworked vectorChar and vectorText to return segments as path2 objects

z3dev:
These change rework both vectorChar and vectorText to convert font segments into path2 objects. This allows the font outlines to be used directly in other functions.

Actually, the results are not lists of path2 objects, but rather lists of vectorChar and vectorText objects (anonymous), and the vectorChar object has a list of path2 objects.

Some subtle bugs were corrected as well, which included the use of letterSpacing and lineSpacing.

### All Submissions:


z3dev:
@platypii @hrgdavor please look over these changes.

the vectorChar and vectorText functions return anonymous objects still, as there's some useful information provided about height and width. if not useful then just list of path2 objects can be produced.

hrgdavor:
> useful information provided about height and width.
I am for keeping that info. I am guessing it is needed to place them correctly ?

hrgdavor (approved):
looks good :)

z3dev:
@rozek You may be interested in these changes. If you have time then please take a look. All comments are welcome.

PR #1255, opened 2023-06-11
refactor(modeling): reworked slice to use vertices in parameters, etc.

z3dev:
Another set of changes to refactor the geometries to use points (2D) and vertices (3D). This is taking on the easy one; slice.

### All Submissions:


hrgdavor (approved):
looks good :)

packages/modeling/src/geometries/slice/calculatePlane.test.js
```diff
@@ -12,7 +12,7 @@ test('slice: calculatePlane() returns correct plans for various slices', (t) =>
   // const slice1 = create()
   // const plane1 = calculatePlane(slice1)
 
-  const slice2 = fromPoints([[0, 0], [1, 0], [1, 1]])
+  const slice2 = fromVertices([[0, 0], [1, 0], [1, 1]])
```
platypii, line 15: Why does this say vertices, but the points are 2D?

I guess the answer is that `slice.fromPoints` can take either `vec2` or `vec3` as inputs. So we don't _need_ to rename it technically. But it is meant to be a 3D slice, so I'm good with renaming to `fromVertices` to make it more clear to users of the api.

platypii (approved):

packages/modeling/src/geometries/slice/calculatePlane.test.js
z3dev: I came to the same conclusion. fromVertices() is probably the better name, as mixing "points' and 'vertices' in the API would be more confusing.

PR #1259, opened 2023-06-18
feat(regl-renderer): added rollup configuration to create ES6 bundle, adjusted imports and exports

z3dev:
These changes complete the bundling of regl-renderer using rollup. There were some adjustments to the imports and exports to get the package into a usable state. Also, a new demo (HTML) was added to demonstrate how to use V3 modeling and regl-renderer together.

A new section for V3 geometry to regl entities was created, as well as tests.

### All Submissions:


z3dev:
@platypii What do you think about removing support for V1 and V2 geometries, and just support V3 geometries? This goes the same for the demos.

hrgdavor (approved):
looks good.

I do not mind if v1 and v2 conversion is not kept here in V3. it is available in V1 and  V2 codebase.

z3dev:
@platypii @hrgdavor thanks for the reviews and assistance.

there have been more improvements to the regl-renderer, which seems 100% functional again.

i have no idea if this will be used in the new V3 web UI but someone may be interested in using package.

Can you do the review again? Thanks

platypii:
@z3dev how can I test this?

Opening `demo.html` and `demo.es.html` in the browser as a file doesn't work, it needs to be hosted.

I tried going into `packages/utils/regl-renderer` and doing `npm run demo-cli` and got an error about missing package `gl`.

`npm i gl` didn't work because `Unsupported URL Type "workspace:": workspace:2.1.4`.

I figured I would try switching to `pnpm`, which seems like it helped. I was able to run `pnpm i gl`. But `pnpm run demo-cli` gives me error:
```
node: symbol lookup error: /home/platypii/code/jscad/node_modules/.pnpm/gl@6.0.2/node_modules/gl/build/Release/webgl.node: undefined symbol: _Z15XextFindDisplayP15_XExtensionInfoP9_XDisplay
```

When I try to run `pnpm run dev`  I get error:
```
sh: 1: budo: not found
```

An update to the README would be helpful?

z3dev:
> @z3dev how can I test this?
> 
> Opening `demo.html` and `demo.es.html` in the browser as a file doesn't work, it needs to be hosted.

Correct. We are at the mercy of the browsers as CORS has a huge affect on 'import'. So, a small server is required. I was running the python built in server.

Do you think we should install the little server for NPM as part of 'dev'?

> I tried going into `packages/utils/regl-renderer` and doing `npm run demo-cli` and got an error about missing package `gl`.
> 
> `npm i gl` didn't work because `Unsupported URL Type "workspace:": workspace:2.1.4`.
> 
> I figured I would try switching to `pnpm`, which seems like it helped. I was able to run `pnpm i gl`. But `pnpm run demo-cli` gives me error:

pnpm is correct.

> node: symbol lookup error: /home/platypii/code/jscad/node_modules/.pnpm/gl@6.0.2/node_modules/gl/build/Release/webgl.node: undefined symbol: _Z15XextFindDisplayP15_XExtensionInfoP9_XDisplay

can't help here... looks like the install of GL didn't work. Maybe you need more compilers, etc.

> When I try to run `pnpm run dev` I get error:
> 
> sh: 1: budo: not found

budo is common js only. It kind of sucked anyway. I'll replace with something useful

> An update to the README would be helpful?

OK. Will try to find a few common solutions,and update the documents.

hrgdavor:
`esbuild` aside from building can also run in mode that serves html and built code.

platypii:
`npx http-server` works for serving local files pretty easily. When I do that, the page loads, but I get error loading `http://localhost:8080/jscad-modeling.es.js`

http-server

z3dev:
> `npx http-server` works for serving local files pretty easily. When I do that, the page loads, but I get error loading `http://localhost:8080/jscad-modeling.es.js`

This will be corrected once V3 packages are published, but for now a local copy of the modeling package is required.

hrgdavor (dismissed):
looks good :)

packages/utils/regl-renderer/package.json
```diff
@@ -27,7 +44,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "types": "types",
```
z3dev, line 30: Revert. There are TS defines in the types directory.

platypii (dismissed):
Confirmed working locally :+1:

mister-jscad (approved):
move on...

Issue #1262, opened 2023-06-24
exporting non-STL formats causes slicing issues


tsdexter:
Is there a way to download an STL with multiple parts? I'm using the mentioned model below as part of a multi part model with parts that fit exactly together (ie: the word is cutout of a box but also placed exactly back inside as a separate model for easy colouring in the slicer), when I download as an STL and import it's all one solid object with no word or separate parts. 

This is causing me to have to use 3MF as file format, which does separate parts well, however, it presents the issue below: 

## Expected Behavior

Exporting an STL or a 3MF should export a file that slices the same

## Actual Behavior

STL slices good, but 3MF or AMF have some weird features and cannot be printed. The geometry looks the same in all 3 files, ie: 
<img width="2819" alt="image" src="https://github.com/jscad/OpenJSCAD.org/assets/4126987/2690cd0c-f6c8-46e0-aa1a-cae71dfdc6d8">

However, once I sliced it the 3MF and AMF have some weird features that make them unprintable: 
<img width="2751" alt="image" src="https://github.com/jscad/OpenJSCAD.org/assets/4126987/b8359e69-02b6-414f-b7fc-5a875aa4a8da">

Notice: 
- the dot over the `i` is only roughly half as thick on 3mf/amf
- the top corner of the `i` has some weird indentations
- the `o` is only 1 layer thick, has no hole, and the single layer that is there exists floating at layer 7

## Steps to Reproduce the Problem

  1. Use the model https://gist.github.com/tsdexter/add64749db71add39908f6640fef20c3
  2. Export STL/AMF
  3. Import and slice

## Specifications

  - Version: 2.11.1
  - Platform: Mac
  - Environment: consistent in browser and node CLI downloading

tsdexter:
note: if I exported the name and the box with the name cutout separately as 2 STLs and then import them into slicer at the same time and select 'yes' to "do you want to load these files as 1 object with multiple parts" they import well and I can easily colour just the name, however, this means I have to export twice, once with each part commented out. Ideally there is someway to easily export it (via CLI) once and get multiple files.

z3dev:
@tsdexter You are correct. STL does not support the concept of parts, just one mesh. 

You can use DXF, 3MF, or X3D to exports parts. Of course, you design should return an array of parts from main(). Most people assign colors to each part in order to see the parts in slicers.

We can probably help out if you can provide the exported 3MF file.

tsdexter:
[...]

PR #1263, opened 2023-06-25
cli option to output files as individual part files or .zip of parts

tsdexter:
Adds the CLI export utilities a `-p` flag to output multiple parts as separate files as well as a `-z` flag to zip the multiple files into one `design.zip` exported file

related: https://github.com/jscad/OpenJSCAD.org/issues/1262 

### All Submissions:

* [ x] Have you followed the guidelines in our Contributing document?
* [ x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [ x] Does your submission pass tests?

### DIFF
--- a/packages/cli/cli.parameters.test.js
+++ b/packages/cli/cli.parameters.test.js
@@ -1,4 +1,5 @@
 const test = require('ava')
+const JSZip = require('jszip')
 
 const path = require('path')
 const { execSync } = require('child_process')
@@ -244,7 +245,7 @@ test('cli (single input file, multiple output files)', (t) => {
   t.true(fs.existsSync(outputPath3))
 })
 
-test('cli (single input file, zipped output file)', (t) => {
+test('cli (single multipart input file, zipped output file)', async (t) => {
   const testID = 8
 
   const inputPath = createJscad(testID, true)
@@ -252,7 +253,6 @@ test('cli (single input file, zipped output file)', (t) => {
 
   t.context.inputPath = inputPath
 
-  
   const outputName = `./test${testID}.zip`
   const outputPath = path.resolve(__dirname, outputName)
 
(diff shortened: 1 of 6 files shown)


tsdexter:
Example CLI output `jscad index.js -p -z`:
image

Example CLI output: `jscad index.js -p`:
image

Example Filesystem output from both commands, after extracting the .zip:
image

packages/cli/src/generateOutputData.js
```diff
@@ -22,7 +22,7 @@ const generateOutputData = (source, params, options) => {
     addMetaData: true
```
z3dev, line 22: Let's add a default value for 'parts'

It would be nice to get proper documentation for the parameters as well but that may be asking too much.

packages/cli/cli.parameters.test.js
```diff
@@ -207,3 +213,56 @@ test('cli (single input file, invalid jscad)', (t) => {
     execSync(cmd, { stdio: [0, 1, 2] })
   })
 })
+
+
+test('cli (single input file, multiple output files)', (t) => {
```
z3dev, line 218: Nice. Thanks much.

z3dev (commented):
@tsdexter Cool! Nice set of changes, and tests!

I just suggested one little change.

Also, I'm wondering if 'parts' parameter should be named something like 'generateParts'.

I'm also wondering if the zip option might be applicable to all designs. It's really a convenience features, so why not allow all output to be zipped?

tsdexter:
I will make adjustments sometime tomorrow. I agree `generateParts` would be less ambiguous. 

I also thought about having zip flag for single part exports as well, but figured it wasn't necessary as they will always be one file anyway. However, I think there is still the use case of having a zip file for sharing and other use cases where certain formats are disallowed. I will add it as an option for all export types.

packages/cli/src/generateOutputData.js
tsdexter: I must've missed this section. I added the defaults to `packages/cli/src/parseArgs.js`. Resolved in: 1898eaf7230e3df29567c46f3a55231ac8ca4e4e

tsdexter:
@z3dev I've added zip functionality to all exports (single or multi-file). I've created a test for single file .zip exports as well as refactoring the zip tests to actually check the contents of the zip file, instead of just its existence.

tsdexter:
@z3dev another feature that would be useful for me specifically is a `license` flag. I would envision its use like: 

`jscad index.js -l "path/to/license.txt"` 

If the flag is used and the file exists, it would automatically use the .zip functionality and export all of your models files alongside the license.txt in a .zip file. 

Should I add this to the CLI as a useful feature for everyone, or do you think this should be handled in my own workflow?

z3dev (approved):
@tsdexter looks good. thanks.

let's see if @platypii or @hrgdavor have any comments

hrgdavor (approved):
Looks good. I like the addition.

Also other mentioned features are interesting and could be another PR.

tsdexter:
@z3dev whats the process for merging/releasing?

platypii (approved):
Overall looks good. I tested locally variations of zipped, unzipped, and various numbers of parts (including 0, 1, many parts) all seemed fine.

My only minor comment would be I wish we could have more descriptive cli flags names like "--parts" and "--zip". But this is a more general comment. But this PR is consistent with the current cli use of short names.

Thanks this is a nice contribution @tsdexter! I've had to use awkward parameter tricks to generate multipart files in the past, and this would make some use cases easier.

Issue #1272, opened 2023-07-18
Idea: functions to generate the stroke of a path or geom2


platypii:
We should introduce functions to generate the `stroke` of a path.

![stroke](https://github.com/jscad/OpenJSCAD.org/assets/1766297/2d6ec06a-6f96-44ee-86ce-3b3f21a0ff5e)

Such functions could be useful for creating the exterior of a shape from a given path. There are cases where having only the path is not enough and one might need to add "thiccness". This could also be helpful for things like importing SVGs which can have both fill and stroke.

For paths, this functionality would essentially be `offset`. However, there is currently no straightforward way to compute the stroke of a `geom2`.

Options for stroking a path, from the [svg spec](https://www.w3.org/TR/SVG2/painting.html#LineCaps):
 - `width`
 - `cap`: butt | round | square
 - `join`: arcs | bevel |miter | miter-clip | round
 - `miterlimit`

z3dev:
Maybe a little more primitive, the WebAPI provides both CanvasRenderingContext2D and Path2D interfaces for constructing 2D paths for rendering. Maybe some of these concepts could be adopted.

PR #1273, opened 2023-07-19
feat(modeling): remove extrudeRectangular

platypii:
This might be controversial. Let's use this PR for discussion on removing `extrudeRectangular` from V3.

The `extrudeRectangular` function is supposed to:

> Extrude the given geometry by following the outline(s) with a rectangle.

But it's not doing that. If you "follow the outline(s) with a rectangle" you would get something more like the red shape, but `extrudeRectangular` returns the shape on the right:

extrude-rect

In reality `extrudeRectangular` is basically just doing:

```js
extrudeLinear({}, offset({}, geometry))
```

This is not easy to fix, so I suggest that we simply delete `extrudeRectangular`. I would be surprised if anyone is using it, but if they are, they can use `offset`+`extrude` instead.

I think what was really intended here was a `stroke` function like the suggestion in #1272 

### All Submissions:


z3dev:
@platypii interesting. i really didn't like this extrude function from day one as, as you mentioned, is just putting two functions together. users should be able to figure out how to do this through a simple example.

i also don't like 'torus' either, as that is simply extrude rotate with a sphere. it makes no sense to implement primitives via complex functionality. primitive should exist by themselves. examples should show how to create a torus.

platypii:
Rebased to fix merge conflicts.

Can anyone tell me what exactly is "rectangular" about `extrudeRectangular`?

z3dev:
> Rebased to fix merge conflicts.
> 
> Can anyone tell me what exactly is "rectangular" about `extrudeRectangular`?

Nope. You could go back and look at V1, which may have had the same issues, or different issues.

z3dev:
From CSG.js V1... math/Path2.js

```
// Expand the path to a CAG
// This traces the path with a circle with radius pathradius
expandToCAG: function (pathradius, resolution) {
}
```

platypii:
Earliest I could find was this commit from 2012: [joostn/OpenJsCad@b059ed95](https://github.com/joostn/OpenJsCad/commit/b059ed95#diff-77464d6f63d562f0ddd14d2cbdef2eea956b7f1e2fe26ef270b9b5bb88768619R3652-R3667)

So... can we kill it? :fire:

z3dev (approved):

PR #1279, opened 2023-08-13
fix(modeling): fix tsd tests

platypii:
This PR fixes the failing `tsd` test.

In order to make the tests pass with our new ES module style imports I had to make the following changes:

 - Update `tsd` npm dependency
 - A couple minor type issues that surfaced
 - In three files I had to add `@typedef` for `Mat4`. I'm not sure why this was needed, but it did fix the issue.
 - Added `V3` branch to the list of branches to run CI on

As part of updating the `tsd` dependency, it updated the lockfile version and caused a bunch of diff noise. I'm assuming it's because on my machine I used the latest node/npm/pnpm version.

### All Submissions:

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -3,7 +3,9 @@ name: Node CI Workflow
 on:
   push:
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
+      - V3
 
 jobs:
   build:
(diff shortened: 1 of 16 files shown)


z3dev (approved):
@platypii thanks for the fixes. TS support is becoming complex. :sigh

PR #1286, opened 2023-09-02
V3 Pre-relese Preparations

z3dev:
These changes adjust the packages as well as the rollup configurations. All packages now rollup to complete self-enclosing libraries, minimal code with un-mangled names.

NOTE: No code logic was changed.

### All Submissions:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -44,8 +44,8 @@
     "jszip": "^3.10.1"
   },
   "devDependencies": {
-    "ava": "4.3.3",
-    "c8": "8.0.0"
+    "ava": "^4.3.3",
+    "c8": "^8.0.0"
   },
   "collective": {
     "type": "opencollective",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -46,8 +46,8 @@
     "strip-bom": "4.0.0"
   },
   "devDependencies": {
-    "ava": "4.3.3",
-    "c8": "8.0.0"
+    "ava": "^4.3.3",
+    "c8": "^8.0.0"
   },
   "collective": {
     "type": "opencollective",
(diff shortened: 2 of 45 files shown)


z3dev:
@platypii @hrgdavor sorry for another large set of changes. the packages are now ready for publishing. ;)

hopefully, i can publish V3 alpha releases in the next few weeks.

platypii (approved):
Looks good 👍

hrgdavor:
@z3dev I would like to draw attention to one thing:
I am personally for moving away from automagic require('') loading STL or other formats, but it will continue to be fully supported for jscad V2 scripts in new jscad prototype. I prefer for V3 code to be as close to a regular js project, so idea is to make jscad io easily usable in any js project, not just jscad scripts. I am working on `input type=file` and from first attempts jcad/io ergonomy is bit awkward.

ideally a script would `import {deserialize} from '@jscad/io'` or more specific `import {deserializeStl} from '@jscad/io'` and then
```js
import {deserialize} from '@jscad/io'
export const main({inputFile, numOfSlices=10}){
  const object = deserialize(inputFile.name, inputFile.content)
  // return sliced object
}
```

that way such script could be used in nodejs
```js
import {readFIleSync} from 'fs'
import * as slicer from './slicer.js'

const name = process.argv[1]
const content = readFileSync(name)
const sliced = slicer.main({inputFile={name,content}})
// do something with sliced object

```

z3dev:
> @z3dev I would like to draw attention to one thing:

@hrgdavor Not sure why this was added to a closed merge request.

Wouldn't this be best as a discussion point? These ideas are going to get lost very soon.

hrgdavor:
@z3dev I was eager to make the point. I will create a discussion page. We should look more into API from a perspective of how we would like to write jscad scripts more easily. ... copied the comment to #1289 discussion

Issue #1137, opened 2022-08-21
Packaging : drop NODE 12 support


z3dev:
Node 12 is no longer supported. The supported versions should as of 2022-08 should be 14, 16, 18

See https://nodejs.org/en/about/releases/

briansturgill:
The version of node that comes with the current Ubuntu (22.04.1) LTS is 12.22.9, its end-of-life is April 2027.
To be clear, a large portion of the Linux world won't normally get an upgrade of node for quite some time.
Not a problem for me, but most people don't upgrade outside of their distributions.
I find it weird that Ubuntu lags so badly, but it's reality.  Any idea if other distributions are better about this?
Ubuntu hasn't been all that stable this time around.

Note, not a big deal on the Mac, brew is installing 18.7.0

z3dev:
Basically, dropping support means no more testing with that version, and if people drop by and say they use an older version then the first answer would be "please try a tested version".

A while back, someone was using an older version of NODE  on windows, but was able to get the test suites to run. So, who knows.

Not sure what happens when we move to pnmp/yarn and lerna-lite. Hopefully, the world will be stable no matter the choice.

briansturgill:
OK, some additional feedback. It looks like non-Debian versions are pretty much upto date.
OpenSuse lags a bit at node 14 (though 16 is in the latest).
Centos/Redhat/Fedora, Arch, Gentoo are all at 16 or 18.

Under Ubuntu you can install a snap with version 16. But all Debian variants seem to default to 12.
No idea what is going on here. I point it out to you only because Debian-based variant probably have more than half the Linux
user base. I understand your desire to move on.

platypii:
Node 12 is still widely deployed, so I would vote to not intentionally break compatibility.

But when we do decide to move forward, there are some polyfills which won't be needed anymore. For example, the [Blob](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/io/io-utils/Blob.js) polyfill won't be needed in Node 15+. See #1006

hrgdavor:
@briansturgill @briansturgill we will not intentionally break node 12 but will be less inclined to fix node 12 specific issues.

One of the steps forward will likely be ES6 and that works in node 12 https://stackoverflow.com/questions/56744272/specifying-es6-modules-in-node-12

z3dev:
There's no choice really, as this project will be using PNPM soon. PNPM requires 14+ versions of Node.js.

[...]

PR #1287, opened 2023-09-02
X3D deserializer: accept commas as MF value delimiters

andreasplesch:
### All Submissions:

see https://github.com/jscad/OpenJSCAD.org/discussions/1241#discussioncomment-6890943

I used a few scripts to refactor the parsing of MF field attribute values and to help with catching all occurrences. The tests run ok and the generated jscad script is identical to inputs without commas. It is still possible that some case was missed.


packages/io/x3d-deserializer/src/objects.js
```diff
@@ -1,3 +1,5 @@
+//sed 's|= \(element\..*[^\.]*\).trim().split.*map.*|= parseNumbers(\1)|' objects.js > objects_2.js
+//sed 's|= \(element\..*[^\.]*\).trim().split.*-1.*|= parseIndices(\1)|' objects_2.js | sed 's|.map((index) => index.trim().split(/ +/).map((v) => parseFloat(v)))||' > objects_3.js
```
andreasplesch, line 2: will be removed

andreasplesch:
Node 14 apparently does not have the replaceAll method for strings. Let me see what the node 14 testing is about. Probably there is a simple replacement.

andreasplesch:
see #1137 . Apparently Node 14 is still widely in use. Let me change to `.replace( /,/g , " " )`.

packages/io/x3d-deserializer/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+- allow commas as delimiter in MF fields
```
z3dev, line 6: Actually, the change logs are automatically updated when new releases are published to NPM. So, this change is not necessary.

there's some information here...

https://github.com/jscad/OpenJSCAD.org/wiki/Admin-:--Merging-Pull-Requests

z3dev (approved):
@andreasplesch very nice set of changes. The code looks really clean after adding this two new functions.

packages/io/x3d-deserializer/CHANGELOG.md
andreasplesch: Ok. Thanks for the info.

Issue #820, opened 2021-05-02
[Feature request] Null 3d object


ikeji:
I want to have a 3d object that have no actual object but able to pass to union() or other boolean function.
This is useful for the function that have return value conditionally.

In V1, world I can use unison([]) for this use case.
But in V2, it will make a error.

Can we have a null 3d object?

## Expected Behavior

union([]) will return empty result.

## Actual Behavior

It make error.
```
ERROR:
Uncaught Error: wrong number of arguments
```

## Steps to Reproduce the Problem

 ```
function main () {
  return union([
      cube(),
      union([]),
  ]);
}
```

## Specifications

  - Version: V2
  - Platform: linux 64
  - Environment: browser

z3dev:
@ikeji Thanks! We will take a look at this.

PR #1293, opened 2023-09-07
fix(modeling): boolean argument checks

platypii:
A user on discord had a question, and I realized that his design revealed a weird behavior in `union` when invalid geometries are given. Simplified user code:
```js
return union(vectorText({ input: 'I' }), vectorText({ input: ' ' }))
```

Rather than throwing an error in `union`, this actually returns the value `4`. Weird.

It makes sense when you look at the code. `vectorText` returns an array of points like:
```js
vectorText({ input: 'I' }) => [[[4, 21], [4, 0]]]
```
Which gets flattened by union, and then union doesn't know how to deal with non-geometries so it just returns the first element from the flattened list which is the number `4`. Very weird.

I don't like throwing errors if we can avoid it. But in this case, it feels like the right thing to do is to tell the user "hey this is not a valid geometry" as soon as possible, to make debugging easier. So in this case I added `throw new Error('union unsupported geometry type')`

I added tests for this behavior. I also tried to make the error messages easier to understand and with better grammar, but open to suggestions here.

### All Submissions:

### DIFF
--- a/packages/modeling/src/operations/booleans/scission.js
+++ b/packages/modeling/src/operations/booleans/scission.js
@@ -27,7 +27,6 @@ import { scissionGeom3 } from './scissionGeom3.js'
  */
 export const scission = (...objects) => {
   objects = flatten(objects)
-  if (objects.length === 0) throw new Error('wrong number of arguments')
 
   const results = objects.map((object) => {
     // if (path2.isA(object)) return path2.transform(matrix, object)
--- a/packages/modeling/src/operations/booleans/scission.js
+++ b/packages/modeling/src/operations/booleans/scission.js
@@ -27,6 +27,7 @@ import { scissionGeom3 } from './scissionGeom3.js'
  */
 export const scission = (...objects) => {
   objects = flatten(objects)
+  if (objects.length === 0) throw new Error('wrong number of arguments')
 
   const results = objects.map((object) => {
     // if (path2.isA(object)) return path2.transform(matrix, object)
--- a/packages/modeling/src/operations/booleans/intersect.js
+++ b/packages/modeling/src/operations/booleans/intersect.js
@@ -31,15 +31,15 @@ import { intersectGeom3 } from './intersectGeom3.js'
  */
 export const intersect = (...geometries) => {
   geometries = flatten(geometries)
-  if (geometries.length === 0) throw new Error('wrong number of arguments')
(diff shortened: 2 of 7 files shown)


z3dev:
Trash in trash out. I'm surprised that the example didn't crash.

z3dev:
Also,  V3 vectorChar and vectorText return different results now. Are you sure this still happens with V3?

platypii:
The specific example here would probably not happen with V3 vectorText because you changed it, but the behavior would remain kind of weird. If you have the above example in V3, since union doesn't support `path2`, I think it would just return the FIRST text but discard the second text.

I think this PR would still be better in that case so that users know that we don't support unioning paths (yet).

z3dev:
@platypii these changes seem reasonable. but there's an issue raised in issue #820 which seems both reasonable and possible. So, I'd really like to seem some changes for...

1. how to handle empty lists in the API
2. how to handle non-supported types in the API
3. how to handle sparse lists in the API

platypii:
@z3dev I am happy to make some suggestions for empty lists, sparse lists, and invalid geometries in general. I have some ideas. 

But I would prefer to put up a separate PR after these changes are merged, if you agree with these changes. I can have another PR up with my suggestions for the rest of the cases in the next few days. I think some of the changes for sparse lists might warrant their own discussion, and I don't want this PR to get huge.

packages/modeling/src/operations/booleans/scission.js
```diff
@@ -27,7 +27,6 @@ import { scissionGeom3 } from './scissionGeom3.js'
  */
 export const scission = (...objects) => {
   objects = flatten(objects)
-  if (objects.length === 0) throw new Error('wrong number of arguments')
```
z3dev, line 30: Seems a bit strange to remove this throw when throw are being added to other functions.

z3dev (changes requested):
@platypii looks fine. I'm looking forward to some of those thoughts in the next PR.

however I did find on inconsistency.

packages/modeling/src/operations/booleans/scission.js
platypii: I reverted this change. We can discuss removing in the next PR :-)

z3dev (approved):
@platypii thanks. looking forward to the discussions on empty/sparse lists

PR #1294, opened 2023-09-07
feat(modeling): flatten exports

platypii:
Flattens the JSCAD export hierarchy, and exports everything as a top-level export from `@jscad/modeling`. As discussed in #1289 

```js
import { circle, geom3, union, colorize, translate } from '@jscad/modeling'
```

In the process of making this change, there were some name conflict with utils. So I made some choices about exactly which util functions to export: `degToRad`, `radToDeg`, `sin`, `cos`. Also exports the constants `EPS`, `NEPS`, `TAU`.

I updated @ aliases and @ examples to reflect the new export structure.

This is a big change, especially for users. But the more I think about it, the more I agree with @hrgdavor that this would make a cleaner API that is better for users in the long run.

Thoughts @z3dev?

### All Submissions:


z3dev:
I was just expecting a new top index with flattened exports, which looks good by the way. 👍 

Are you sure that the doc changes are necessary now? I'm doing battle with JSDOC right now as the latest version breaks our documentation. And pretty much all examples in the docs have to be reviewed again. Just getting the docs back is going to be a huge task.

platypii:
I'm not certain about the doc changes, I can revert those and leave them for another PR.

I am working on updating IO to make the tests pass though with the flattened exports.

platypii:
@z3dev I rolled back the changes to doc aliases. I also got tests passing by making changes to IO and core packages. The changes to modeling are small and easy to check (+48 lines, -80 lines)

Changes to modeling: https://github.com/jscad/OpenJSCAD.org/pull/1294/commits/6a8a208751799b446bb2864064c1417f4d9e0e3f

The changes to IO were much more extensive, mostly just changing imports to use `import * from '@jscad/modeling'` and removing use of `geometries.` and `primitives.`.

packages/io/x3d-deserializer/src/translate.js
```diff
@@ -56,22 +56,9 @@ const codify = (x3d, data) => {
 // Units : ${length.name} (${length.factor})
 // Angles : ${angle.name} (${angle.factor})
 
-const {booleans, extrusions, geometries, maths, primitives, transforms, utils} = require('@jscad/modeling')
-const { colorize } =  require('@jscad/modeling').colors
-
-const applyTransform = (matrix, ...objects) => {
```
z3dev, line 62: i think this needs to be left in as the translateDefinitions() creates functions which call this special routine. X3D is pretty complex.

packages/modeling/src/index.d.ts
```diff
@@ -5,7 +5,7 @@ export * as maths from './maths'
 export * as measurements from './measurements'
 export * as primitives from './primitives'
 export * as text from './text'
-export * as utils from './utils'
+export { degToRad, flatten, radToDeg } from './utils'
```
z3dev, line 8: Please add radiusToSegments as this is for users only.

packages/modeling/src/index.js
```diff
@@ -1,15 +1,15 @@
...
-export * as text from './text/index.js'
-export * as utils from './utils/index.js'
+export * from './colors/index.js'
+export * from './curves/index.js'
+export * from './geometries/index.js'
+export * from './maths/index.js'
+export * from './measurements/index.js'
+export * from './primitives/index.js'
+export * from './text/index.js'
+export { degToRad, flatten, radToDeg } from './utils/index.js'
```
z3dev, line 8: add radiusToSegments

z3dev (changes requested):
Just a few little requests. Thanks.

I did notice some doc changes. ;)

packages/io/x3d-deserializer/src/translate.js
platypii: Sorry, I had meant to comment about this change. I appreciate that you are reviewing carefully @z3dev  :-)

It appears to me that this `applyTransform` does _exactly_ the same thing as the `transform` operation. The `transform` operation also accepts a matrix and a list of objects, and applies the transformation to each of the objects, while preserving color. From `transform.js`:

```js
export const transform = (matrix, ...objects) => {
```

So it seems like we can just delete this `applyTransform` function from the generated script, and just use `transform`. That is why I also changed the function in `x3d-deserializer/src/translateDefinitions.js`:

https://github.com/jscad/OpenJSCAD.org/pull/1294/files/809f1ad9232ebd7f8b52ddeb231e409b9f688d9b#diff-a6901a2707f9e65bf11d963375a9f993fd19496573e5d5a46041e9a0d124a7cf

```diff
-  return applyTransform(matrix, objects)
+  return transform(matrix, objects)
```

This works since we are importing everything as `import * from '@jscad/modeling'`, which includes `transform` in the flattened modeling package.

I was just cleaning up some of the imports and it seemed unnecessary so I made the change. But let me know if you'd like to me to revert this change for this PR.

z3dev: Hmm... I think these are fine. I doubled checked the reason for applyTransform(). Basically, this was a work around for the support of colors on geometries. I believe the transform() will be sufficient as you mentioned. As a side, I added a new issue to test the IO libraries as part of V3.

z3dev (approved):
@platypii once again, thanks for the improvements. The flat V3 API will be quite different for users, so we will have to create some documentation for users who are moving V2 designs to V3.

PR #1297, opened 2023-09-12
feat(modeling): text options are not optional

platypii:
This PR changes `vectorText` and `vectorChar` to have the first arg be a **required** `options = {}` parameter.

Why?
1. I don't like having a function signature that changes based on the inputs
2. there shouldn't be two different ways to pass input (either as 2nd arg, or as `input` field in options)
3. jsdoc doesn't like options having type `object|string`
4. no other functions in JSCAD work like this
5. `vectorParams` was used to handle these quirks, I removed it

I will admit that it was kind of nice to be able to just write `vectorText('string')` and now you would need to do `vectorText({}, 'string')`. But the same thing could be said about any operation with default options, like `extrudeLinear` for example. I think it's better to be explicit about requiring options.

I don't really feel that strongly about this particular issue. But if we're going to make API changes, now is the time to do it for V3. Let me know what you guys think.

### All Submissions:


z3dev (approved):
@platypii thanks. these changes make the text functions far easier to understand.

PR #1298, opened 2023-09-14
Declare Typescript modules in IO packages

bruceborrett:
### All Submissions:


platypii (commented):
Overall this seems fine. It would be helpful if you gave some additional details in the PR write up however. Questions:
 - What is this fixing?
 - How can I repro/verify the changes have made an improvement?
 - What good does it do to `declare module` for the IO packages if they don't actually contain any type definitions yet?
 - We have usually followed the convention of putting `.d.ts` files in the `src` directory, next to the files they provide types for. So I think it would be more consistent to move your `types/index.d.ts` files to `src/index.d.ts` and update `package.json` accordingly. Is there a benefit to having the types in their own `types` directory?

bruceborrett:
Sorry for the lack of information in this PR.

The issue I was having was when I tried to import an IO package in my project I would get the following error:

> Could not find a declaration file for module '@jscad/stl-serializer'.  Try `npm i --save-dev @types/jscad__stl-serializer` if it exists or add a new declaration (.d.ts) file containing `declare module '@jscad/stl-serializer';`

Simply adding the module declaration resolved the error, Im not sure if this is the best way to solve this but may be a good enough interim solution until proper Typescript support is added?

I added a seperate types directory as I noticed that is the way it was done in the regl-renderer package, did not check the other packages, I can move it to src/index.d.ts if you prefer?

z3dev:
> Simply adding the module declaration resolved the error, Im not sure if this is the best way to solve this but may be a good enough interim solution until proper Typescript support is added.

After adding this interim solution, was there any typing available for the deserialize / serialize functions? Were you able to use the exports of the IO packages?

z3dev:
@bruceborrett sorry. that was a question for you. Were you able to import and use the IO libraries after these definitions were added?

Also, please move the type definitions into src/index.d.ts

z3dev:
@bruceborrett Sorry. Can you move the changes to src/index.d.ts ?

bruceborrett:
Im sorry I have not had time to come back to this, I think its better if I just close this PR. As you rightfully pointed out in your first comment above, "what good does it do to declare module for the IO packages if they don't actually contain any type definitions yet". If this has not been an issue for anyone else, then possibly there was just some misunderstanding from my side in what the issue actually is and unfortunately I do not have time to further investigate or improve on this at the moment.

Issue #1249, opened 2023-05-17
Suggestion for V3 API


hjfreyer:
I happened to see this wiki page and I thought I'd offer my unsolicited opinion: https://github.com/jscad/OpenJSCAD.org/wiki/JSCAD-V3

I really like jscad, but I often find the APIs surprising, particularly with respect to how operations affect groups of geometries I pass to them. 

For instance, I'm modeling a table right now, and I have an array of two `legs` and a `plate` that they'll connect to. I just ran this:

```
[legs, plate] = align({modes: ['none', 'center', 'none'], grouped: true}, [legs, plate]);
```

However, this didn't work. With `console.log`, I saw that the align operation had converted my argument with structure `[[leg1, leg2], plate]` into a flat array: `[leg1, leg2, plate]`. I run into this kind of confusion all the time.

My advice would be to consistently treat arrays of geometries as single objects, or perhaps introduce a first-class "group" type. For instance, `centerX([a, b, c])` shouldn't center each object individually. If I wanted that behavior, I'd write `[a, b, c].map(x => centerX(x))`. Some operations may still want to recurse over the structure (e.g. `intersect([a, [b, c]], d)` should go in and intersect `d` with each item in the first argument), but it shouldn't disrupt the structure I provided.

Thanks for the great library! Hope this helps.

z3dev:
@hjfreyer nice suggestions. thanks.

The suggestion to 'align' based on the structure of the array seems reasonable, but probably requires another 'option'. Feel free to submit some API changes. We are working on V3 now, and have already broken a few of the APIs.

The reason for the flattening of objects is to treat the parameters in a consistent manner. For example...

```
let results = align({}, circle())
results = align({}, circle(), square())
results = align({}, ...contents)
results = align({}, [a, b, c])
```

Yeah. It probably doesn't make sense for lots of people, nor all the time, but at least it's consistent throughout the functions of JSCAD. No surprises. All input is flattened.

z3dev:
@hjfreyer V3 recently had a major change to the APIs in general.

I think that @platypii could add a few comments on the changes here.

PR #1299, opened 2023-09-19
feat(modeling): input array structure

platypii:
This PR contains my suggestions for JSCAD v3 handling of input arrays. Some of these are breaking changes from V2. I tried to follow a couple principles:

1. Avoid throws unless the operation would be clearly invalid
2. Preserve input structure when possible, don't flatten unless we need to. This is nice for grouping objects, and also preserves information when composing operations.
3. Accept nested arrays anywhere a single geometry is accepted. So anywhere a user could pass a geometry, should also accept nested arrays of geometries.
4. Make choices that give users more control. If there's two choices for how we define an API, but one way gives users more flexibility, then do that. See `align` below.

Before this PR:
|                            | returns    | Input arrays | Mix 2d 3d    | Empty input  | null/undefined | Invalid geom |
|----------------------------|------------|--------------|--------------|--------------|----------------|--------------|
| Booleans, extrusions, hulls, offsets | Geometry  | flatten      | throw        | []           | filtered       | throw*       |
| modifiers                   | Geometry[] | flatten      | apply to each| []           | throw          | throw*       |
| transforms                  | Geometry[] | flatten      | apply to each| return input | copy to output | throw*       |

*throws for obviously incorrect geometries, but will incorrectly flatten and return some things like `[[1,2],[3,4]]`

After this PR:
|                            | returns                    | Input arrays          | Mix 2d 3d    | Empty input  | null/undefined | Invalid geom |
|----------------------------|----------------------------|-----------------------|--------------|--------------|----------------|--------------|
| Booleans, extrusions, hulls, offsets | Geometry                   | flatten               | throw        | []           | filtered       | throw        |
| modifiers                   | RecursiveArray<Geometry>    | Preserve input structure | apply to each | return input | copy to output  | throw        |
| transforms                  | RecursiveArray<Geometry>    | Preserve input structure | apply to each | return input | copy to output  | throw        |

## Nested Arrays as Groups
I suggest that we preserve nested arrays where it makes sense to do so. This information can be treated like “groups” in svg and other file formats. For operations which return a single geometry, like the booleans, we should still flatten before we apply the operation, but if it makes sense to do so, I propose that we preserve the recursive structure as much as possible.

## null/undefined
I suggest that we filter out null-ish values when we do a flatten operation for booleans, and most operations. To make working with nested arrays easier, I added a new util function `coalesce` which combines flatten and filtering out nullish values.

```ts
export function coalesce<T>(arr: RecursiveArray<T>): Array<T>
```

## Empty arrays
What if a user calls a boolean operation with an empty array, or an array which is empty when flattened? We have three options 1) throw error, but I would prefer not to do this 2) return geom2.create() or geom3.create(), but the problem is that we don’t actually know if it’s intended to be 2D or 3D. I propose that in this case we return undefined, which can be filtered out by other operations using coalesce.

## Mix 2D/3D
If you call a boolean between a 2D and 3D geometry, it’s not clear what to return in that case, so we should throw if not all types are the same. However for transforms, and modifiers, it’s okay to mix 2D and 3D. I propose that we apply functions where it makes sense. If an object is nullish, or not a geometry, I suggest we simply map the object to itself and return it in the output. This makes the operation robust against unexpected inputs.

## Alignment
One other suggestion is that I believe that align and center operations should respect groupings, and use them when aligning objects rather than flattening. See suggestion from @hjfreyer in #1249. The reason I like this change is that it gives more flexibility to users. In V2, inputs will be flattened regardless of the user's intent. Whereas with this change, users can align groups of objects. The previous behavior can be accessed by setting `options.grouped = true`.

## Examples

```js
const g3 = h3 = cube()
const g2 = h2 = square()

generalize({}, g3) => out3
generalize({}, g2, g3) => [out2, out3]
generalize({}, [[g2], g3]) => [[out2], out3]
generalize({}, []) => []

offset({}, g3) => out3
offset({}, g2, g3) => [out2, out3]
offset({}, [[g2], g3]) => [[out2], out3]
offset({}, []) => []

union(g3) => g3
union(g2) => g2
union(g2, null) => g2
union(g2, undefined) => g2
union([g2, undefined, null]) => g2
union(g3, h3) => out3
union(g2, h2) => out2
union([[g2], h2]) => out2
union(g2, g3) => throw “union arguments must be the same geometry type”
union([]) => undefined

scission(g3, h3) => [[g3a, g3b], h3]
```

### All Submissions:


z3dev:
@platypii sorry for the lack of a review. hopefully, i can play with the new style this weekend.

platypii:
Just to add a bit more justification for this PR:

One of my goals is to improve things like SVG serialize/deserialize. I want to be able to convert SVG groups into nested sub-arrays of geometries. And then it would be really nice if information was not lost along the way. So for example: 1) Load geometry from SVG, 2) Apply `transform`, or `extrude`, or `offset`, or `align`, etc to the nested geometry, 3) Re-export to SVG while still preserving the nested group structure.

Currently that information is lost when any operation is applied to a design because arrays get flattened. This PR attempts to preserve that structure where possible.

z3dev (commented):
A few small comments. And a few questions...

packages/modeling/src/operations/transforms/center.js
```diff
@@ -43,8 +41,6 @@ export const center = (options, ...objects) => {
   }
   const { axes, relativeTo } = Object.assign({}, defaults, options)
 
-  objects = flatten(objects)
-  if (objects.length === 0) throw new Error('wrong number of arguments')
   if (relativeTo.length !== 3) throw new Error('relativeTo must be an array of length 3')
 
   options = { axes, relativeTo }
```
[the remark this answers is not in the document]
platypii: You're right, good catch. I added a test for this too.

packages/modeling/src/operations/hulls/hull.js
```diff
@@ -33,8 +33,10 @@ import { hullGeom3 } from './hullGeom3.js'
  *       +-------+           +-------+
  */
 export const hull = (...geometries) => {
-  geometries = flatten(geometries)
-  if (geometries.length === 0) throw new Error('wrong number of arguments')
+  geometries = coalesce(geometries)
```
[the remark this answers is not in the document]
platypii: I think you may be misunderstanding the effect of `coalesce`. It takes input array, and flattens the array while also remove null/undefined values. The output of `coalesce` is always a flat array of non-null objects.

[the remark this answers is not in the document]
platypii: Added a test too

packages/modeling/src/operations/booleans/union.js
```diff
@@ -29,13 +29,14 @@ import { unionGeom3 } from './unionGeom3.js'
  *      +-------+            +-------+
  */
 export const union = (...geometries) => {
-  geometries = flatten(geometries)
-  if (geometries.length === 0) throw new Error('union wrong number of arguments')
+  geometries = coalesce(geometries)
```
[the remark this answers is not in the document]
platypii: See comment above about the output of coalesce.

platypii:
@z3dev updated PR based on feedback, and added a few more tests to try and make things clearer. Thanks!

packages/modeling/src/operations/booleans/intersect.js
```diff
@@ -30,13 +30,14 @@ import { intersectGeom3 } from './intersectGeom3.js'
...
 export const intersect = (...geometries) => {
-  geometries = flatten(geometries)
-  if (geometries.length === 0) throw new Error('intersect wrong number of arguments')
+  geometries = coalesce(geometries)
 
   if (!areAllShapesTheSameType(geometries)) {
     throw new Error('intersect arguments must be the same geometry type')
   }
 
+  if (geometries.length === 0) return undefined
```
z3dev, line 39: maybe move up above areAllShapesTheSameType()?

packages/modeling/src/operations/booleans/intersectGeom2.js
```diff
@@ -1,17 +1,13 @@
-import { flatten } from '../../utils/flatten.js'
-
 import { INTERSECTION } from './martinez/operation.js'
 import { boolean } from './martinez/index.js'
 
 /*
  * Return a new 2D geometry representing space in both the first geometry and
  * in the subsequent geometries. None of the given geometries are modified.
- * @param {...geom2} geometries - list of 2D geometries
+ * @param {...Geom2} geometries - a flat list of 2D geometries
```
z3dev, line 7: Geom2[]

packages/modeling/src/operations/booleans/intersectGeom3.js
```diff
@@ -1,18 +1,14 @@
...
-
 import { retessellate } from '../modifiers/retessellate.js'
 
 import { intersectGeom3Sub } from './intersectGeom3Sub.js'
 
 /*
  * Return a new 3D geometry representing space in both the first geometry and
  * in the subsequent geometries. None of the given geometries are modified.
- * @param {...geom3} geometries - list of 3D geometries
+ * @param {...Geom3} geometries - a flat list of 3D geometries
```
z3dev, line 8: Geom3[]

packages/modeling/src/operations/booleans/subtract.js
```diff
@@ -30,13 +30,14 @@ import { subtractGeom3 } from './subtractGeom3.js'
...
 export const subtract = (...geometries) => {
-  geometries = flatten(geometries)
-  if (geometries.length === 0) throw new Error('subtract wrong number of arguments')
+  geometries = coalesce(geometries)
 
   if (!areAllShapesTheSameType(geometries)) {
     throw new Error('subtract arguments must be the same geometry type')
   }
 
+  if (geometries.length === 0) return undefined
```
z3dev, line 39: move up?

packages/modeling/src/operations/booleans/subtractGeom2.js
```diff
@@ -1,17 +1,13 @@
-import { flatten } from '../../utils/flatten.js'
-
 import { DIFFERENCE } from './martinez/operation.js'
 import { boolean } from './martinez/index.js'
 
 /*
  * Return a new 2D geometry representing space in the first geometry but
  * not in the subsequent geometries. None of the given geometries are modified.
- * @param {...geom2} geometries - list of geometries
+ * @param {...Geom2} geometries - a flat list of 2D geometries
```
z3dev, line 7: Geom2[]

packages/modeling/src/operations/booleans/subtractGeom3.js
```diff
@@ -1,18 +1,14 @@
...
-
 import { retessellate } from '../modifiers/retessellate.js'
 
 import { subtractGeom3Sub } from './subtractGeom3Sub.js'
 
 /*
  * Return a new 3D geometry representing space in this geometry but not in the given geometries.
  * Neither this geometry nor the given geometries are modified.
- * @param {...geom3} geometries - list of geometries
+ * @param {...Geom3} geometries - a flat list of 3D geometries
```
z3dev, line 8: Geom3[]

packages/modeling/src/operations/booleans/union.js
```diff
@@ -29,13 +29,14 @@ import { unionGeom3 } from './unionGeom3.js'
...
 
   if (!areAllShapesTheSameType(geometries)) {
     throw new Error('union arguments must be the same geometry type')
   }
 
+  if (geometries.length === 0) return undefined
```
z3dev, line 38: move up?

packages/modeling/src/operations/booleans/unionGeom2.js
```diff
@@ -1,16 +1,12 @@
-import { flatten } from '../../utils/flatten.js'
-
 import { UNION } from './martinez/operation.js'
 import { boolean } from './martinez/index.js'
 
 /*
  * Return a new 2D geometry representing the total space in the given 2D geometries.
- * @param {...geom2} geometries - list of 2D geometries to union
+ * @param {...geom2} geometries - a flat list of 2D geometries to union
```
z3dev, line 6: Geom2[]

packages/modeling/src/operations/booleans/unionGeom3.js
```diff
@@ -1,17 +1,13 @@
-import { flatten } from '../../utils/flatten.js'
-
 import { retessellate } from '../modifiers/retessellate.js'
 
 import { unionGeom3Sub } from './unionGeom3Sub.js'
 
 /*
  * Return a new 3D geometry representing the space in the given 3D geometries.
- * @param {...objects} geometries - list of geometries to union
+ * @param {...Geom3} geometries - a flat list of 3D geometries to union
```
z3dev, line 7: Geom3[]

packages/modeling/src/operations/hulls/hullChain.js
```diff
@@ -30,12 +31,16 @@ import { hull } from './hull.js'
  *       +-------+                +-------+
  */
 export const hullChain = (...geometries) => {
```
z3dev, line 33: I just noticed... can hullChain() handle path2 geometries?

packages/modeling/src/operations/hulls/hullGeom3.js
```diff
@@ -8,12 +6,10 @@ import { toUniquePoints } from './toUniquePoints.js'
 
 /*
  * Create a convex hull of the given geometries (geom3).
- * @param {...geometries} geometries - list of geom3 geometries
+ * @param {...Geom3} geometries - a flat list of 3D geometries
```
z3dev, line 9: Geom3[]

packages/modeling/src/operations/transforms/align.test.js
```diff
@@ -53,7 +53,7 @@ test('align: multiple objects ungrouped returns geometry aligned, different mode
     cube({ size: 4, center: [10, 10, 10] }),
     cube({ size: 2, center: [4, 4, 4] })
   ]
-  const aligned = align({ modes: ['center', 'min', 'max'], relativeTo: [30, 30, 30] }, original)
+  const aligned = align({ modes: ['center', 'min', 'max'], relativeTo: [30, 30, 30] }, ...original)
```
z3dev, line 56: why the change? shouldn't align() determine what to do with an array of objects versus a single object?

z3dev (changes requested):
The changes to handle arrays recursively was BIG.

Some small requests to change comments. And one question...

packages/modeling/src/operations/booleans/subtractGeom2.js
platypii: I am happy to make these changes, but I am curious on your thoughts as to why?

"..." is legal in jsdoc as a way to define something as an array type. So `...Geom2` and `Geom2[]` mean the same thing. I fixed the capitalization of the types in this PR, but I just left it as "..." for consistency. Should we make a lint rule that prefers [] over ...?

packages/modeling/src/operations/transforms/align.test.js
platypii: This is related to the change in `align` behavior that was discussed in the original PR description. Basically I am proposing a change where each separate argument to `align` will be aligned as a group, but different top-level args are treated separately.

Previously, in V2, the behavior was to flatten everything:

`align({}, [a, b], [c, d]) === [align({}, a), align({}, b), align({}, c), align({}, d)]`
`align({}, [[a, b], [c, d]]) === [align({}, a), align({}, b), align({}, c), align({}, d)]`

Whereas with this PR I propose that:

`align({}, [a, b], [c, d]) === [align({}, [a, b]), align({}, [c, d])]`
`align({}, [[a, b], [c, d]]) === align({}, [[a, b], [c, d]])` (cannot be decomposed because the whole array is grouped)

This was probably one of the more debatable changes in this PR, but I included it based on my experience using `align` for designs with groups, and feedback in #1249. Following design principle number 4:  This gives more flexibility to the users, because users can change the grouping (multiple args, vs one array arg), they can select different alignment behavior. Whereas before users had less flexibility.

packages/modeling/src/operations/booleans/subtractGeom2.js
z3dev: Hmm... I suggested the change because the API is really expecting an array now, rather than an arbitrary number of parameters. Does that make sense?

I don't really have a preference on the capitalization, etc in the documentation but jsdoc needs to spit out something nice for the users.

I suspect that there will be a few rounds of documentation changes for V3.

packages/modeling/src/operations/transforms/align.test.js
z3dev: That's good documentation for the V2 -> V3 migration, and probably worth saving somewhere.

OK. So, going back to the test. I now see the subtle difference in the V3 API, which changed the results of the test. Shouldn't the test be changed to verify the new results, rather then tweaking the input to '...original'? Or even better, change the test to pass in two parameters instead of the array, which is probably what most users will do.

All of this is probably just 'style' debates. Sorry.

packages/modeling/src/operations/hulls/hullChain.js
platypii: It can. I updated jsdoc to match behavior

z3dev (approved):
@platypii thanks again.

sorry that i haven't been active lately... it's part of the busy season.

PR #1300, opened 2023-09-24
test(modeling): measureArea for all tests

platypii:
Improvements to tests by adding `measureArea` to as many tests as possible. Continues the work started in #1197 

Volume and Area tests are nice because they provide a check against all the vertices of a geometry, but don't depend on the exact order of vertices. These measures should be stable, even if we change underlying geometry algorithms. The goal is to have solid tests so that we can make changes with confidence.

In a couple places, I changed unnecessary `t.deepEquals` to `t.is` tests.

### All Submissions:


z3dev (approved):
@platypii nice improvements. Someday, we can take out the unnecessary checks of point sets.

the improvements to the 2D space is really nice.

PR #1302, opened 2023-10-03
fix(modeling): geom2 mirroring transform

### LINKED ISSUE #1301: Union of 2D polygon and its mirrorX doesn't work
## Expected Behavior
Have full design from its half.

## Actual Behavior
Only have one half of design made from union of a polygon and mirrorX of itself.

```
  let P =  [[0,30],[-16,13],[-10,13],[-23,-3],[-16,-3],[-32,-21],[-16,-21],[-16,-30],[13,-30],[13,-20],[28,-20],[13,-2],[19,-2],[8,13],[13,13]]

  let half2D = polygon({points:P})
  let mirroredHalf2D = mirrorX(half2D)

  let full2D = union(half2D, mirroredHalf2D)

  return full2D
```
## Specifications

  - Version: latest online
  - Platform: latest KUbuntu 64 bits
  - Environment: latest Firefox 64 bits

platypii:
Fixes #1301 (thanks @gilboonet)

When a mirroring transform is applied to a geometry, we should reverse the point order so that orientation is preserved.

This was already done for geom3/poly3 geometries [here](https://github.com/jscad/OpenJSCAD.org/blob/master/packages/modeling/src/geometries/poly3/transform.js#L15-L18), but was not applied to geom2.

This PR uses the 2D determinant in `geom2.transform` to check if a transform is a mirroring transform in 2D.

Added to V3 branch since this is technically a breaking change. Although I would argue this is a fix of bad behavior.

### All Submissions:

### DIFF
--- a/packages/modeling/src/geometries/geom2/transform.js
+++ b/packages/modeling/src/geometries/geom2/transform.js
@@ -17,7 +17,7 @@ import { reverse } from './reverse.js'
 export const transform = (matrix, geometry) => {
   const transforms = mat4.multiply(mat4.create(), matrix, geometry.transforms)
   const transformed = Object.assign({}, geometry, { transforms })
-  // 2D determinant
+  // determine if the transform is mirroring in 2D
   if (matrix[0] * matrix[5] - matrix[4] * matrix[1] < 0) {
     // reverse the order to preserve the orientation
     return reverse(transformed)
--- a/packages/modeling/src/geometries/geom2/transform.js
+++ b/packages/modeling/src/geometries/geom2/transform.js
@@ -1,5 +1,7 @@
 import * as mat4 from '../../maths/mat4/index.js'
 
+import { reverse } from './reverse.js'
+
 /**
  * Transform the given geometry using the given matrix.
  * This is a lazy transform of the outlines, as this function only adjusts the transforms.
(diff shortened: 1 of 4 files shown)


packages/modeling/src/geometries/geom2/transform.test.js
```diff
@@ -50,3 +56,38 @@ test('transform: adjusts the transforms of geom2', (t) => {
   t.true(comparePoints(another.outlines[0], expected.outlines[0]))
   t.true(compareVectors(another.transforms, expected.transforms))
 })
+
+test('transform: mirroring transform geom2', (t) => {
+  const geometry = square()
+  const matrix = mat4.fromScaling(mat4.create(), [-1, -1, -1])
```
z3dev, line 62: used?

packages/modeling/src/geometries/geom2/transform.test.js
```diff
@@ -50,3 +56,38 @@ test('transform: adjusts the transforms of geom2', (t) => {
...
+  const exp = [[1, 1], [-1, 1], [-1, -1], [1, -1]]
+  t.true(comparePoints(pts, exp))
+  t.deepEqual(toSides(transformed), [
+    [[1, 1], [-1, 1]],
+    [[-1, 1], [-1, -1]],
+    [[-1, -1], [1, -1]],
+    [[1, -1], [1, 1]]
+  ])
+})
+
```
z3dev, line 77: how about testing mirrorY?

packages/modeling/src/geometries/geom2/transform.js
```diff
@@ -14,5 +16,11 @@ import * as mat4 from '../../maths/mat4/index.js'
  */
 export const transform = (matrix, geometry) => {
   const transforms = mat4.multiply(mat4.create(), matrix, geometry.transforms)
-  return Object.assign({}, geometry, { transforms })
+  const transformed = Object.assign({}, geometry, { transforms })
+  // 2D determinant
```
z3dev, line 20: `// determine if the transform is mirroring in 2D`

z3dev (changes requested):
@platypii a few nits. but nice changes.

packages/modeling/src/geometries/geom2/transform.test.js
platypii: not used, good catch

z3dev (approved):

Issue #1321, opened 2024-02-11
plane.fromPoints() implementation buggy


Hermann-SW:
## Expected Behavior

For new testcase
```
const obs5 = fromPoints(obs1, [-1, -1, -1], [1, -1, -1], [-1, 1, -1], [1, 1, -1])
```
all pointes lie on plane z==-1, and the normalized normal vector returned should be either [0,0,-1] or [0,0,1].

## Actual Behavior

Instead [0,0,0] is returned which cannot be a normal vector. Reason is that two times normal vector computed in loop is [0,0,1], and two times it is [0,0,-1], and summation cancels them 0 as [0,0,0]. Summaion is just wrong. 

## Steps to Reproduce the Problem

  1. add this test to OpenJSCAD.org/packages/modeling/src/maths/plane/fromPoints.test.js:
   ```
pi@raspberrypi5:~/old/OpenJSCAD.org/packages/modeling/src/maths/plane $ git diff fromPoints.test.js
diff --git a/packages/modeling/src/maths/plane/fromPoints.test.js b/packages/modeling/src/maths/plane/fromPoints.test.js
index 7e924d3b..b4c9297a 100644
--- a/packages/modeling/src/maths/plane/fromPoints.test.js
+++ b/packages/modeling/src/maths/plane/fromPoints.test.js
@@ -17,4 +17,7 @@ test('plane: fromPoints() should return a new plane with correct values', (t) =>
   // polygon with co-linear vertices
   const obs4 = fromPoints(obs1, [0, 0, 0], [1, 0, 0], [2, 0, 0], [0, 1, 0])
   t.true(compareVectors(obs4, [0, 0, 1, 0]))
+
+  const obs5 = fromPoints(obs1, [-1, -1, -1], [1, -1, -1], [-1, 1, -1], [1, 1, -1])
+  t.true(compareVectors(obs5, [0, 0, 1, -1]))  
 })
pi@raspberrypi5:~/old/OpenJSCAD.org/packages/modeling/src/maths/plane $ 
```
  2. run 
```
pi@raspberrypi5:~/old/OpenJSCAD.org/packages/modeling/src/maths/plane $ ../../../../../node_modules/.bin/ava fromPoints.test.js 

  plane: fromPoints() should return a new plane with correct values

  src/maths/plane/fromPoints.test.js:22

   21:   const obs5 = fromPoints(obs1, [-1, -1, -1], [1, -1, -1], [-1, 1, -1]…
[...]

PR #1322, opened 2024-02-12
fix(modeling): added fromNoisyPoints() to plane

z3dev:
Thanks to @Hermann-SW , the current implement of plane.fromPoints() was found to be very limited, requiring a set of vertices that is both planar and ordered.

See #1321 for the issue.

These changes add a new function to calculate the best-fit plane across any set of vertices. Testing has shown some differences in precision, and fromNoisyPoints() always produces a normal in the positive orientation.

Given that fromPoints() is used internally and produces a directional normal, the new function was created.

### All Submissions:


z3dev:
@platypii What do you think? I was kind of surprised at the current implementation of fromPoints(). It's correct but requires careful positioning of vertices. But in the general sense, a best-fit version is probably better for users.

platypii (approved):
Looks good to me. I can imagine scenarios where this might be helpful.

PR #1323, opened 2024-02-17
feat(modeling): V2 api compatibility shim

platypii:
This PR adds a "V2 compatibility shim" to V3.

In PR #1294 we flattened the JSCAD package heirarchy so that functions like `union` and `extrudeLinear` are now in the top level package object:

```diff
-import { booleans } from '@jscad/modeling'
-const { union } = boolean
+import { union } from '@jscad/modeling'
```

But what about all the hundreds of designs out there for V2? They will all break, annoyingly, in the migration to V3.

This PR keeps the flattened structure for V3 but ALSO exports a similar structure to V2. This is not 100% api compatible, due to some fundamental changes in V3, but it is pretty close, and I tried a couple of my designs from V2 against this, and they actually all just-worked!

Feedback welcome, but I think this would make the transition from V2 to V3 much smoother for users of JSCAD.

### All Submissions:

### DIFF
--- a/packages/modeling/src/index.js
+++ b/packages/modeling/src/index.js
@@ -13,3 +13,44 @@ export * from './operations/hulls/index.js'
 export * from './operations/modifiers/index.js'
 export * from './operations/offsets/index.js'
 export * from './operations/transforms/index.js'
+
+// V2 API compatibility:
+export * as colors from './colors/index.js'
+export * as curves from './curves/index.js'
+import { geom2, geom3, path2, poly2, poly3 } from './geometries/index.js'
+export const geometries = {
+  geom2: {
+    ...geom2,
+    create: (sides) => geom2.fromSides(sides),
+    fromPoints: (points) => geometries.geom2.create([points])
+  },
+  geom3,
+  path2,
+  poly2,
+  poly3: {
+    ...poly3,
+    fromPoints: (points) => poly3.create([points]),
+    fromPointsAndPlane: poly3.fromVerticesAndPlane,
+    toPoints: poly3.toVertices,
+  },
+}
+export * as maths from './maths/index.js'
+export * as measurements from './measurements/index.js'
+export * as primitives from './primitives/index.js'
+export * as text from './text/index.js'
+export * as booleans from './operations/booleans/index.js'
+import * as extrusion from './operations/extrusions/index.js'
+export const extrusions = {
+  ...extrusion,
+  extrudeRectangular: (opt, geom) => extrusions.extrudeLinear(opt, offsets.offset(opt, geom)),
+  slice: geometries.slice,
+}
+export * as hulls from './operations/hulls/index.js'
+export * as modifiers from './operations/modifiers/index.js'
+import * as offsets from './operations/offsets/index.js'
+export const expansions = {
+  ...offsets,
+  expand: offsets.offset,
+}
+export * as transforms from './operations/transforms/index.js'
+export * as utils from './utils/index.js'


hrgdavor (approved):
this will help with transition to V3, and I do not see a negative impact

z3dev:
I kind of like it and kind of don't, but as @hrgdavor said... these changes help in the transition from V2 to V3.

But is it enough? Would a full V2 compatible shim be better?

z3dev:
I get it but a little skeptical.

To use the shim requires changes to the imports. So, why not just make a V2 fully compatible shim, which will also requires changes to imports?

I'm not sure which is better.

V2 was criticized by those that like the 'object' chaining. But it provided a new core set of functionality, which surpassed V1 in many ways. Hopefully, V3 will also do the same.

I think the shim is fine, but really think it should be V2 as much as possible. Is it possible?

platypii:
> To use the shim requires changes to the imports.

What do you mean? By adding this shim, users do NOT need to change imports from V2 designs. With the shim, this will still work:

```js
const { cube } = jscad.primitives

const main = () => cube({ size: 8 })

module.exports = { main }
```

But so will the more modern V3 syntax:

```js
import { cube } from '@jscad/modeling'
export const main = () => cube({ size: 8 })
```

This seems like the best of both worlds. We can deprecate the old syntax, and remove the shim in V4?

z3dev:
Interesting. So, this shim is just for transcompiling COMOMJS to ES6.

z3dev:
So there are several things being discussed here...

- transcompiling CommonJS to ES5
- adding the V2 namespaces back to V3
- adding full or partial V2 compatibility
- support by the CLI

For transcompiling, this is already supporting in the JSCADUI (Nice!).

Adding the V2 namespaces back seems reasonable, as that will ease the transition of designs from V2 to V3.

As for the **full** V2 compatibility, I think this should be possible but some special functions have to be written and tested. This compatibility shim could be part of this package, but it would have to be imported specifically by designs. And the compatibility shim might be better as a separate package, again possible.

Also, I think @hrgdavor is now looking at some ways to identify V1 vs V2 vs V3 designs. It would be cool if this could be automatic. But maybe a design can select the API version in a comment, e.g. 'V2'.

Finally, the CLI may be another problem. The CLI may need a total re-write to support transcompiling, different versions of the API, etc.

z3dev:
@platypii can we focus these changes on just part 2, adding the V2 namespaces back to V3? it's shouldn't be that different.

then we can decide where to put the V2 compatibly shim, and make it 100% if possible.

hrgdavor:
> So, the dynamic imports really don't help? Big shock!

I did not try to inject override for dynamic import. Trying to eval a script that uses the `import .... from '...'` just fails, so at that point it seemed just better to convert to CJS require. Luckily with transpilers working so well for years, this not a big deal, and with sourcemaps, you can still see proper error reporting and debugging.

also there was compatibility issue for dynamic import in workers last time I looked (if I remember correctly).

platypii:
> can we focus these changes on just point # 2, adding the V2 namespaces back to V3? it's shouldn't be that different.
> then we can decide where to put the V2 compatibly shim, and make it 100% compatible, if possible.

I am not clear on what you are asking for here? You want the exports in the v2 and v3 namespaces, but none of the compatibility functions? If that is the intent I have to say... why?? You are asking for a world where someone who had a project that used V2 and now upgrades to V3 will have imports that seem to work, but break in mysterious ways. That seems less compatible, and worse than not having the V2-like exports at all!

If you are worried about bloat from adding the functions here? I would agree, except this PR is only 41 lines of code and provides backward compatibility for most jscad V2 designs to upgrade seamlessly to V3. It is a tiny addition, with great benefits. I'm not sure why we wouldn't do this for users? (See #1326 for example of user pain)

z3dev:
> > can we focus these changes on just point # 2, adding the V2 namespaces back to V3? it's shouldn't be that different.
> > then we can decide where to put the V2 compatibly shim, and make it 100% compatible, if possible.
> 
> I am not clear on what you are asking for here? 

In simple terms, a 100% compatible V2 shim. These changes look good initially but there are differences still. And V3 changes are not complete yet. For example, the use of a shared set of attributes for color, etc.

Also, these V2 compatibility changes are being added directly into the V3 API. V2 and V3 in the same package? Why? Will V1 be added as well? There must be another way to handle this.

platypii:
Being totally honest here... The resistance to this PR is why I stopped contributing to JSCAD about a year ago.

If we completely break the jscad api from V2 to V3 I am convinced this project will die. There are countless examples where a project makes major API changes, without careful thought on how users will migrate, and the result is that the project dies. If people come to JSCAD, drop a design that has worked for _years_ and in V3 everything breaks, no one is going to fight to re-write all their designs, they will probably just go find another project.

So I made this PR. Which provides almost complete backward compatibility so that V2 designs will keep working with V3. And yet the reaction to this PR was "not good enough unless it's 100% compatible". Except thats impossible with the changes to the V3 internal representation. This PR is as close to backward compatibility as we're going to get. So why can't we merge it?

z3dev:
Done.
But don't confuse questions with resistance. I'm still not convinced 100% but did see the value after @hrgdavor mentioned that transcompiling is required for projects. JS imports are a total disappointment.

platypii:
Thanks z3. Sorry if I was harsh earlier but I really believe in the importance of respecting the user. I love the JSCAD project so it would make me sad if people gave up using it because of major breaking changes. Particularly if we can do something to make it less of an issue. :pray:

z3dev:
No problem.

 I was hoping that some kind of V2 shim could be selected in the jscadapp, and allow V2 designs to work without any changes. But JS imports are not able to handle this.

I was also pretty shocked that Babel is being used as well. :(

hrgdavor:
> I was hoping that some kind of V2 shim could be selected in the jscadapp

That is certainly doable, and not very complicated.

Also V3 designs can use imports, and V2 scripts 99.99% only use require. So we can even be smart about it and make a good initial guess (import = V3, require = V2). Also on first load of script, if we guess V2/V3 wrong, we can re-run with the other. Also multifile projects for V3 should be advised to use package.json and then we have exact version.

Although shim is not ideal in a sense, it can also be very valuable when refactoring  from V2 to V3, a script or multifile project.

## V1 !!
As bonus, I have initial version of V1 support https://github.com/hrgdavor/jscadui/tree/feat-v1-support. That way we are not leaving any users behind. V! scipt is easily recognizable by `.jscad` extension.

I have tried jscad examples from v1.
demo: https://3d.hrg.hr/jscad/v1-support/#https://raw.githubusercontent.com/jscad/OpenJSCAD.org/refs/heads/V1/packages/examples/logo.jscad


it uses a shim, and this intentional. I do not want to hide the fact that those V1 scripts can be be migrated to regular JS, that is not some fancy scripting language.

Header shim

```js
const csg = require('@jscad/csg')
const { OpenJsCad, debug } = csg
const {circle, square, polygon, triangle} = csg.primitives2d
const {cube, sphere, cylinder, geodesicSphere, torus, polyhedron} = csg.primitives3d
const {union, difference, intersection} = csg.booleanOps
const {translate, center, scale, rotate, transform, mirror, expand, contract, minkowski, hull, chain_hull} = csg.transformations
const {extrudeInOrthonormalBasis, extrudeInPlane, extrude, linear_extrude, rotate_extrude, rotateExtrude, rectangular_extrude} = csg.extrusions
const {css2rgb, color, rgb2hsl, hsl2rgb, rgb2hsv, hsv2rgb, html2rgb, rgb2html} = csg.color
const {sin, cos, asin, acos, tan, atan, atan2, ceil, floor, abs, min, max, rands, log, lookup, pow, sign, sqrt, round} = csg.maths
const {vector_char, vector_text, vectorText} = csg.text
const { CSG, CAG } = csg.csg
function echo(){console.warn("echo() will be deprecated in the near future: please use console.log/warn/error instead");for(var e="",t=arguments,n=0;n<t.length;n++)n&&(e+=", "),e+=t[n];console.log(e)}
// lines above are a JS shim, to make .jscad scripts work as regular JS
// ------------------------------ JS V1 SHIM HEADER ---------------------------------------------------------------------------------------
  
```

footer shim

```js
// ---------------------------- JS V1 SHIM FOOTER -----------------------------------------------------------------------------------------
// this is footer of the JS shim, to export the main and getParameterDefinitions
module.exports = { main }
// some scripts will not have parameters, so we silently ignore the error line below would cause
try{ module.exports.getParameterDefinitions = getParameterDefinitions }catch(e){}

```


## call for action
I just need more users actually trying things, having issues and complaining :) ... I am more motivated scrapping some extra time when ppl come with concrete problems.

hrgdavor:
> I was also pretty shocked that Babel is being used as well. :(

babel is not a requirement for running V2, V3 scripts , but it needs to be suplemented with import maps. And there are limitations how far one can go with import maps (scriptcad does this for jscad).

I use babel, as I am familiar with it, and it allows me to be more flexible towards users. (typescript support, mixing require and imports...).

PR #1331, opened 2024-03-30
Fixed types for snap & generalize functions

iyegoroff:
### All Submissions:

js uses default exports so ts should use them too

### DIFF
--- a/packages/modeling/src/operations/modifiers/snap.d.ts
+++ b/packages/modeling/src/operations/modifiers/snap.d.ts
@@ -1,6 +1,6 @@
 import { Geometry } from '../../geometries/types'
 import RecursiveArray from '../../utils/recursiveArray'
 
-export function snap<T extends Geometry>(geometry: T): T
-export function snap<T extends Geometry>(...geometries: RecursiveArray<T>): Array<T>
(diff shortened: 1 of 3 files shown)


packages/modeling/src/operations/modifiers/snap.d.ts
```diff
@@ -1,6 +1,6 @@
 import { Geometry } from '../../geometries/types'
 import RecursiveArray from '../../utils/recursiveArray'
 
-export function snap<T extends Geometry>(geometry: T): T
-export function snap<T extends Geometry>(...geometries: RecursiveArray<T>): Array<T>
-export function snap(...geometries: RecursiveArray<Geometry>): Array<Geometry>
+export default function snap<T extends Geometry>(geometry: T): T
+export default function snap<T extends Geometry>(...geometries: RecursiveArray<T>): Array<T>
+export default function snap(...geometries: RecursiveArray<Geometry>): Array<Geometry>
```
z3dev, line 6: I don't even know what happens if the default is defined multiple times. I suggest reverting.

z3dev (changes requested):

PR #1348, opened 2024-06-04
fix(all): sync changes to v3

z3dev:
These changes bring in the latest changes from master into the V3 branch, with some changes.

### All Submissions:


platypii (approved):
Changes look good.

Side note, but at some point before releasing V3, we MERGE from master into V3, and then, importantly, DO NOT SQUASH MERGE. That way when eventually V3 becomes the master branch... authors get credit for their commits (right now commits from master post v3 split will be lost).

z3dev:
> Side note, but at some point before releasing V3, we MERGE from master into V3, and then, importantly, DO NOT SQUASH MERGE. That way when eventually V3 becomes the master branch... authors get credit for their commits (right now commits from master post v3 split will be lost).

That would be cool but I haven't seen large merges work well.

PR #1355, opened 2024-07-26
Update scission.d.ts

FishOrBear:
Fix this return result.


platypii (dismissed):
Good catch

platypii:
Technically scission does not support Geom2 so that first line could actually be deleted...

z3dev:
@FishOrBear technically, this is correct, however scission of geom2 is not performed, only the original geom2 objects are returned. Is that ok?

of course, if you submit the code for geom2 scission then we would be really happy.

FishOrBear:
@z3dev ok, I have updated the code

z3dev (approved):
@FishOrBear looks fine. Thanks

PR #1356, opened 2024-07-26
Update type.d.ts

FishOrBear:


platypii (commented):
We use `Poly3.plane` internally, but it's not really meant to be exposed to the user.

Question for @z3dev do we want:
1. types to reflect the internal usage, so that we can typecheck more thoroughly within the project
2. "hide" the internal representations from the user, so that users of the library don't assume it will be there

z3dev:
@FishOrBear Why add the plane to the class? Does this assist in some of you more difficult designs?

FishOrBear:
@z3dev I will access it in my program, but it is not always present, so I marked it as an optional parameter.

z3dev:
@platypii what do you think? This is probably another example of users doing something that is unexpected. The changes are correct.

z3dev (approved):
@FishOrBear these are interesting... but I'm a fan of being detailed. So, let's go with these changes.

please note that the API documents will not provide information on the optional planes.

platypii (approved):
I am in general a fan of having more type safety. I think having it as optional type means that users can't depend on it being there so we have optionality in the future. :+1:

Issue #1357, opened 2024-07-26
hull with a single unconvex solid


her01n:
## Expected Behavior

*hull(solid)* returns a convex hull of solid.

## Actual Behavior

*hull(solid)* returns the argument *solid* even when it is not convex.

## Steps to Reproduce the Problem

```
    const { torus } = require('@jscad/modeling').primitives
    const { hull } = require('@jscad/modeling').hulls
    
    const main = () => {
        return hull(torus())
    }
    
    module.exports = { main }
```

Running this code produces a torus.

## Workaround

Repeat the solid to get the correct behavior, for example:

```
    const { torus } = require('@jscad/modeling').primitives
    const { hull } = require('@jscad/modeling').hulls
    
    const main = () => {
        return hull(torus(), torus())
    }
    
    module.exports = { main }
```

## Specifications

  - Version: JSCAD V2.6.7
  - Platform: tested on https://openjscad.xyz/ web site
  - Environment: Firefox 128.0.2

z3dev:
@her01n please try the latest release. This issue should be resolved. If satisfied then close this issue.

thanks for the fixes!

her01n:
I can confirm the issue is fixed in the recent release.

PR #1358, opened 2024-07-26
fix(modeling): hull of single unconvex solid

her01n:
Issue #1357


platypii (approved):
Nice catch, and thanks for adding a test :+1:

Issue #1360, opened 2024-08-27
Uncaught TypeError in reTesselateCoplanarPolygons


kirevdokimov:
I use **.union()** to merge geometries, but when I do so, it causes errors.

## Expected Behavior
Geometries are merged with no issues

## Actual Behavior
**.union()** throws an exception
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')
    at reTesselateCoplanarPolygons (reTesselateCoplanarPolygons.js:176:40)
    at retessellate.js:25:44
    at Array.forEach (<anonymous>)
    at retessellate (retessellate.js:23:14)
    at union8 (unionGeom3.js:21:17)
    at union8 (union.js:42:35)
 ```

## Steps to Reproduce the Problem

I assume this basic example works for everyone
```
const c1 = cylinder({ radius: 3, height: 12, center: [0, 0, 5], segments: 12 })
const c2 = cylinder({ radius: 3, height: 18, center: [0, 0, 5], segments: 12 })

const u = union(c1, c2)
```

## Specifications

  - Version:  "@jscad/core": "^2.6.9", "@jscad/modeling": "^2.12.2",
  - Platform: Windows 11, Brave Browser,
  - Environment: Node v16.15.1, esbuild, Brave Browser Version 1.69.153 Chromium: 128.0.6613.85 (Official Build) (64-bit)
  
  Same issue for Google Chrome Version 127.0.6533.122 (Official Build) (64-bit)

**Comment by kirevdokimov** [comment]
polygonindexKey is equal to "clear" or "remove" while we iterate over `startingpolygonindexes`, what causes `vertices2d` to be undefined.
![image](https://github.com/user-attachments/assets/a2fe1d5b-d5a6-4a07-9fea-6a6d58cb8111)
![image](https://github.com/user-attachments/assets/429a78c0-9d82-465e-a93c-b9360edcbf99)

If we check for Nan with `if (isNaN(polygonindexKey)) continue` line, other issue occurs:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '1')
    at interpolateBetween2DPointsForY (interpolateBetween2DPointsForY.js:11:16)
    at reTesselateCoplanarPolygons (reTesselateCoplanarPolygons.js:226:15)
    at retessellate.js:25:44
    at Array.forEach (<anonymous>)
    at retessellate (retessellate.js:23:14)
    at union8 (unionGeom3.js:21:17)
[...]

PR #1361, opened 2024-08-27
Patch Issue #1360

kirevdokimov:
### All Submissions:

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
- [ ] Does your submission pass tests? (did not run tests yet)

It works smoothly for a sample case from the issue and for my complex real case

### DIFF
--- a/packages/modeling/src/operations/modifiers/reTesselateCoplanarPolygons.js
+++ b/packages/modeling/src/operations/modifiers/reTesselateCoplanarPolygons.js
@@ -171,6 +171,7 @@ const reTesselateCoplanarPolygons = (sourcepolygons) => {
       // update activepolygons by adding any polygons that start here:
       const startingpolygonindexes = topy2polygonindexes.get(ycoordinate)
       for (const polygonindexKey in startingpolygonindexes) {
+        if(isNaN(polygonindexKey)) continue
         const polygonindex = startingpolygonindexes[polygonindexKey]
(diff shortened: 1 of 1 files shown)


z3dev:
@kirevdokimov The changes are straight forward, and probably fix the special case. But I'm wondering if there's another issue with the algorithm, or the union is returning some strange polygons.

im not sure that I'll have time to research further, so I added @platypii to the review.

kirevdokimov:
@z3dev Retesselation works perfectly fine with this one issue. It is not a specific case. There are cases with union, subtract and anything else that uses retesselation. The only problem with commenting retesselation out from these functions is that some triangles are left in air.
image
image

z3dev:
> The only problem with commenting retesselation out from these functions is that some triangles are left in the air.

I was suspecting something like this. Hopefully, @platypii will have some time to investigate as the simple design is producing a nice test case.

FYI, the fix looks fine.

platypii:
The fix seems fine, but I haven't yet been able to reproduce the original error on master?
https://github.com/jscad/OpenJSCAD.org/issues/1360#issuecomment-2329728567

z3dev:
@kirevdokimov Can you supply a small test case?

z3dev:
No tests, no merge. It should be easy.

PR #1393, opened 2025-05-11
fix(regl-renderer): added handling of element resize events

z3dev:
### All Submissions:


hrgdavor:
It would ne better to use resize observer. It reacts to changes in canvas size that can be result of layout change or some other Dom changes

https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

Issue #1253, opened 2023-05-29
trying to extrude an SVG ends up with weird non-solid object that behaves unexpectedly


tsdexter:
## Expected Behavior

I expect to see a solid 'f' 

## Actual Behavior
The object that is extruded from the SVG looks like it's not solid, has no top face:
![image](https://github.com/jscad/OpenJSCAD.org/assets/4126987/d992632c-93f5-45b3-a6ec-017528d630af)

However, when I rotate the viewport around, for example, to look at it from underneath then is appears there is a top surface: 
![image](https://github.com/jscad/OpenJSCAD.org/assets/4126987/6ffc86e8-c893-46e3-9826-952db6093526)

When I subtract the extruded SVG from a cube, it behaves strangely, looks more like an intersection of the two: 

Cube and 'f' with no subtraction:
![image](https://github.com/jscad/OpenJSCAD.org/assets/4126987/d05954c8-0984-44a7-a263-c69985f53d58)

Result of `subtract(cube({ size: 10 }), extrudedSvg)`:
![image](https://github.com/jscad/OpenJSCAD.org/assets/4126987/e115109c-e605-408c-85d1-7c385abc2dbe)

Even a basic path to create a square results in this behaviour (comment out line 13 to reproduce): 
![image](https://github.com/jscad/OpenJSCAD.org/assets/4126987/0223b04c-1e1a-4e9e-99b6-1729c48abeea)

## Steps to Reproduce the Problem

https://codesandbox.io/s/jscad-react-js-demo-forked-7soe65?file=/src/App.js

## Specifications

see codesandbox

tsdexter:
I also tried testing this directly on openjscad.xyz but I get: 

`SyntaxError: Cannot find module @jscad/svg-deserializer`

How do I load serializers into openjscad.xyz?

z3dev:
@tsdexter see #1087 which discusses the issues with SVG. Hopefully, you will find a few hints.

tsdexter:
@z3dev thanks, I did see that and was hoping it wasn't related... Going back for a closer look, I was able to get some working extrusions by inverting them all and then knowingly combining specific entries in the returned array via `intersect` like: 

```js
union(solids[0], intersect(solids[1], solids[2]), solids[3], intersect(solids[4], solids[5]))
```

I had to know which indexes were holes on which solids. However, I am doing this on dynamic/generated SVG paths, so this is not ideal. Is there any plans to improve the SVG importing in the near future?

[...]

Issue #1087, opened 2022-05-12
SVG path with "hole" does not import correctly


kbruneel:
## Expected Behavior
When I import an svg path with a hole, the hole is not shown.

## Actual Behavior
Hole is not visible.

## Steps to Reproduce the Problem

  1. Past the following code in https://openjscad.xyz/
```
const { extrusions } = require('@jscad/modeling')
const { deserializers } = require('@jscad/io')

const main = () => {

  let s_text = `
  <svg>
    <path
       style="color:#000000;fill:#000000;-inkscape-stroke:none"
       d="m 30,24.390625 c -9.148585,0 -17.605395,4.881781 -22.1796875,12.804687 -4.5742924,7.922907 -4.5742924,17.686469 0,25.609375 C 12.394605,70.727594 20.851415,75.609375 30,75.609375 H 70 C 84.136772,75.609374 95.609375,64.136772 95.609375,50 95.609375,35.863227 84.136773,24.390625 70,24.390625 Z m 0,1.21875 H 70 C 83.477437,25.609375 94.390625,36.522563 94.390625,50 94.390625,63.477436 83.477436,74.390624 70,74.390625 H 30 c -8.714677,0 -16.765708,-4.648181 -21.1230469,-12.195313 -4.3573385,-7.547131 -4.3573385,-16.843493 0,-24.390625 C 13.234292,30.257556 21.285323,25.609375 30,25.609375 Z"
       id="path880" 
    />
  </svg>
  `

  let s = deserializers.svg({ output: "geometry", target: "geom2"}, s_text)
  s = extrusions.extrudeLinear({ height: 10 }, s)
  return s
}


module.exports = { main }

```

2. What it should look like:
```
const { extrusions, transforms, geometries, booleans } = require('@jscad/modeling')
const { deserializers } = require('@jscad/io')

const main = () => {

  let svg_text = `
  <svg>
    <path
       style="color:#000000;fill:#000000;-inkscape-stroke:none"
[...]

Issue #888, opened 2021-07-22
svg deserializer constructs paths in reverse order


brucedjones:
## Expected Behavior
Geometries have correct normals when extruded from svg

## Actual Behavior
Geometry normals are flipped.
```

## Steps to Reproduce the Problem
Run the code below. Note you must be using the fixes described here: https://github.com/jscad/OpenJSCAD.org/pull/887

```
const { deserializers } = require("@jscad/io");

const main = () => {
  const svg = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="5.6298828125" height="11.71875"><path d="M5.37 6.60L5.37 6.68Q5.37 7.98 4.83 8.68Q4.29 9.38 3.32 9.38L3.32 9.38Q2.51 9.38 2.02 8.74L2.02 8.74L1.95 9.28L0.47 9.28L0.47 1.78L2.12 1.78L2.12 4.42Q2.58 3.90 3.31 3.90L3.31 3.90Q4.29 3.90 4.83 4.60Q5.37 5.31 5.37 6.60L5.37 6.60ZM3.72 6.58L3.72 6.58Q3.72 5.82 3.51 5.49Q3.31 5.17 2.89 5.17L2.89 5.17Q2.33 5.17 2.12 5.59L2.12 5.59L2.12 7.69Q2.33 8.11 2.90 8.11L2.90 8.11Q3.48 8.11 3.64 7.54L3.64 7.54Q3.72 7.27 3.72 6.58Z"/></svg>`;
  const shapes = deserializers.svg({ output: "geometry", target: "geom2" }, svg);
  return extrudeLinear({ height: 10 }, shapes);
}

module.exports = { main }
```


## Specifications

  - Version: 2.5.0
  - Platform: macos / chrome
  - Environment: browser

<!-- Love openjscad? Please consider supporting our collective:
👉  https://opencollective.com/openjscad/donate -->

z3dev:
SVG paths don’t have a concept of rotation. So, yeah. Some paths with definitely be going clockwise about Z. You’ll have to do some calculations to determine if the path needs to be reversed before extruding, or converting to 2D geometry.

brucedjones:
Hm I'm going to have to do some reading on this... Fwiw the svg in the example is a filled polygon, there must be some convention to that for it to work and I think a user would intuitively expect the svg deserializer to figure it out

z3dev:
[...]

PR #1400, opened 2025-08-17
fix(io): svg path should return one geom2

platypii:
I was trying to use JSCAD to extrude an SVG into a 3d object and was frustrated that it didn't handle holes right.

This fixes the long standing issue of svg holes (#1087, #888, #1253)

Before:
Screenshot 2025-08-17 at 13 54 52

After:
Screenshot 2025-08-17 at 13 55 01

### All Submissions:


z3dev:
@platypii I hanks for the fixes. I won't be able to review / merger until this weekend. But I was going to make releases this weekend anyway. So, hopefully this will be included.

just for kickers, did you try some of the issues against this?

PR #1408, opened 2025-11-18
Improve .obj serialization times - fixes #1407

### LINKED ISSUE #1407: Serilization duration varies greatly between obj and stl
Hello OpenJSCAD team! First of all thank you for your hard work on this library, this is by far the most promising we encountered for boolean operations on the web.

As we tried implementing different file formats to our tool, we noticed the format has a strong impact on processing times. 

```js
    const blob = solidsAsBlob([...finalMeshes], {
      format: outputFormat,
    })
    return blob
```

When `outputFormat` is `stlb` it's way faster than `obj`, but I can't find the reason for this. Passing `triangulate: false` (as our mesh should already be triangulated) to the options doesn't seem to change anything either.  

Do you have any idea on how we can speed up that process, or what area of `obj-serializer` could need a PR for this?   

Thank you in advance!

ZephyrRaine:
### All Submissions:

### Description: 
Replaced `vertices` array with a Map in order to check existence with simple O(1) lookup instead of O(n) `indexOf`. Greatly reduces export times and no diff detected on .obj outputs. This fixes #1407 where more details can be found.

### Tests:
All tests passed, except one for web that already wasn't passing after a fresh clone.


z3dev (approved):
@ZephyrRaine These changes look fine. It would be nice to get some statistics on the performance improvement.

By the way, V2 changes are released periodically so these changes will probably be released in January. Is that ok?

PR #1433, opened 2026-02-06
feat(modeling): add Minkowski sum operation for 3D geometries

jbroll:
Implements Minkowski sum A ⊕ B which "inflates" geometry A by shape B. Common uses: rounding edges with sphere, chamfering with cube.

- Convex + convex: uses convex hull of pairwise vertex sums
- Non-convex + convex: decomposes into tetrahedra, unions results
- Uses face-local apex decomposition (handles torus, U-shapes correctly)
- Exported from both minkowski and booleans namespaces

### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?

### DIFF
--- a/packages/modeling/src/operations/minkowski/minkowskiSum.js
+++ b/packages/modeling/src/operations/minkowski/minkowskiSum.js
@@ -1,0 +1,224 @@
+const flatten = require('../../utils/flatten')
+
+const geom3 = require('../../geometries/geom3')
+const poly3 = require('../../geometries/poly3')
+
+const hullPoints3 = require('../hulls/hullPoints3')
+const unionGeom3 = require('../booleans/unionGeom3')
+
+/**
+ * Compute the Minkowski sum of two 3D geometries.
+ *
+ * The Minkowski sum A ⊕ B is the set of all points a + b where a ∈ A and b ∈ B.
+ * Geometrically, this "inflates" geometry A by the shape of geometry B.
+ *
+ * Common use cases:
+ * - Offset a solid by a sphere to round all edges and corners
+ * - Offset a solid by a cube to create chamfered edges
+ * - Collision detection (if Minkowski sum contains origin, shapes overlap)
+ *
+ * For best performance, use convex geometries. Non-convex geometries are supported
+ * when the second operand is convex, but require decomposition and are slower.
+ *
+ * @param {...Object} geometries - two geom3 geometries (second should be convex for non-convex first)
+ * @returns {geom3} new 3D geometry representing the Minkowski sum
+ * @alias module:modeling/operations/minkowski.minkowskiSum
+ *
+ * @example
+ * const { primitives, minkowski } = require('@jscad/modeling')
+ * const cube = primitives.cuboid({ size: [10, 10, 10] })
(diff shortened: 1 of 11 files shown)


platypii:
I haven't tried this yet, but very impressive that it was implemented in so few lines of code. Very cool!

Minkowski is used very widely in SCAD designs... if we had this in JSCAD it would be much more feasible to port a lot of existing SCAD designs to JSCAD

packages/modeling/src/operations/minkowski/minkowskiSum.test.js
```diff
@@ -0,0 +1,199 @@
...
+
+test('minkowskiSum: cube + cube produces correct bounds', (t) => {
+  // Cube1: size 10 (±5 from origin)
+  // Cube2: size 4 (±2 from origin)
+  // Minkowski sum should be size 14 (±7 from origin)
+  const cube1 = cuboid({ size: [10, 10, 10] })
+  const cube2 = cuboid({ size: [4, 4, 4] })
+
+  const result = minkowskiSum(cube1, cube2)
+
```
mister-jscad: @platypii has criticised the old test suites that check vertices. he suggests something like...

  t.notThrows(() => geom3.validate(obs))
  t.is(measureArea(obs), 24)
  t.is(measureVolume(obs), 7.999999999999999)
  t.is(pts.length, 6)

It kind of makes sense as we really want proper 3D geometries. FYI, we have some failing tests. :(

hrgdavor:
It is not unusual to assume wrong, you were wise enough to check the assumptions. 

You mention choosing manifold engine ... can the example be run with manifold too ?  or you have other test scripts with manifold where that stat is needed.

jbroll:
See the drop down menu at the top in the Options area - you can choose geometry and rendering engines dynamically.  The Manifold engine gets the Minkowski wrong for "hollow" Geom3 shapes.  :)

packages/modeling/src/operations/minkowski/minkowskiSum.js
```diff
@@ -0,0 +1,224 @@
...
+    // Get polygon plane (normal + offset)
+    const plane = poly3.plane(polygon)
+    const nx = plane[0], ny = plane[1], nz = plane[2]
+
+    // Offset inward along negative normal to create face-local apex
+    // The normal points outward, so we go in the negative direction
+    // Use a small offset - the actual distance doesn't matter much
+    // as long as the apex is on the interior side of the face
+    const offset = 0.1
+    const apex = [
```
mister-jscad: If I understand correctly, the apex is calculated, and used as one of the vertices in the tetrahedron polygons. Maybe vec3.create() should be used instead of [] nomenclature, which will make this more apparent.

jbroll:
Fixed TypeScript test failure - `isConvex.d.ts` was using a named import instead of default import for `Geom3` type. All tests now passing (including `test:tsd`).

packages/modeling/src/operations/minkowski/minkowskiSum.js
jbroll: Added inline comment clarifying apex is used as a vertex in the tetrahedron polygons

hrgdavor:
🎉

PR #1445, opened 2026-03-07
build(all): switch to NPM workspaces, update all dependencies, rework rollups

z3dev:
Pretty massive change in the V3 packaging. This now uses NPM workspaces, which required changing all the package.json dependencies. (I got tired of working through PNPM issues during installations.)

In addition, most of the dev dependencies have been moved up to supported versions. This required a few changes to tests due to AVA updates.

And finally, the latest Rollup required some changes to the configurations for creating banners.

Lots of fun!

### All Submissions:


platypii (approved):
WOOO!! this is a great change! Npm is not perfect but I always prefer to use native tooling if possible. pnpm caused so many headaches

PR #1463, opened 2026-06-27
 fix(modeling): various fixes, style changes, and small enhancements

z3dev:
retessellateCoplanarPolygons() was creating invalid poly3 objects with less than two vertices. this is the fix.

FYI, the original code has the same issue, and this bug is present in V2 still.

### All Submissions:


mister-jscad (approved):
Looks good.

